PR #15, opened 2021-06-08
Morphological Transformation for Question Answering and Text Classification

samsontmr:


transformations/english_inflectional_variation/transformation.py
```diff
@@ -0,0 +1,98 @@
+import random, lemminflect
+from typing import Dict, List, Optional, Tuple
+from tokenizers.pre_tokenizers import BertPreTokenizer
+from nltk.tag.mapping import map_tag
+from nltk.tag.perceptron import PerceptronTagger
+try:
+    PerceptronTagger()
```
kaustubhdhole, line 7: I think all is good in your PR -> only line 6 to 17 should be moved into the constructor.

kaustubhdhole (approved):

PR #25, opened 2021-06-12
rename wrongly named keywords, make tagging look better

kaustubhdhole:


test/test_main.py
```diff
@@ -6,30 +6,30 @@
 from interfaces.TaggingOperation import TaggingOperation
 
 
-def getMessage(transformation, impl_name):
-    return "For {transformation}, no transformation or test case found for {impl_name}".format(
-        transformation=transformation, impl_name=impl_name
+def getMessage(transformation_type, impl_name):
```
aadesh11, line 9: We can remove "impl_name" if we are not using it.

transformations/longer_names_ner/test.json
```diff
@@ -1,137 +1,29 @@
...
-        "I-PER",
-        "I-PER",
-        "O",
-        "O",
-        "B-COUNTRY",
-        "B-COUNTRY"
-      ]
+      "input_sequence": "Neil Alden Armstrong was an American astronaut",
+      "output_sequence": "Neil Y Alden Armstrong was an American astronaut",
+      "input_tag": "B-PER I-PER I-PER O O B-COUNTRY B-COUNTRY",
```
aadesh11, line 13: Input tag should be: "B-PER I-PER **B**-PER O O B-COUNTRY B-COUNTRY".
Can we double-check this?

kaustubhdhole: No.. the tags are fine..

TestRunner.py
```diff
@@ -28,30 +29,43 @@ def load_test_cases(test_json):
 
 
 class TransformationRuns(object):
-    def __init__(self, interface, perturbation_type, load_tests=True):
+    def __init__(self, interface, name_of_transformation, load_tests=True):
```
samsontmr, line 32: Would it be more convenient to use `transformation_name` instead?

TestRunner.py
```diff
@@ -62,9 +76,26 @@ def get_all_transformations(query_task_type: TaskType) -> Iterable:
...
 
+    @staticmethod
+    def get_all_transformation_names(heavy=False) -> Iterable:
+        # iterate through the modules in the current package
+        package_dir = Path(__file__).resolve()  # --> TestRunner.py
+        transformations_dir = package_dir.parent.joinpath("transformations")
+        for (_, m, _) in iter_modules([transformations_dir]): # ---> ["back_translation", ...]
+            t_py = import_module(f"transformations.{m}.transformation")
+            for name, obj in inspect.getmembers(t_py):
+                if convert_to_camel_case(m) == name:
```
kaustubhdhole, line 87: @samsontmr For handling your case - this would need to be changed.

TestRunner.py
kaustubhdhole: Done..

test/test_main.py
kaustubhdhole: Done..

aadesh11 (approved):

samsontmr (approved):

PR #27, opened 2021-06-12
Filter evaluation&dataset class change

Nickeilf:
- modifying filter testcase, now can specify which filter to test with pytest option
- changing the name convention of "test.json", changing from "filter_args" to "inputs"
- adding labels into TextLineDataset, all dataset-level operation will change labels as well
- changing the return type of several functions in dataset classes (before was returning data, now returning a new dataset class instance)
- integrating dataset class into the evaluation flow, adding classmethods to construct dataset from huggingface dataset. Such encapsulation might be helpful when we what to evaluation dataset-level operations (e.g. contrast sets) in the future implementations.
- Updating the calculation of BLEU from averaging sentence bleu to corpus bleu. Averaging sentence bleu is not commonly used, see [here](https://github.com/mjpost/sacrebleu/blob/e22640/sacrebleu/compat.py#L66-L67).

### DIFF
--- a/TestRunner.py
+++ b/TestRunner.py
@@ -96,24 +96,24 @@ class TransformationRuns(object):
 
 class FilterRuns(object):
 
-    def __init__(self):
+    def __init__(self, name_of_filter):
         filters = []
         filter_test_cases = []
         package_dir = Path(__file__).resolve()  # --> TestRunner.py
-        filters_dir = package_dir.parent.joinpath("filters")
-        for (_, m, _) in iter_modules([filters_dir]):
-            t_py = import_module(f"filters.{m}")
-            t_js = os.path.join(filters_dir, m, "test.json")
+        filters_dir = package_dir.parent.joinpath("filters", name_of_filter)
+        
+        t_py = import_module(f"filters.{name_of_filter}")
+        t_js = os.path.join(filters_dir, "test.json")
 
-            for test_case in load_test_cases(t_js):
-                class_name = test_case["class"]
-                class_args = test_case["args"]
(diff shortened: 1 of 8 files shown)


TestRunner.py
```diff
@@ -96,24 +96,24 @@ def get_all_transformation_names(heavy=False) -> Iterable:
 
 class FilterRuns(object):
 
-    def __init__(self):
+    def __init__(self, name_of_filter):
         filters = []
         filter_test_cases = []
         package_dir = Path(__file__).resolve()  # --> TestRunner.py
-        filters_dir = package_dir.parent.joinpath("filters")
-        for (_, m, _) in iter_modules([filters_dir]):
```
kaustubhdhole, line 104: You might still want to keep this for the automated runs -
Like the below command runs all transformations which are not heavy for every PR!
 https://github.com/GEM-benchmark/NL-Augmenter/blob/a5905a08de2f7278ea5fb370bb0fcf4a71c7720a/.github/workflows/nl-action.yml#L30

Nickeilf: I see. I think there are two options for testing filters: 
1) we add in the test_filter function with a few if clauses to determine the test mode (light, all, etc.) 
2) we can merge the test case for transformation and filter, so there will be only one `test_operation()` function and will determine whether to execute the transformation or filter test cases.

kaustubhdhole: I think the second option might be better..

kaustubhdhole: Tagging @aadesh11 here so that we don't duplicate anything!

PR #29, opened 2021-06-13
Some refactoring on evaluation engine.

tongshuangwu:
Some changes to the code as I try to create the leaderboard:

1. change the arg name `model` to `model_name` in `evaluation_engine` to make it consistent with others. It also fixes the bug in the colab
2. add returns to all the eval function.
3. add a simple input key arg to text classification. Also added exception to this function when cannot find the data entry. 

TODO: should extend text classification beyond just sentiment analysis; And should expose the instance keyword and label keywords to let people build the mapping (rather than pre-specify as pos/neg/1/0)

### DIFF
--- a/evaluation/evaluation_engine.py
+++ b/evaluation/evaluation_engine.py
@@ -2,6 +2,7 @@ from datasets import load_dataset
 from transformers import pipeline
 from sacrebleu import sentence_bleu
 from seqeval.metrics import accuracy_score
+import numpy as np
 
 from interfaces.QuestionAnswerOperation import QuestionAnswerOperation
 from interfaces.SentenceOperation import SentenceOperation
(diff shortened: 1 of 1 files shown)


evaluation/evaluation_engine.py
```diff
@@ -226,33 +235,57 @@ def evaluate_text_summarization(
...
     if model_name is None:
         model_name = "aychang/roberta-base-imdb"
     # (2) load test set
     if dataset_name is None:
         dataset_name = "imdb"
-    print("Loading <%s> dataset to train <%s> model", dataset_name, model_name)
-    dataset = load_dataset(dataset_name, split=split)
+        input_key = "text"
+    print(f"Loading <{dataset_name}> dataset to evaluate <{model_name}> model.")
+    if dataset_name in ["qqp", "sst2"]:
```
kaustubhdhole, line 259: Probably QQP would need a separate interface than SST - taking in two sentences..

kaustubhdhole (approved):

PR #33, opened 2021-06-14
Test and evaluation code refactor

aadesh11:
Added support to run bulk and single transformation test case. 
Refactored the code of test_main, TestRunner and evaluation engine.

### DIFF
--- a/TestRunner.py
+++ b/TestRunner.py
@@ -1,13 +1,12 @@
+import inspect
 import json
 import os
-import inspect
 import re
-
 from importlib import import_module
 from pathlib import Path
 from pkgutil import iter_modules
-
 from typing import Iterable
+
 from tasks.TaskTypes import TaskType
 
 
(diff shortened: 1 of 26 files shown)


transformations/butter_fingers_perturbation/test.json
```diff
@@ -2,26 +2,32 @@
   "type": "butter_fingers_perturbation",
   "test_cases": [
     {
+      "class": "ButterFingersPerturbation",
```
kaustubhdhole, line 5: It's better that this be at the outside level. We will need to repeat this in every test case then.

aadesh11: Yeah, I was also thinking. Will do the required changes.

TestRunner.py
```diff
@@ -29,42 +28,37 @@ def load_test_cases(test_json):
...
+                if transformation_name:
+                    if name == transformation_name:
+                        for test_case in load_test_cases(t_js):
+                            if test_case["class"] == transformation_name:
+                                transformation_test_cases.append(test_case)
+                        is_done = True
                         break
+                elif inspect.isclass(obj) and hasattr(obj, "heavy"):
+                    # Run all the test cases based on heavy parameter.
+                    if heavy == obj.is_heavy():
```
kaustubhdhole, line 50: If this is the case, then "all" won't run the heavy cases, right?

test/test_main.py
```diff
@@ -1,112 +1,127 @@
...
-
-
-def execute_test_case_for_filter(filter_name):
+        execute_tagging_test_case(impl, t_test_cases)
+
+
+def test_transformations(transformation_name):
+    """Entry point to run transformation test cases based on transformation_name
+    transformation_name: Should be "light"/"all" or any transformation class name like 'ButterFingersPerturbation'"""
+    is_heavy = False if transformation_name == "light" else True
```
kaustubhdhole, line 103: is_heavy = not transformation_name == "light"

TestRunner.py
aadesh11: It will be based on the heavy parameter, if that's true and the user has passed "all", then it will run all the transformations where the "heavy" flag is true.

kaustubhdhole: But it should run those where the heavy flag is false also.. if you want to keep the function as it is, you should call this twice when "all" is called.

PR #38, opened 2021-06-16
Spellchecker perturbation

gurunath-p:
- This transformation helps add noise to a sentence by replacing words using a specified  
 probability with suggestions generated by a spell-checker. 
 - PyEnchant library is used to generate suggestions for the selected word and a random   
 word from the top-5 suggestions is picked as a replacement for perturbation.
 - This can make the model adapt to noise via replaced noisy words.


tuetschek (changes requested):
Thanks for the PR, looks good to me overall:
* The code looks OK
* The transformation is reasonable
* The task types are chosen well
* There are examples, the added dependency is included in requirements.txt

There are still a few things to fix:
* Do you make sure that the spelling suggestion isn't exactly the same as the source word? I don't see anything like that in the code, but maybe the library handles that itself? In the examples, it looks like some sentences are unchanged -- that might just be randomness, but I would like to rule this out.
* I think that the probability parameter should be mentioned in the readme as well.
* Could you format the JSON with examples so that it's better readable (just add indent=4 to the dumps method)?
* The automatic tests aren't passing -- any idea why?

transformations/spellchecker_perturbation/transformation.py
```diff
@@ -0,0 +1,70 @@
+import random
+
+import enchant
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def is_case_same(char_1: str, char_2: str) -> bool:
```
Saad-Mahamood, line 9: Is there a need for this function to be outside the class? Is it used anywhere else?

transformations/spellchecker_perturbation/test.json
```diff
@@ -0,0 +1 @@
...ng, such as Paul likes coffee and Mary tea, lack an overt predicate to indicate the relation between two or more arguments."}, "outputs": {"sentence": "Sentences wits gapping, such as Paul likest coffee and Mary tea, lack an overt predicate to indicate the relation between two or more arguments."}}, {"class": "SpellCheckerPerturbation", "inputs": {"sentence": "Alice in Wonderland is a 2010 American live-action/animated dark fantasy adventure film"}, "outputs": {"sentence": "Alice in Wonderland is a 2010 American live-action/animated dark fantasy adventure film"}}, {"class": "SpellCheckerPerturbation", "inputs": {"sentence": "Ujjal Dev Dosanjh served as 33rd Premier of British Columbia from 2000 to 2001"}, "outputs": {"sentence": "Ujjal Dev Dosanjh swerved as 33rd Premier of British Columbia from 2000 to 2001"}}, {"class": "SpellCheckerPerturbation", "inputs": {"sentence": "Neuroplasticity is a continuous processing allowing short-term, medium-term, and long-term remodeling of the neuronosynaptic organization."}, "outputs": {"sentence": "Neuroplasticity is a continuous pr\u00e9cising alloying short-term, medium-term, and long-term remodeling of the neuronosynaptic organization."}}]}
```
Saad-Mahamood, line 1: Can you please format this JSON appropriately instead of being on a single line.

transformations/spellchecker_perturbation/README.md
```diff
@@ -0,0 +1,41 @@
...
+    month = nov,
+    year = "2019",
+    address = "Hong Kong, China",
+    publisher = "Association for Computational Linguistics",
+    url = "https://www.aclweb.org/anthology/D19-1435",
+    doi = "10.18653/v1/D19-1435",
+    pages = "4259--4269",
+}
+```
+
```
Saad-Mahamood, line 38: I would add a section to this README.md that gives a concrete example of input and the output.

Saad-Mahamood (changes requested):
I echo the comments made by @tuetschek. His changes should be acted upon first before this PR can be approved.

requirements.txt
```diff
@@ -8,6 +8,9 @@ torch==1.8.1
 num2words==0.5.10
 word2number==1.1
 
+# for SpellCheckerPerturbation
+pyenchant==3.2.0
```
kaustubhdhole, line 12: This can't be integrated since the library used is not MIT/Apache Licensed. I would suggest to look for a replacement.

kaustubhdhole:
Hi @gurunath-p, would you like to complete this pending PR?

transformations/spellchecker_perturbation/transformation.py
gurunath-p: No, will change this as requested.

transformations/spellchecker_perturbation/test.json
gurunath-p: Sure, will do

gurunath-p:
Hi @kaustubhdhole @Saad-Mahamood @tuetschek 
Apologies for my late reply.
I will check the requested changes and make it ready this weekend. Regarding the tests, because of randomness, the outputs don't always seem to match the input. Will check how to do this better.

kaustubhdhole:
I am closing this now since there haven't been changes since a while and the library pyenchant is not MIT/Apache License. Please feel free to repoen it later when you make changes before September 1.

PR #40, opened 2021-06-16
Add Mixed Language Perturbation Transformation

gentaiscool:
This perturbation translates randomly picked words in the text from English to other languages (e.g., German). It can be used to test the robustness of a model in a multilingual setting.

Authors: Genta Indra Winata, Samuel Cahyawijaya, and Bryan Wilie.


transformations/mixed_language_perturbation/test.json
```diff
@@ -0,0 +1,50 @@
+{
+  "type": "mixed_language_perturbation",
+  "test_cases": [
+    {
+      "class": "MixedLanguagePerturbation",
+      "inputs": {
+        "sentence": "Andrew finally returned the Comic book to Chris that I bought last week"
+      },
+      "outputs": {
+        "sentence": "Andrew finally returned Die Comic book to Chris that I bought last Woche"
```
Nickeilf, line 10: It seems like all the code-mixed words are by default in uppercasing. I guess that's caused by the translation model, which would always uppercase the first input. I think the perturbation would be more natural if the casing for the original word and the code-mixed word is the same. For example, "the" -> "die", "Comic" -> "Komik".

transformations/mixed_language_perturbation/transformation.py
```diff
@@ -0,0 +1,89 @@
...
+    return mixed_text
+
+class MixedLanguagePerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION
+    ]
+    languages = ["en"]
+    tgt_languages = ["es", "de", "fr", "zh"]
+
```
Nickeilf, line 51: It would be good to add a `heavy=True` mark just like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/0672d7163338c3c744c9964e28084372781ab349/transformations/back_translation/transformation.py#L10). Therefore, when running pytest, it would by default ignore the heavy transformations so that the pretrained model won't be loaded.

transformations/mixed_language_perturbation/transformation.py
```diff
@@ -0,0 +1,89 @@
...
+                continue
+
+            mixed_text += translate(model, tokenizer, plain_word, src_lang, trg_lang)[0]
+
+            if word[-1] in string.punctuation:
+                mixed_text += word[-1]
+        else:
+            mixed_text += word
+
+    print(text, mixed_text)
```
Nickeilf, line 41: Maybe it's better to avoid printing inside the transformation because one might want to transform the whole dataset and this can overwhelm the output screen.

Nickeilf (commented):
Hi @gentaiscool, thanks for this interesting transformation!

**Specificity**: The change is specific to multilingual NLP tasks.
**Novelty**: This transformation is novel and tests the robustness of models in a multilingual setting.

@kaustubhdhole we haven't thought about testing the robustness in multilingual settings. It might not make sense to test this transformation with a unilingual model because it would definitely decrease the performance. Maybe we should add one evaluation entry for a multilingual task, such as XNLI?

transformations/mixed_language_perturbation/test.json
gentaiscool: Good point! We followed your suggestion and pushed it in our latest commit :)

transformations/mixed_language_perturbation/transformation.py
gentaiscool: Thank you 👍 We added `heavy=true` in our latest commit.

gentaiscool:
> Hi @gentaiscool, thanks for this interesting transformation!
> 
> I've left some comments in the code and here's my general review:
> **Correctness**: All checks have passed - please pull "main" once into your branch.
> **Interface**: I think you've chosen the correct interface.
> **Specificity**: The change is specific to multilingual NLP tasks.
> **Novelty**: This transformation is novel and tests the robustness of models in a multilingual setting.
> **Adding New Libraries**: No new library is added.
> **Test Cases**: 5 test cases have been added.
> **Evaluating Robustness**: Robustness evaluation is not conducted.
> 
> @kaustubhdhole we haven't thought about testing the robustness in multilingual settings. It might not make sense to test this transformation with a unilingual model because it would definitely decrease the performance. Maybe we should add one evaluation entry for a multilingual task, such as XNLI?

Thank you, @Nickeilf, for the review! We have addressed your comments and pushed the commits to solve the lower case issue and add `heavy=true`. I also checked the pytest. The current script follows the current input-output procedure, and there should not be any errors from pytest.

Let us know if we need to change anything from the PR. Thanks a lot!

transformations/mixed_language_perturbation/transformation.py
gentaiscool: Thanks for mentioning this! We removed the print.

transformations/mixed_language_perturbation/transformation.py
```diff
@@ -0,0 +1,93 @@
...
+
+def mixed_language(model, tokenizer, text, prob_mix=0.1, src_lang="en", trg_lang="fr", seed=0):
+    random.seed(seed)
+
+    words = text.split()
+    mixed_text = ""
+    for word in words:
+        if mixed_text != "":
+            mixed_text += " "
+        if random.random() < prob_mix:
```
kaustubhdhole, line 28: In the paper you cited, you seem to select the word with the highest attention score rather than at random. Is there any performance difference on cross-lingual training between random and highest attention?
Even if you select words at random, that's fine as German words in this case can probably work as regularizers - but why not use a bilingual lexicon - since you are anyways translating words?

kaustubhdhole:
> @kaustubhdhole we haven't thought about testing the robustness in multilingual settings. It might not make sense to test this transformation with a unilingual model because it would definitely decrease the performance. Maybe we should add one evaluation entry for a multilingual task, such as XNLI?

@Nickeilf you are right. We could add an evaluation entry for a multilingual task like XNLI. There are a few models and the XNLI dataset on HF too - so shouldn't be a big problem. I think this PR can skip robustness evaluation since they've also proved it in the attached papers. :)

transformations/mixed_language_perturbation/transformation.py
SamuelCahyawijaya: @kaustubhdhole : We choose random instead of taking the highest attention score, as it is a much cheaper approach and can produce more diverse augmentation. And you are also correct that it can work as a regularizer as well. We don't use a bilingual lexicon as it will produce a large multilingual dictionary, and the replacement is limited to the word inside the dictionary.

We think it is another good idea to create a light (`heavy=False`) version of this perturbation with a bilingual lexicon. Should we add another perturbation for this matter or combine both approaches using lexicon for common words and model translation for uncommon words?

kaustubhdhole: Thanks for the clarification. That'd be a great idea! You can add it as a separate perturbation/transformation from the current one.

Nickeilf (approved):
Thanks for the changes @gentaiscool! 

I think this transformation is pretty nice now and I've passed my review. We'll add an entry for evaluating cross-lingual tasks, i.e. XNLI, in the later updates. It would be great if you can test this transformation later and update the scores in the README.

Cheers.

transformations/mixed_language_perturbation/transformation.py
```diff
@@ -0,0 +1,93 @@
...
+        TaskType.TEXT_TO_TEXT_GENERATION
+    ]
+    languages = ["en"]
+    tgt_languages = ["es", "de", "fr", "zh"]
+    heavy = True
+
+    def __init__(self, seed=0, prob_mix=0.3, src_lang="en", trg_lang="de"):
+        super().__init__(seed)
+        
+        self.model = M2M100ForConditionalGeneration.from_pretrained("facebook/m2m100_418M")
```
kaustubhdhole, line 59: One small thing - please cite the paper of this model too in your readme.

gentaiscool: We just added the citation in the README. Thanks for the review! @kaustubhdhole

kaustubhdhole (approved):

kaustubhdhole:
Thank you very much for your changes. We are merging them now. As NLP has largely focussed on high-resource languages and domains, we also encourage you to add a transformation in your mother tongue and/or a low-resource language which might need more attention from the NLP community. The simplest case would be maybe the ButterFingersPerturbation for Bahasa or Malay or another language keyboards. This is not mandatory. 

Thank you again for all your changes!

PR #45, opened 2021-06-19
Evaluation of filters

kaustubhdhole:
1. Introduce robustness evaluation for filters
2. Split services according to task
3. Use dataset.py for both transformation as well as filters


Nickeilf (approved):
Brilliant work! The code structure looks quite decent! Just wondering why the placeholder function for evaluate_mt is in the evaluation_engine.py? I think MT should be one of the tasks to be evaluated and maybe it would be clearer to have another evaluate_machine_translation.py just like those with ner_tagging, text_classification, etc.

kaustubhdhole:
Thanks.. oh yes.. It'd better be in a separate file itself! I will make the change in a subsequent PR!

PR #47, opened 2021-06-20
Added leet_letters transformation

Muennighoff:


transformations/leet_letters/README.md
```diff
@@ -0,0 +1,21 @@
+# Leet Letters 🦎 + 🎮 → 🐍
+This perturbation adds noise to all types of text sources (sentence, paragraph, etc.) proportional to noise erupting 
+from keyboard typos making common spelling errors.
```
kaustubhdhole, line 3: I think you missed out the description.

transformations/leet_letters/transformation.py
```diff
@@ -0,0 +1,55 @@
...
+class LeetLetters(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self):
+        super().__init__()
+
+    def generate(self, sentence: str):
+        perturbed = ""
+        for letter in sentence:
```
kaustubhdhole, line 53: I really like this idea of using leet letters as substitutes. But I think it's better to not convert all the characters in a sentence to Leet - eg.a word like "63t333n" can be hard to figure out - maybe you could control some percentage of them using a configurable parameter.

transformations/leet_letters/transformation.py
```diff
@@ -0,0 +1,55 @@
...
+            perturbed += leet_letter_mappings.get(letter, letter)
+        return perturbed
```
kaustubhdhole, line 55: Also, please pull "main" once so that you can return multiple generations. There will be a minor change in the format test cases.

kaustubhdhole:
Thank you very much for your changes and your PR!

Substituting with leet characters seems to be a wonderful idea to me for NL-Augmenter. The following is my overall review:

**Correctness**: All checks have passed - please pull "main" once into your branch since we added a facility to allow multiple generations. There's a minor change in the syntax of the outputs.
**Interface**: You've chosen the correct interface.
**Specificity**: The change is specific and would be hopefully useful for models relying on Internet slang.
**Novelty**: Visual modifications to text as perturbations has been explored in one recent study: https://arxiv.org/pdf/1903.11508.pdf This should be mentioned in the README. 
**Adding New Libraries**: No new libraries have been added.
**Evaluating Robustness**: Robustness evaluation has not been conducted.

Thank you again for your changes!

transformations/leet_letters/transformation.py
Muennighoff: Good point - Max leet replacements in a sentence is now configurable and set to 50% by default.

Muennighoff:
Applied all suggestions -- thanks!

transformations/leet_letters/transformation.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def generate(self, sentence: str):
+        random.seed(self.seed)
+        max_leet_replacements = int(self.max_leet * len(sentence))
+
+        # Determine what to replace
+        leet_candidates = []
+        for letter in sentence:
+            if letter in leet_letter_mappings:
+                leet_candidates.append(sentence.index(letter), leet_letter_mappings[letter])
```
kaustubhdhole, line 63: >               leet_candidates.append(sentence.index(letter), leet_letter_mappings[letter])
E               TypeError: append() takes exactly one argument (2 given)

Muennighoff: Oops should have been a tuple, my bad - Fixed!

kaustubhdhole:
Hi @Muennighoff thank you very much for incorporating the changes. Could you please pull main once so that your build passes.

Muennighoff:
Done

tuetschek (commented):
The code looks good to me! However, for some reason, the automatic checks don't go through – any idea why?

One more tiny idea for extension – I don't think all of the leet replacements are used in practice with the same frequency. Would it make sense to have a list of "common" and "uncommon" replacements and have the option to turn on just one or both, or use the uncommon less frequently?

tuetschek:
One more thing: Wouldn't `TaskType.TEXT_TAGGING` also apply here?

kaustubhdhole:
Yes - even Tagging would also apply here since the labels/tag indices would remain intact.

kaustubhdhole:
@Muennighoff please fix the build. This seems good to merge then.

Muennighoff:
Alright, just merged upstream, can you approve the workflow? @kaustubhdhole

Muennighoff:
Should be good now @kaustubhdhole

kaustubhdhole (approved):

PR #49, opened 2021-06-20
Add transformation: Whitespace Perturbation

uxyi:


transformations/whitespace_perturbation/transformation.py
```diff
@@ -0,0 +1,33 @@
+import random
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def whitespace(text, remove_prob=0.1, add_prob=0.05, seed=0):
+    random.seed(seed)
```
samsontmr, line 8: `random.seed()` should be called once before each call of `random.random()` for it to work as expected.

samsontmr: It might also be better to generate a range of random integers and use them as seeds for the `random()` calls below so you can introduce some variety into the randomization while still making the randomization reproducible.

transformations/whitespace_perturbation/transformation.py
```diff
@@ -0,0 +1,33 @@
...
+
+    return newtext
+
+
+class WhitespacePerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
```
samsontmr, line 26: Consider adding more languages that are whitespace separated beyond just English.

transformations/whitespace_perturbation/transformation.py
```diff
@@ -0,0 +1,33 @@
+import random
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def whitespace(text, remove_prob=0.1, add_prob=0.05, seed=0):
+    random.seed(seed)
+    newtext = ''
```
samsontmr, line 9: You should use a list to accumulate chars and join at the end to avoid creating a new string at every step.

samsontmr (changes requested):
Thanks for the contribution! Generally looks good, I have left some comments on a few lines.

transformations/whitespace_perturbation/transformation.py
uxyi: I added 4 more languages (`es`, `de`, `fr`, `it`) to the list, and added a discussion about which other languages won’t work in which cases to `README.md`.

transformations/whitespace_perturbation/transformation.py
uxyi: Good catch! I updated the code per your suggestion.

uxyi:
> Thanks for the contribution! Generally looks good, I have left some comments on a few lines.

Hi @samsontmr, thanks for the review! I updated the files accordingly. I also made some changes in correspondence to the recent changes in `master`. Please let me know if you have further questions.

transformations/whitespace_perturbation/transformation.py
```diff
@@ -1,33 +1,37 @@
...
-    newtext = ''
-    for char in text:
-        random_num = random.random()
-        if char.isspace() and random_num < remove_prob:
-            continue
-        newtext += char
-        if (not char.isspace()) and random_num < add_prob:
-            newtext += ' '
+    perturbed_texts = []
+    for _ in itertools.repeat(None, max_outputs):
```
samsontmr, line 11: Why not just use `range(max_outputs)`?

transformations/whitespace_perturbation/transformation.py
```diff
@@ -1,33 +1,37 @@
...
+    return perturbed_texts
 
 
 class WhitespacePerturbation(SentenceOperation):
     tasks = [
         TaskType.TEXT_CLASSIFICATION,
         TaskType.TEXT_TO_TEXT_GENERATION,
     ]
-    languages = ["en"]
+    languages = ["en", "es", "de", "fr", "it"]
```
samsontmr, line 30: I would recommend adding more languages (as many as possible) this transformation supports.

samsontmr: By this I mean "what other languages does this existing transformation apply to beyond `["en", "es", "de", "fr", "it"]`".

transformations/whitespace_perturbation/transformation.py
samsontmr: Oh, just saw this reply! You can have a look at https://github.com/GEM-benchmark/NL-Augmenter/pull/84/files/c9f1e4bdf8fbdacce706c3a6000a42455019c535..c38dd6535fc18b107b3f3bfe924cc370a323df89 as an example

transformations/whitespace_perturbation/transformation.py
uxyi: I agree that `range(max_outputs`) is better and changed it in the latest commit. To answer your question, this particular line of code is actually from the example task [Butter Fingers Perturbation](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/butter_fingers_perturbation/transformation.py).

transformations/whitespace_perturbation/transformation.py
uxyi: This is a very good question. For now, I only put the few languages I at least know some basics and have very high confidence that could work. Unfortunately, I don’t have knowledge in a lot other languages and how they are tokenized usually, and thus could not include those confidently. Do you have any suggestions? Maybe can the organizers provide an exhaustive list of languages of interest, some examples for each language, and tokenizer(s)? If those are provided, people can easily check whether a transformation works for a language by looking at the transformed examples.

samsontmr: In general, I believe languages that use the Latin (including non-European languages like Swahili, Indonesian, Malay, Vietnamese, etc) and Cyrillic scripts (and also Greek) have space-delimited words. Some Latin examples are in this PR: https://github.com/GEM-benchmark/NL-Augmenter/pull/84/files/c9f1e4bdf8fbdacce706c3a6000a42455019c535..c38dd6535fc18b107b3f3bfe924cc370a323df89

@kaustubhdhole might be able to chime in on the South Asian languages

mille-s:
For this perturbation, I think the same holds as for the character swaps that can involve spaces, that is, we'd need to list the tasks/languages for which we should not use it, or at least use it cautiously. For tasks such as PoS tagging or syntactic/semantic parsing for instance, a drop in accuracy will be expected and we can take that into account in the analysis of the results, but I'm not sure about the usefulness of such analysis; the perturbation is not really applicable for D2T tasks. Maybe it's more interesting for text-to-text or text-to-abstract representations tasks?

transformations/whitespace_perturbation/transformation.py
kaustubhdhole: I think most Indian languages would be applicable here : "mr", "hi", "ur", "gj", "ta", "te", "kn", "pa", "kok"

kaustubhdhole:
Hi @uxyi, would you like to address/respond to the above comments?

transformations/whitespace_perturbation/transformation.py
uxyi: @samsontmr @kaustubhdhole Thanks for your inputs! I added more languages to the list.

uxyi: @kaustubhdhole I added the languages you suggested except "gj" and "kok". Are they Gujarat and Konkani? Given that they are not listed in the [ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes), I did not add them yet. Feel free to let me know if I could add those without causing any confusion.

uxyi:
> For this perturbation, I think the same holds as for the character swaps that can involve spaces, that is, we'd need to list the tasks/languages for which we should not use it, or at least use it cautiously. For tasks such as PoS tagging or syntactic/semantic parsing for instance, a drop in accuracy will be expected and we can take that into account in the analysis of the results, but I'm not sure about the usefulness of such analysis; the perturbation is not really applicable for D2T tasks. Maybe it's more interesting for text-to-text or text-to-abstract representations tasks?

Hi @mille-s , thanks for the feedback! For languages, I already listed a few that should use the transformation cautiously (Chinese, Japanese, Thai; for details, see the "limitations" section of the readme). For the tasks, I think the perturbation could still be applicable to the text part of the structured inputs for D2T tasks?

uxyi:
> Hi @uxyi, would you like to address/respond to the above comments?

Yes. See other comments for details :)

transformations/whitespace_perturbation/transformation.py
kaustubhdhole: Okay - that's fine - I just noticed Gujarati is "gu".

mille-s:
@uxyi : thanks for the answer! I'm not sure it makes sense to perturb the reference texts, since these are supposed to be clean in a real-life scenario, and evaluating systems against faulty references may not say much about the quality of the outputs. This perturbation could make sense on the training texts, to see to what extent the models are able to make it up for low-quality training data, but modifying the training data is something we haven't undertaken at this point I think. Definitely something that could be discussed with more people!

transformations/whitespace_perturbation/transformation.py
```diff
@@ -0,0 +1,46 @@
...
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def whitespace(text, remove_prob=0.1, add_prob=0.05, seed=0, max_outputs=1):
+    random.seed(seed)
+    perturbed_texts = []
+    for _ in range(max_outputs):
+        perturbed_text = []
+        for char in text:
```
samsontmr, line 13: Could you move this loop and the outer one to the `generate` function? This is so the whitespace function would operate at the character level and will make it easier for us to attach it to a general adversarial search function. Thanks! Otherwise the PR looks good (to me).

transformations/whitespace_perturbation/transformation.py
uxyi: added "gu". thanks!

transformations/whitespace_perturbation/transformation.py
uxyi: done!

uxyi:
> @uxyi : thanks for the answer! I'm not sure it makes sense to perturb the reference texts, since these are supposed to be clean in a real-life scenario, and evaluating systems against faulty references may not say much about the quality of the outputs. This perturbation could make sense on the training texts, to see to what extent the models are able to make it up for low-quality training data, but modifying the training data is something we haven't undertaken at this point I think. Definitely something that could be discussed with more people!

@mille-s: I see. I think this involves a bigger project-wise decision from the organizers. This transformation and any other transformations affecting the text quality, such as [butter_fingers_perturbation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/butter_fingers_perturbation) and [random_deletion](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/random_deletion), should share the same recommendation for tasks. Any suggestions for next steps on my side?

samsontmr (approved):
Thanks for the contribution!

kaustubhdhole (approved):

PR #51, opened 2021-06-20
Added synonym substitution transformation

zijwang:


transformations/synonym_substitution/test.json
```diff
@@ -0,0 +1,50 @@
...
+        "sentence": "Albany Theatre is a historic theater in Albany , Georgia . It was lend to the national Register of Historic Places on August 21 , 2006 ."
+      }
+    },
+    {
+      "class": "SynonymSubstitution",
+      "inputs": {
+        "sentence": "The Royal Air Force (RAF) is the United Kingdom's aerial warfare force. It was formed towards the end of the First World War on 1 April 1918, becoming the first independent air force in the world, by regrouping the Royal Flying Corps (RFC) and the Royal Naval Air Service (RNAS)."
+      },
+      "outputs": {
+        "sentence": "The Royal Air Force ( RAF ) is the joined Kingdom 's aeriform warfare force . It was formed towards the conclusion of the first World War on 1 April 1918 , become the first independent air effect in the world , by regrouping the royal Flying Corps ( RFC ) and the royal naval Air Service ( RNAS ) ."
```
Nickeilf, line 37: The output seems to be tokenized. Is it possible to detokenize the transformed text so that it would look more similar to the original text?

requirements.txt
```diff
@@ -1,5 +1,7 @@
 checklist==0.0.10
 spacy==2.2.4
+nltk==3.6.2
```
Nickeilf, line 3: It would be good to put the requirements in the transformation folder like did in [this one](https://github.com/GEM-benchmark/NL-Augmenter/blob/48135d4c636b304a0c38d1cc4136db7e600c23b8/transformations/punctuation/requirements.txt). This is to avoid other people's transformation might change the dependency version and might lead to a conflict.

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+        "NOUN": "n",
+        "ADV": "r",
+        "ADJ": "s",
+    }
+    result = []
+    doc = stanza_pipeline(text)
+    for sent in doc.sentences:
+        for word_dict in sent.words:
+            word = word_dict.text
+            wn_pos = upos_wn_dict.get(word_dict.upos)
```
Nickeilf, line 30: Maybe it's better to use the more detailed `xpos` here? I think the current implementation might result in some non-grammatical segments, such as the first test case (returned -> come).
```
{
  "type": "synonym_substitution",
  "test_cases": [
    {
      "class": "SynonymSubstitution",
      "inputs": {
        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
      },
      "outputs": {
        "sentence": "Andrew finally come back the french book to Chris that I bought last workweek"
      }
```

Is it possible to add some sanity check to ensure that the original word and the synonym word have the same POS tag? I think that would make this generated text more natural. One way I can think of is to run some stanza pass on the transformed text with candidate synonym words and check if the POS tags for original/synonym words are the same. Then you could sample from those synonyms that matches the POS tags.

Maybe that would also make this transformation heavier :thinking:

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+            if wn_pos is None:
+                result.append(word)
+            else:
+                syns = wordnet.synsets(word, pos=wn_pos)
+                syns = [syn.name().split(".")[0] for syn in syns]
+                syns = [syn for syn in syns if syn != word]
```
Nickeilf, line 36: I can see you've removed the original word from the list of candidate synonyms. But it seems a bit strange that there're still cases where words being transformed to itself, such as in the last test case (American -> American).
```
    {
      "class": "SynonymSubstitution",
      "inputs": {
        "sentence": "A pisco sour is a cocktail typical of South American cuisine. The drink's name is a combination of the word pisco, which is its base liquor, and the term sour, in reference to sour citrus juice and sweetener components."
      },
      "outputs": {
        "sentence": "A pisco sour is a cocktail distinctive of South american cuisine . The drink 's name is a combination of the word pisco , which is its base liquor , and the terminus sourness , in character to sour citrus juice and sweetener components ."
      }
    }
  ]
} 
```
Is it because the synonyms in wordnet are case-sensitive? Or it might be caused by the stanza tokenization (when `random.random()<prob` and `result.append(word)` but the `word` here is lower-cased)?

Nickeilf (commented):
Hi @zijwang, thanks for this interesting transformation!

I've left some comments in the code, please take a look for some potential suggestions. Here's my general review:
**Specificity**: The change is specific to text classification and generation tasks.
**Novelty**: This transformation is not implemented in NL-Augmenter.
**Adding New Libraries**: NLTK and stanza are added.
**Evaluating Robustness**: Robustness evaluation is conducted on both text classification and generation tasks.

transformations/synonym_substitution/test.json
```diff
@@ -0,0 +1,50 @@
+{
+  "type": "synonym_substitution",
+  "test_cases": [
+    {
+      "class": "SynonymSubstitution",
+      "inputs": {
+        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
+      },
+      "outputs": {
```
aadesh11, line 9: Please take a master pull, outputs should list of dict.

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+
+class SynonymSubstitution(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=42, prob=0.5):
+        super().__init__(seed)
```
aadesh11, line 57: The constructor should also take 'max_outputs' as one of the params, based on which generate function should return perturbed sentences.

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+        self.stanza_pipeline = stanza.Pipeline("en", processors="tokenize,mwt,pos")
+        self.prob = prob
+
+    def generate(self, sentence: str):
+        pertubed = synonym_substitution(
```
aadesh11, line 62: Minor cosmetic change: 
pertubed -> perturbed

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+            text=sentence,
+            stanza_pipeline=self.stanza_pipeline,
+            seed=self.seed,
+            prob=self.prob,
+        )
+        return pertubed
```
aadesh11, line 68: Should be a list of perturbed sentences.

transformations/synonym_substitution/test.json
zijwang: Updated.

transformations/synonym_substitution/transformation.py
zijwang: Updated.

transformations/synonym_substitution/transformation.py
zijwang: Updated.

transformations/synonym_substitution/transformation.py
zijwang: Updated.

transformations/synonym_substitution/test.json
zijwang: Good point! I added detokenization into the code base and it should look better now.

requirements.txt
zijwang: Updated.

transformations/synonym_substitution/transformation.py
zijwang: Good point. The bottomneck here is the synonyms from NLTK/worknet, which do not come with XPOS labels. I think it is doable to do a second round check, but as you mentioned that may make things more complicated, and more importantly, may introduce additional noises as no POS tagger is perfect. Currently, the synonyms will not be substituted if the UPOS doesn't match, which, in most of the cases, shouldn't affect human understanding of the sentences. How do you think about it?

transformations/synonym_substitution/transformation.py
zijwang: Yeah this is due to the nltk synonyms, which are case-insensitive. For example,

```
>>> syns = wordnet.synsets('American')
>>> syns
[Synset('american.n.01'),
 Synset('american_english.n.01'),
 Synset('american.n.03'),
 Synset('american.a.01'),
 Synset('american.a.02')]
```

zijwang:
Hey @Nickeilf and @aadesh11 ! Thanks a lot for the detailed review. I have updated the code and replied to the comments. Could you take a look and let me know how you think about these changes?

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,98 @@
...
+def synonym_substitution(text, stanza_pipeline, seed=42, prob=0.5, max_outputs=1):
+    random.seed(seed)
+    upos_wn_dict = {
+        "VERB": "v",
+        "NOUN": "n",
+        "ADV": "r",
+        "ADJ": "s",
+    }
+
+    doc = stanza_pipeline(text)
```
kaustubhdhole, line 47: I would suggest you to use spacy since its already being used everywhere in NL-Augmenter unless stanza is required for anything unique.

transformations/synonym_substitution/transformation.py
Nickeilf: It looks fine to me. Maybe you could mention in the readme file that this transformation might introduce non-grammatical segments.

transformations/synonym_substitution/transformation.py
zijwang: Updated. Thanks for the suggestion!

transformations/synonym_substitution/transformation.py
zijwang: Added. Thanks @Nickeilf !

transformations/synonym_substitution/transformation.py
Nickeilf: I see. Would it be good to reject the same word? Because if only one word is selected and replaced with the same word, it does not actually change the sentence.

zijwang: Arguably upper/lower case also changes the semantic meaning of a word. But to disentangle the effect of casing, I have updated the code to make sure that the exact match check is independent of casing.

Nickeilf (approved):
Thanks for the changes. Now it looks good to me. I've passed the review

transformations/synonym_substitution/transformation.py
```diff
@@ -0,0 +1,104 @@
...
+import re
+
+import nltk
+import spacy
+from nltk.corpus import wordnet
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+nltk.download("wordnet")
```
kaustubhdhole, line 11: This line should be in the constructor.

transformations/synonym_substitution/requirements.txt
```diff
@@ -0,0 +1,2 @@
+nltk==3.6.2
+stanza==1.2.1
```
kaustubhdhole, line 2: Is Stanza being used?

kaustubhdhole (approved):

PR #55, opened 2021-06-22
Contractions/Expansions transformation

tanay2001:


transformations/contraction_expansions/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+        Parameters
+        ----------
+        sentence : str
+            input string
+        Returns
+        -------
+        string
+            String with contractions contracted (if any)
+        """
+        reverse_contraction_map = {
```
Nickeilf, line 28: I would suggest moving this map to the `__init__()` function so that it wouldn't get loaded everytime the `generate()` function is called.

Nickeilf: Another question: Is this `reverse_contraction_map` the same as the `contraction_map` but just reversing the key-value?

If yes, I think you can simply do a dict comprehension rather than hard-coding it twice.
```
contraction_map = {.....}
reverse_contraction_map = {value:key for key, value in contraction_map.items()}
```

transformations/contraction_expansions/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+        -------
+        list
+            List of strings with contractions expanded or contracted, or []
+        """
+        expanded = [self.expand_contractions(sentence), self.contract(sentence)]
+        return [t for t in expanded if t != sentence]
+
+    def generate(self, sentence):
+        pertubed = self.contractions(sentence)[0]
+        return pertubed
```
Nickeilf, line 227: Please pull the main into your branch.

The returned texts should be a list now. You can either wrap up the perturbed, like `[perturbed]`, or returning multiple outputs in a list (e.g. both expanded and contracted version).

Don't forget to update your test.json accordingly.

Nickeilf (commented):
Hi @tanay2001, thanks for this interesting transformation!

**Correctness**: pytest not passed - please pull "main" once into your branch and update the test.json file, we have updated the output format to list of texts (can be list of only one text).
**Specificity**: The transformation is specific to texts with contractions or expansions.
**Novelty**: This transformation is novel and NL-Augmenter has no similar implementation.

tanay2001:
> Hi @tanay2001, thanks for this interesting transformation!
> 
> I've left some comments in the code and here's my general review:
> **Correctness**: pytest not passed - please pull "main" once into your branch and update the test.json file, we have updated the output format to list of texts (can be list of only one text).
> **Interface**: I think you've chosen the correct interface.
> **Specificity**: The transformation is specific to texts with contractions or expansions.
> **Novelty**: This transformation is novel and NL-Augmenter has no similar implementation.
> **Adding New Libraries**: No new library is added.
> **Test Cases**: 5 test cases have been added.
> **Evaluating Robustness**: Robustness evaluation is not conducted.

Thanks for the review @Nickeilf 
I have made the changes in the code as suggested

transformations/contraction_expansions/transformation.py
```diff
@@ -0,0 +1,171 @@
...
+        ----------
+        sentence : str
+            input
+        Returns
+        -------
+        list
+            List of strings with contractions expanded or contracted, or []
+        """
+        expanded = [self.expand_contractions(sentence), self.contract(sentence)]
+        return [t for t in expanded if t != sentence]
```
Nickeilf, line 167: One more question, if the input sentence contains both expanded and contracted segments, it would return both fully expanded and contracted sentences, right?

For example:
Input: "He is a doctor, but she's a teacher"
The output would be: ["He is a doctor, but she is a teacher", "He's a doctor, but she's a teacher"]

Nickeilf: Also, I think maybe it's fine to return the original sentence rather than an empty list when you **cannot** find any expanded or contracted phrases. Because one might want to apply this transformation to the dataset and the current implementation might result in empty inputs in the transformed data.

I guess the best way is to remove the original sentence if there **is** a successful expansion or contraction. Otherwise, you can just return the unchanged input.

tanay2001: > One more question, if the input sentence contains both expanded and contracted segments, it would return both fully expanded and contracted sentences, right?
> 
> For example:
> Input: "He is a doctor, but she's a teacher"
> The output would be: ["He is a doctor, but she is a teacher", "He's a doctor, but she's a teacher"]

Yes it will, have added another test case to clear things out

tanay2001: > Also, I think maybe it's fine to return the original sentence rather than an empty list when you **cannot** find any expanded or contracted phrases. Because one might want to apply this transformation to the dataset and the current implementation might result in empty inputs in the transformed data.
> 
> I guess the best way is to remove the original sentence if there **is** a successful expansion or contraction. Otherwise, you can just return the unchanged input.

Thanks for pointing this out, I wasn't very sure about this case , will update the code

Nickeilf (approved):
Thanks for the changes! I've approved the review now.

kaustubhdhole (approved):

PR #60, opened 2021-06-22
Added date format transformation

pasricha:
Transformation to change the format of dates appearing in text. I have followed a similar approach as the [replace_numerical_values](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/replace_numerical_values) transformation to change the date format.

Right now only English is supported with different locale formats using Babel, but this could be extended to support other locales/languages too.


transformations/change_date_format/transformation.py
```diff
@@ -0,0 +1,164 @@
...
+            if entity.label_ == "DATE":
+                date, has_year, has_month, has_day = self.parse_date(entity.text)
+
+                if date:
+                    if has_year and has_month and has_day:
+                        new_value = format_date(
+                            date,
+                            format=random.choice(self.ymd_formats),
+                            locale=random.choice(self.locales),
+                        )
```
kaustubhdhole, line 98: Doesn't random.seed need to be called before every random.choice?

transformations/change_date_format/transformation.py
```diff
@@ -0,0 +1,164 @@
...
+        random.seed(self.seed)
+        super().__init__()
+        self.date_format_transformation = DateFormatTransformation()
+        self.max_output = max_output
+
+    def generate(self, sentence: str):
+        result = self.date_format_transformation.transform(sentence)
+        if self.verbose:
+            print(f"Perturbed Input from {self.name()} : {result}")
+        return [result]
```
kaustubhdhole, line 137: Maybe you can make use of max_output and generate multiple dates too for a single sentence?

transformations/change_date_format/transformation.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# Sample code to demonstrate usage.
```
kaustubhdhole, line 140: This can be deleted or commented out.

transformations/change_date_format/transformation.py
kaustubhdhole: Also, you could add more variety by passing format as “full”, “long”, “medium”, or “short”, besides a custom date/time pattern?

kaustubhdhole:
Hi @pasricha , thank you very much for your changes! Here is my overall review:

Correctness: I think the tests are not passing because of the seed.
Interface: The correct interface has been chosen. 
Specificity: The change is very specific. 
Adding New Libraries: The new libraries should be added in change_date_format/requirements.txt
Test Cases: 3 cases have been added.
Evaluating Robustness: No evaluation has been conducted.

pasricha:
Thanks @kaustubhdhole for the review. I will make the changes you suggested and add an update here when I am done.

transformations/change_date_format/transformation.py
pasricha: The "short/medium/long" options are useful only if the text contains all three fields - year, month and day. If any of the fields is missing in the original text, then by default it will be filled with the current value in the replacement. For example, if a text contains "March 2020", then it would be replaced as "22 March 2020" or "23 March 2020" depending on whether the current day is 22 or 23. Similarly, something like "December 31" which doesn't have the year specified  will be replaced as "December 31, 2021". Since this transformation is only intended to modify the format of the date, and not the actual date itself, I used custom date patterns to match whatever the original text contains so as to not introduce incorrect information.

kaustubhdhole (approved):

kaustubhdhole:
I will fix the test cases during merge. Thanks for your changes @pasricha :)

PR #65, opened 2021-06-23
Token amount filter

SirRob1997:
## Token-Amount filter

Author: Robin M. Schmidt (https://scholar.google.de/citations?user=20vb63kAAAAJ&hl=de)

## What type of a filter is this?

This filter filters an example text based on a set of keywords and returns whether the supported operations are true for the input text. It is a more powerful version of the already existing `TextContainsKeywordsFilter` and should contain its use-cases as a subset.

Inputs:

- Array of words (e.g. ["is", "are"]), a single string is also possible
- Array of integers of elementwise required occurrences (e.g. [2, 4]), if a single integer is given the same threshold is applied to all keywords
- Array of strings of comparison operations (e.g. ["<", ">"]), if a single string is given the same operation is applied to all keywords

Supported operations:
- greater than: ">"
- less than: "<"
- greater equal to: ">="
- less equal to: "<="
- equal to: "=="


SirRob1997:
There should probably be another label called "filter" to quickly check in the PR's which transformations/filters have already been implemented. This is a filter and should therefore have the new tag.

filters/token-amount/filter.py
```diff
@@ -0,0 +1,77 @@
+import operator
+import spacy
+
+from interfaces.SentenceOperation import SentenceOperation, SentenceAndTargetOperation
```
Nickeilf, line 4: The `SentenceAndTargetOperation` seems to be not used here.

filters/token-amount/filter.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class TokenAmountFilter(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(
+        self,
+        keywords: Union[list, str] = None,
+        thresholds: Union[list, str] = None,
```
Nickeilf, line 21: The `thresholds` should be either list or int, right?

filters/token-amount/filter.py
```diff
@@ -0,0 +1,77 @@
...
+        all_inputs = [keywords, thresholds, operations]
+        lengths = defaultdict(int)
+        for user_input in all_inputs:
+            if isinstance(user_input, list):
+                lengths[len(user_input)] += 1
+
+        if len(list(lengths.keys())) > 1:
+            raise ValueError("One or more lists given with non-matching lengths")
+
+        return list(lengths.keys())[0]
```
Nickeilf, line 41: What if `keywords`, `thresholds`, and `operations` are all not lists?

filters/token-amount/filter.py
```diff
@@ -0,0 +1,77 @@
...
+        self.max_input_length = self.get_input_length(keywords, thresholds, operations)
+        self.final_operators = self.parse_operator(operations)
+        self.final_keywords = self.convert_scalar_to_list(keywords)
+        self.final_thresholds = self.convert_scalar_to_list(thresholds)
+        self.nlp = spacy.load("en_core_web_sm")
+
+    def get_input_length(self, keywords, thresholds, operations):
+        all_inputs = [keywords, thresholds, operations]
+        lengths = defaultdict(int)
+        for user_input in all_inputs:
```
Nickeilf, line 34: This implementation seems a bit strange to me. Maybe you could just specify the length explicitly to the `lengths` dict and then do some assert for sanity check.
```
for key, user_input in zip(['keywords', 'thresholds', 'operations'], all inputs):
    if isinstance(keywords, list):
        lengths[key] = len(user_input)
    else:
        lengths[key] = 1
```

filters/token-amount/test.json
```diff
@@ -0,0 +1,89 @@
...
+            },
+            "inputs": {
+                "sentence": "Andrew played cricket in a soccer stadium in India at 9pm"
+            },
+            "outputs": true
+        },
+		{
+            "class": "TokenAmountFilter",
+            "args": {
+                "keywords": "in",
```
Nickeilf, line 79: Maybe you can do a sanity check for only one keyword with multiple conditions (to check whether there's any contradiction). For example:
thresholds: [3, 1]
operations: [">", "<"]
It would be good to warn the user that the input thresholds/operations would cause a conflict than simply returning a false.

Nickeilf (commented):
Hi @SirRob1997, thanks for contributing this filter!

**Correctness**: I've noticed some potential places to improve in the code, please take a look at my comments  - also don't forget to pull "main" once into your branch later.
**Specificity**: The filter is not specified to certain types of texts.
**Novelty**: This filter is novel in implementation and supports multiple conditions and flexible inputs (threshold/operations).
**Test Cases**: 5+ test cases have been added.
**Evaluating Robustness**: Robustness evaluation is not conducted. You can evaluate your filter with `python evaluate.py -f TokenAmountFilter -task TEXT_CLASSIFICATION -p 10`. If you do the evaluation, don't forget to update the README and report the results

filters/token-amount/filter.py
SirRob1997: good catch!

filters/token-amount/test.json
SirRob1997: I gave it a shot implementing this, not sure if there is a more clean way than just manually constructing the bounds for each keyword but I've tested it with your example as well as:

thresholds: [3, 1]
operations: ["==", "=="]

and raise appropriate ValueErrors for each. Seems like it's working as intended. Let me know in case you have any improvements for the new part of the code.

filters/token-amount/filter.py
```diff
@@ -0,0 +1,123 @@
...
+        for key, user_input in zip(
+            ["keywords", "thresholds", "operations"], all_inputs
+        ):
+            if isinstance(user_input, list):
+                lengths[key] = len(user_input)
+            else:
+                lengths[key] = 1
+
+        unique_lengths = set(lengths.values())
+        if len(unique_lengths) == 1:
```
Nickeilf, line 44: I guess this is the case for inputs with all non-list, right?

The if clause should be `unique_lengths == {1}`, because if the three inputs are **all** list with the same length, len(unique_lengths) will be equal to 1 as well.

Nickeilf: It seems this issue does not affect the filter function, and all pytest are passed.

It would be good to fix it though :smile:.

SirRob1997: ha! yes indeed, good catch again.

filters/token-amount/filter.py
```diff
@@ -0,0 +1,123 @@
...
+            else [pos_scalar for _ in range(self.max_input_length)]
+        )
+
+    def parse_operator(self, op):
+        ops = {
+            ">": operator.gt,
+            "<": operator.lt,
+            ">=": operator.ge,
+            "<=": operator.le,
+            "==": operator.eq,
```
Saad-Mahamood, line 64: I wonder if it would make practical sense to have a `!=` operator here as well for integer values?

Saad-Mahamood (commented):
Looks good to me. I do have a query whether this filter can be improved by including a `!=` operator. Would be happy to approve once this query has been answered.

filters/token-amount/filter.py
SirRob1997: Hmm, for a condition with `!= 0` we could just use `> 0` instead, and for any other natural number `n` we can always have two conditions where `> n` and `< n`. I feel like having `!=` as an option would just enable more human errors, for example, in the case with `!= 1` one might forget that this doesn't filter sentences with `0` occurrences. Explicitly typing out the two cases might help to prevent that.

Nickeilf (approved):

filters/token-amount/filter.py
Saad-Mahamood: No, that is a fairly reasonable explanation.

Saad-Mahamood (approved):

kaustubhdhole:

PR #66, opened 2021-06-23
Speech tag filter

SirRob1997:
## Speech-Tag filter

Author: Robin M. Schmidt (https://scholar.google.de/citations?user=20vb63kAAAAJ&hl=de)

## What type of a filter is this?

This filter filters an example text based on a set of speech tags and returns whether the supported operations are true for the input text.

Example Speech Tags (see spacy framework):

- NOUN
- VERB
- PROPN
- AUX
- ADP
- SYM
- NUM
- CCONJ
- DET
- PRON

Inputs:

- Array of speech tags (e.g. ["NOUN", "VERB"]), a single string is also possible
- Array of integers of elementwise required occurrences (e.g. [2, 4]), if a single integer is given the same threshold is applied to all speech tags
- Array of strings of comparison operations (e.g. ["<", ">"]), if a single string is given the same operation is applied to all speech tags
- Optional: percentages flag to change the occurrences to be treated as percentages

Supported operations:

- greater than: ">"
- less than: "<"
- greater equal to: ">="
- less equal to: "<="
- equal to: "=="


kaustubhdhole:
Hi @SirRob1997 thank you very much for your changes. I think it is useful to have a filter which does a POS analysis on the test. I will add comments inline!

filters/speech-tag/filter.py
```diff
@@ -0,0 +1,87 @@
...
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(
+        self,
+        speech_tags: Union[list, str],
+        thresholds: Union[list, str],
+        operations: Union[list, str],
+        percentages: bool = False,
+    ):
```
kaustubhdhole, line 24: It might be helpful to add default arguments here. When all filters would be called together - it might not be feasible to provide separate arguments for each.

SirRob1997: What would you recommend for the defaults? I was considering adding some, but decided to only add one for the `percentages` flag since I couldn't think of a default case for the others.

kaustubhdhole: It does not have to be symbolic of any task - just to have placeholders. But I guess we could expect performance drops in longer sentences in general - so maybe verb & noun counts of more than 2 or 3 each?

kaustubhdhole:
Correctness: The code is nicely written. The build needs to pass though.
Interface & TaskTypes: The right ones have been chosen. 
Specificity: The code is specific to filter sentences with the provided amount of POS tags. 
Novelty: It has not been implemented in NL-Augmenter before.
README: Probably papers where such analysis was useful in the past could be added. If you know work which has a difference in performance for say - sentences with 2 verbs vs sentences with 4 verbs, they could be mentioned.
Evaluating Robustness: Robustness evaluation has been conducted.

Overall, this is a very useful PR. If you can make the changes too, that'll be great! :)

SirRob1997:
It seems like the build is failing while installing dependencies, although I didn't introduce any new ones. Do you have any idea why this is happening?

kaustubhdhole (approved):

SirRob1997:
@Nickeilf or @aadesh11 can you take a look as well and see if I need to change anything?

filters/speech-tag/filter.py
```diff
@@ -0,0 +1,133 @@
...
+        super().__init__()
+        self.max_input_length = self.get_input_length(
+            speech_tags, thresholds, operations
+        )
+        self.final_operators = self.parse_operator(operations)
+        self.final_speech_tags = self.convert_scalar_to_list(speech_tags)
+        self.final_thresholds = self.convert_scalar_to_list(thresholds)
+        self.nlp = spacy.load("en_core_web_sm")
```
aadesh11, line 32: For optimization, we are loading spacy only once in 'initialize.py'. Please change this line with:

You can refer to other filters like 'keywords', 'length'.

SirRob1997: done

PR #71, opened 2021-06-24
Numeric to word

bryanwilie:
This transformation translates numeric string in major formats in the text into its words form. It can be used to introduce the number as a character token to the model.

This transformation translates numbers in numeric form form written amongst the texts for:
- General numbers
    - general number (commas, thousands)
    - sticky numbers (2x, 5th, 8pm, 10%)
    - negatives (-0.5)
    - percentage
    - natural log (e(9))
    - fraction
    - long number
    - long number with stripes
    - sticky ranges ( (1-2) )
    - range not sticky ( ( 1-2 ) )
    - math formula equality
    - mathematical power of ten (10-4)
    - numeric in begin end bracket ( (34) )
    - numeric beside end bracket ( 34) )
    - numeric in math_bracket
    - special numbers (911)
    - currency (with the currency symbols and cents)
    - and more ..
- Datetime
    - incomplete_date
        - '01/2020'
        - '2020/01'
        - '20/01'
        - '01/20'
    - complete date
    - year
    - yime
    - date time
- Phone Number
    - general phone number
    - special phone number ('*#')
    
This transformation also translates the cases existed in test samples taken from:
- [SemEval 2019 Task 10: Math Question Answering](https://www.aclweb.org/anthology/S19-2153.pdf)
- [ChemistryQA](https://openreview.net/pdf?id=oeHTRAehiFF)
- [PubmedQA](https://www.aclweb.org/anthology/D19-1259.pdf)
- [SMD / KVRET](https://www.aclweb.org/anthology/2020.findings-emnlp.215/)
- [Mathematics Dataset](https://openreview.net/pdf?id=H1gR5iR5FX)
- [PubMed 200k RCT](https://www.aclweb.org/anthology/I17-2052.pdf)
- [BBC News](https://www.kaggle.com/c/learn-ai-bbc)

Authors: Bryan Wilie, Genta Indra Winata, Samuel Cahyawijaya.


mille-s:
I like this perturbation, there's a significant amount of cases to be covered, as show the number of examples mentioned by @bryanwilie, and it's a typical phenomenon that gives problems in current systems. For instance, TTS systems I've been using seem to have strong preferences about the number formatting, and if the format is not good the system is just not able to produce natural sequences. It will also be quite challenging for analysis tasks, and I expect this perturbation to be quite annoying in general, so thumbs up! One thing is that there may be so many different perturbations and ways to compromise the data for each perturbation that we may think of some kind of classification/taxonomy to allow for an easy overview of what's being done and implemented.

transformations/numeric_to_word/requirements.txt
```diff
@@ -0,0 +1,3 @@
+phonenumbers==8.12.25
+docopt==0.6.2
+Unidecode==1.2.0
```
kaustubhdhole, line 3: This library can't be integrated. Can you please find a good substitute?

transformations/numeric_to_word/test.json
```diff
@@ -0,0 +1 @@
...e not ruling out a late rally from former happy mondays star bez"}]}, {"class": "NumericToWord", "inputs": {"sentence": "moya fights back for indian title carlos moya became the first man to successfully defend the chennai open title by beating four-times finalist paradorn srichaphan 3-6 6-4 7-6 (7/5)"}, "outputs": [{"sentence": "moya fights back for indian title carlos moya became the first man to successfully defend the chennai open title by beating four-times finalist paradorn srichaphan three to six six to four seven to six (seven over five)"}]}, {"class": "NumericToWord", "inputs": {"sentence": "last week  mci agreed to a buyout by verizon communications in a deal valued at $6.75bn.,business"}, "outputs": [{"sentence": "last week mci agreed to a buyout by verizon communications in a deal valued at six point seven five bn. dollar business"}]}, {"class": "NumericToWord", "inputs": {"sentence": "the sequel took $44.7m (\u00a323.2m) between 24 and 26 december  according to studio estimates"}, "outputs": [{"sentence": "the sequel took forty-four point seven m dollar (twenty-three point two m pound) between twenty-four and twenty-sixth of december according to studio estimates"}]}]}
```
kaustubhdhole, line 1: Also, please format the json - If you are using pycharm - you can -select all and do CTRL+ALT+SHIFT+L :)

transformations/numeric_to_word/requirements.txt
bryanwilie: Will try and update here asap @kaustubhdhole. Thank you for the confirmation

transformations/numeric_to_word/test.json
bryanwilie: Will try and fix, thank you @kaustubhdhole

bryanwilie:
Hi @kaustubhdhole,

I searched through the issue with GPL library, and I found [this issue here](https://github.com/wagtail/wagtail/issues/3311) faced the same issue that I think you're anticipating.

Some of the people there recommends [anyascii](https://pypi.org/project/anyascii/), but the licence is ISC. Is it okay if the substitute is in ISC licence? This is a good substitute I believe if the licence is compatible.
> The ISC licence seems to be the same as the MIT with some changes in language so it should be fine for using it in any kind of project!

I'll update the code and dependency nonetheless, but I'll still be waiting for your feedback.

Thanks @kaustubhdhole !

bryanwilie:
Hi @kaustubhdhole,

I searched through the issue with GPL library, and I found [this issue here](https://github.com/wagtail/wagtail/issues/3311) faced the same issue that I think you're anticipating.

Some of the people there recommends [anyascii](https://pypi.org/project/anyascii/), but the licence is ISC. Is it okay if the substitute is in ISC licence? This is not a good substitute, as it is not catering some of the cases e.g. `'×'== 'x'`(notice the left x is a multiplication, not alphabet x), but this is the closest substitute I can find right now.
> The ISC licence seems to be the same as the MIT with some changes in language so it should be fine for using it in any kind of project!

 I'll be waiting for your feedback.

Thanks @kaustubhdhole

bryanwilie:
Hi @kaustubhdhole,

I searched through the issue with GPL library, and I found [this issue here](https://github.com/wagtail/wagtail/issues/3311) faced the same issue that I think you're anticipating.

Some of the people there recommends [anyascii](https://pypi.org/project/anyascii/) as the closest substitute (not 100% similar), but the licence is ISC and this is not a good substitute, as it is not catering some of the cases e.g. `'×'== 'x'`(notice the left x is a multiplication, not alphabet x), but this is the closest substitute I can find right now. 

I think that's the best that we could come up right now, but both libraries is not the same.

I've removed the dependency to `unidecode`, as the transformation in its current state only needs `unidecode` for mathematical multiplication sign and minus sign.

I've updated this PR for your review again.

Thanks @kaustubhdhole

kaustubhdhole:
Yes @bryanwilie, it should be fine to use anyascii if it uses the ISC license. Thanks for confirming.

bryanwilie:
I ended up not using `anyascii`, because I hardcoded some of the mathematical expressions in the code, thus I also removed the dependency to `unidecode`.

Thank you @kaustubhdhole, it's my pleasure to be able to contribute 👍🏻

transformations/numeric_to_word/numeric2word.py
```diff
@@ -0,0 +1,1025 @@
...
+### Implementations 
+
+def recognize_transform(word, prev_word, next_word):
+    
+    # transforming by looking at previous or next words
+    if recognized_as_power_of_ten(word, prev_word):
+        remaining_word = word[2:]
+        remaining_word = '-' + remaining_word[1:] if remaining_word[0] == '−' else remaining_word
+        words = 'ten power ' + num2words(remaining_word)
+#         print('1', word, words)
```
aadesh11, line 31: Please remove all the commented print statements, better not to have that in master.

transformations/numeric_to_word/numeric2word.py
```diff
@@ -0,0 +1,1025 @@
...
+    word_list = recognize_transform(x[:end_digit_index], '', '').split()
+    
+    m_or_bn = x[end_digit_index:].split(',')[0]
+    sticky_following = x[end_digit_index:].split(',')[1:]
+    currency = [word_list[-1]]
+    
+    words = ' '.join(word_list[:-1] + [m_or_bn] + currency + sticky_following)
+    return words
+    
+symbol_to_currency_name_dict = {"؋":"Afghanistani Afghani",
```
aadesh11, line 744: Just for clean code, create a constant file and move all the constants to that file instead of putting it in a single file.

transformations/numeric_to_word/transformation.py
```diff
@@ -0,0 +1,59 @@
+import random
+import string
+from .numeric2word import recognize_transform
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+class NumericToWord(SentenceOperation):
```
aadesh11, line 8: For good readability, better to add few lines of comments for this transformation here.

bryanwilie: done, added docstring from this PR message

transformations/numeric_to_word/numeric2word.py
bryanwilie: done

transformations/numeric_to_word/numeric2word.py
bryanwilie: done, sorry for that

transformations/numeric_to_word/abbreviated_currency_symbols_to_currency_name_dict.py
```diff
@@ -0,0 +1,151 @@
+abbreviated_currency_symbols_to_currency_name_dict = {'AFN': 'Afghanistani Afghani',
```
aadesh11, line 1: Better to have one file for constants, it would be great if we can put both 'abbreviated_currency_symbols_to_currency_name_dict' and 'symbol_to_currency_name_dict' in single file.

bryanwilie: Yes you're correct. My bad on those. I've put both constants in single file, `constants.py`. Thanks for the feedback @aadesh11!

kaustubhdhole:
Thank you very much to everyone. I will merge this now.

kaustubhdhole (approved):

PR #72, opened 2021-06-24
Add Repetitions filter

gentaiscool:
This module filters examples which contain repetitions. This filter can be helpful to find text with repetitions that frequently occur in the spoken language data.

Author: Genta Indra Winata

### DIFF
--- a/filters/repetitions/README.md
+++ b/filters/repetitions/README.md
@@ -1,0 +1,9 @@
+## Repetitions filter
+
+## What type of a filter is this?
+
+This module filters examples which contain repeitions.
+Author: Genta Indra Winata (giwinata@connect.ust.hk)
+
+## Use Cases
+This filter can be useful to find natural language with a lot of repetitions that frequently occurs in the spoken language data.
--- a/filters/repetitions/README.md
+++ b/filters/repetitions/README.md
@@ -1,9 +1,9 @@
 ## Repetitions filter
 
 ## What type of a filter is this?
+This module filters examples that contain repetitions. 
 
-This module filters examples which contain repeitions.
 Author: Genta Indra Winata (giwinata@connect.ust.hk)
 
 ## Use Cases
-This filter can be useful to find natural language with a lot of repetitions that frequently occurs in the spoken language data.
+This filter can be helpful to find text with repetitions that frequently occur in the spoken language data.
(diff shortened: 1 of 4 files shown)


mille-s:
That's an interesting subpopulation that can be very useful for analysis of noisy texts (coming from speech or forums for instance). I'm only concerned about being able to distinguish between erroneous ("I I like cats") and intentional ("I have many many cats") repetitions. Is there a way to assess the "erroneousness" of repetitions? Or don't we want to get into this kind of considerations yet?

gentaiscool:
> That's an interesting subpopulation that can be very useful for analysis of noisy texts (coming from speech or forums for instance). I'm only concerned about being able to distinguish between erroneous ("I I like cats") and intentional ("I have many many cats") repetitions. Is there a way to assess the "erroneousness" of repetitions? Or don't we want to get into this kind of considerations yet?

Hi @mille-s, thanks for the insightful review. I also consider this. However, we need to apply heuristics/rules to classify if the duplicates are generated intentionally or not. And, it might not be straightforward to apply it to any text since the phenomenon happened more in spoken language. But, for sure, it is a useful addition to this filter.

kaustubhdhole:
Okay, I think you should definitely mention one limitation of not being able to differentiate from sentences with "[epizeuxis](https://en.wikipedia.org/wiki/Epizeuxis)" 
Here is the wikipedia definition: In rhetoric, epizeuxis is the repetition of a word or phrase in immediate succession, typically within the same sentence, for vehemence or emphasis. Found this paper used to detect this epizeuxis: https://esnbu.org/data/files/2020/esnbu.20.2.3.pdf

kaustubhdhole:
@gentaiscool

gentaiscool:
> Okay, I think you should definitely mention one limitation of not being able to differentiate from sentences with "[epizeuxis](https://en.wikipedia.org/wiki/Epizeuxis)"
> Here is the wikipedia definition: In rhetoric, epizeuxis is the repetition of a word or phrase in immediate succession, typically within the same sentence, for vehemence or emphasis. Found this paper used to detect this epizeuxis: https://esnbu.org/data/files/2020/esnbu.20.2.3.pdf

Hi, @kaustubhdhole thanks for the comments and insightful suggestion. Sorry for the late reply. I would add the limitation and check the paper you refer to.

kaustubhdhole (approved):

Issue #75, opened 2021-06-26
Informal & Untested Suggestions for Possible Transformations


kaustubhdhole:
Here are some random ideas informally put which could be used for perturbations & augmentations. @vgtomahawk is making a formal list in [this](https://github.com/GEM-benchmark/NL-Augmenter/tree/vgtomahawk-startGuidelineDoc-1) PR. 

Meanwhile here is an informal list for the benefit of the participants. 

1) Interchange positions of SRL AM arguments for non-overlapping AM arguments: 
   - Alex left for Delhi with his wife at 5 pm. --> Alex left for Delhi at 5 pm with his wife.
   - "at 5 pm" (AM-TMP) and "with his wife" (AM-COM) can be exchanged: This is safe to do only with non-core arguments and non-overlapping arguments. Check what SRL is [here](https://demo.allennlp.org/semantic-role-labeling).

2) The ButterFingersPertubation could be implemented for keyboard types other than English - like Devanagiri (Hindi, Marathi, Nepail), Shahmukhi (Urdu, Persian), South Indian languages (Tamil, Telugu, Kannada, Malayalam) or Chinese, etc.

3) Style transfer approaches could be interesting to look at - Changing formal to informal and vice versa. [Check this model](https://huggingface.co/prithivida/informal_to_formal_styletransfer).
 - What the heck is going on?  --> What is going on?
 - What you upto? --> What are you doing? 

4) Word Order Changes: Active to Passive & vice versa, [Topicalisation](https://en.wikipedia.org/wiki/Topicalization), Extraposition, Wh-fronting, (& vice versa) & other used in constituency tests.
[Scrambling](https://en.wikipedia.org/wiki/Scrambling_(linguistics)) (for German, Turkic languages)
John went to the store to buy bread. --> To buy bread, John went to the store.

The above are only related to [SentenceOperation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/interfaces). There are other transformation types too which could be looked at.

kaustubhdhole:
[Adversarial SQUAD](https://arxiv.org/pdf/1707.07328.pdf) adds wrong but similar facts at the end of the context in a QuestionAnswerTransformation.

kaustubhdhole:
These two surveys provide a great overview of previous approaches - This is a great place to look for ideas:
https://github.com/AgaMiko/data-augmentation-review
https://arxiv.org/pdf/2105.03075.pdf

[...]

PR #77, opened 2021-06-27
added formal2casual

rg089:
This transformation transfers style from formal to casual.

### DIFF
--- a/transformations/casual2formal/README.md
+++ b/transformations/casual2formal/README.md
@@ -12,4 +12,7 @@ This perturbation would benefit all tasks which have a sentence/paragraph/docume
 text generation, etc. 
 
 ## What are the limitations of this transformation?
-The transformation's outputs may lose some information in case of longer sequences and may not be exact.
+The transformation's outputs may lose some information in case of longer sequences and may not change significantly.
+
+## Data and Code Provenance
+This transformation makes use of the code and model from: https://github.com/PrithivirajDamodaran/Styleformer
--- a/transformations/casual2formal/README.md
+++ b/transformations/casual2formal/README.md
@@ -1,0 +1,15 @@
+# Casual to Formal 🦎 → 🐍
+This transformation transfers text style from casual (informal) to formal.
+
+Author Name: Rishabh Gupta (rishabh19089@iiitd.ac.in)
+
+## What type of a transformation is this?
+This transformation acts like a paraphraser by performing style transfer. It changes the tone of the text from informal to formal.
+Example: What you upto? --> What are you currently doing? 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc. 
+
+## What are the limitations of this transformation?
+The transformation's outputs may lose some information in case of longer sequences and may not be exact.
--- a/transformations/casual2formal/transformation.py
+++ b/transformations/casual2formal/transformation.py
@@ -61,7 +61,6 @@ class Adequacy:
 class Casual2Formal(SentenceOperation):
     tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
     languages = ["en"]
-    heavy = True
 
     def __init__(self, num_beams=5, max_length=32, quality_filter=0.95, device=None):
         super().__init__()
--- a/transformations/casual2formal/transformation.py
+++ b/transformations/casual2formal/transformation.py
@@ -1,0 +1,115 @@
+from transformers import (
+    AutoTokenizer,
+    AutoModelForSeq2SeqLM,
+    AutoModelForSequenceClassification,
+)
+import torch
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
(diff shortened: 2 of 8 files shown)


rg089:
#75

transformations/casual2formal/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Author Name: Rishabh Gupta (rishabh19089@iiitd.ac.in)
+
+## What type of a transformation is this?
+This transformation acts like a paraphraser by performing style transfer. It changes the tone of the text from informal to formal.
+Example: What you upto? --> What are you currently doing? 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc.
```
kaustubhdhole, line 12: Please add what is the type of training data used here - I think the source of the model & code are taken from here: https://github.com/PrithivirajDamodaran/Styleformer
You should mention it here.

transformations/casual2formal/README.md
```diff
@@ -0,0 +1,15 @@
...
+## What type of a transformation is this?
+This transformation acts like a paraphraser by performing style transfer. It changes the tone of the text from informal to formal.
+Example: What you upto? --> What are you currently doing? 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc. 
+
+## What are the limitations of this transformation?
+The transformation's outputs may lose some information in case of longer sequences and may not be exact.
```
kaustubhdhole, line 15: Maybe you could add some of the bad generations too in the test file?

transformations/casual2formal/transformation.py
```diff
@@ -0,0 +1,115 @@
...
+    AutoModelForSeq2SeqLM,
+    AutoModelForSequenceClassification,
+)
+import torch
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+"""
+code and model from: https://github.com/PrithivirajDamodaran/Styleformer
+"""
```
kaustubhdhole, line 12: Yeah.. this line could in the README too. :)

transformations/casual2formal/transformation.py
```diff
@@ -0,0 +1,115 @@
...
+            adequacy_score = prob_label_is_true[0].item()
+            if adequacy_score >= adequacy_threshold:
+                adequacy_scores[para_phrase] = adequacy_score
+        return adequacy_scores
+
+
+class Casual2Formal(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+    heavy = True
```
kaustubhdhole, line 64: We've not made this mandatory now. You can remove this parameter to check whether the build passes. I will add it later.

transformations/formal2casual/README.md
```diff
@@ -0,0 +1,15 @@
...
+Author Name: Rishabh Gupta (rishabh19089@iiitd.ac.in)
+
+## What type of a transformation is this?
+This transformation acts like a paraphraser by performing style transfer. It changes the tone of the text from formal to informal.
+Example: This car looks fascinating --> This car is so cool
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc. 
+
```
kaustubhdhole, line 13: Please add a data and model section too like this: https://github.com/GEM-benchmark/NL-Augmenter/pull/116/files

transformations/casual2formal/README.md
rg089: Added it to README

transformations/casual2formal/README.md
rg089: Added negative sample

transformations/casual2formal/transformation.py
```diff
@@ -0,0 +1,114 @@
...
+        self.max_output = num_beams
+        self.num_beams = num_beams
+        self.max_length = max_length
+        self.quality_filter = quality_filter
+        self.device = device
+        if self.device is None:
+            self.device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    def generate(self, sentence: str):
+        ctf_prefix = "transfer Casual to Formal: "
```
kaustubhdhole, line 84: I think this all seems good now. I noticed that "Adequacy" is present in both the classes and hence would take up twice the memory. I would suggest one refactoring since most of the code in formal2casual and casual2formal is the same. You could make just one transformation folder - maybe called "formality_change", "style_change" or something better and write both the classes inside one single file _transformation.py_. And you could additionally include the active/passive conversion too if that works well.

kaustubhdhole: Unlike other transformations, it is okay in your case to load Adequacy (i.e. the model and the tokenizer) outside init (so that it is loaded only once for all 4 models).

rg089: Combined the two in #152.

kaustubhdhole:
Formality Change PR #152 is merged.

PR #83, opened 2021-06-28
Random deletion

EmileChapuis:


PierreColombo:
@EmileChapuis there is still an issue with the .json

Saad-Mahamood (approved):
Looks good to me. Approved.

Saad-Mahamood (approved):

tuetschek (commented):
Hi, looks good to me:
* The transformation is reasonable
* The code looks OK (minor comment: wouldn't it make sense to do all randomization in `numpy` so you don't need to seed twice?)
* The task types are appropriate
* There are examples

There are still things that might need solving:
* Any reason why the automated checks are failing? (Might not be related to the current PR?)
* Do you really need a specific version of numpy? Wouldn't just `numpy` without a version info be OK as a dependency?
* This approach simply tokenizes on whitespace. Other transformations (#100 I noticed) tokenize using spacy – don't we want that here too?

@kaustubhdhole I'm not sure about the latter two points – what do you think?

kaustubhdhole:
Yes, I agree with @tuetschek  - no need of updating the numpy version.
@EmileChapuis please pull master once and resolve conflicts - the checks might work after that.
I think whitespace tokenization should be fine too.

transformations/random_deletion/README.md
```diff
@@ -0,0 +1,33 @@
+# Butter Fingers Perturbation 🦎  + ⌨️ → 🐍
```
kaustubhdhole, line 1: Rename.

transformations/random_deletion/README.md
```diff
@@ -0,0 +1,33 @@
...
+  volume    = {abs/1901.11196},
+  year      = {2019},
+  url       = {http://arxiv.org/abs/1901.11196},
+  archivePrefix = {arXiv},
+  eprint    = {1901.11196},
+  timestamp = {Mon, 04 Feb 2019 08:11:03 +0100},
+  biburl    = {https://dblp.org/rec/journals/corr/abs-1901-11196.bib},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+```
```
kaustubhdhole, line 33: Please add limitations too - like deleting certain words can drastically change meaning - eg. negations.

transformations/random_deletion/test.json
```diff
@@ -0,0 +1 @@
...ence": "Andrew finally returned the French book to I last week"}]}, {"class": "RandomDeletion", "inputs": {"sentence": "Sentences with gapping, such as Paul likes coffee and Mary tea, lack an overt predicate to indicate the relation between two or more arguments."}, "outputs": [{"sentence": "with gapping, such Mary lack an overt to indicate the between two or more arguments."}]}, {"class": "RandomDeletion", "inputs": {"sentence": "Alice in Wonderland is a 2010 American live-action/animated dark fantasy adventure film"}, "outputs": [{"sentence": "Alice Wonderland is a 2010 American live-action/animated dark fantasy adventure film"}]}, {"class": "RandomDeletion", "inputs": {"sentence": "Ujjal Dev Dosanjh served as 33rd Premier of British Columbia from 2000 to 2001"}, "outputs": [{"sentence": "Ujjal Dev Dosanjh as 33rd Premier of British Columbia from 2000 to 2001"}]}, {"class": "RandomDeletion", "inputs": {"sentence": "Neuroplasticity is a continuous processing allowing short-term, medium-term, and long-term remodeling of the neuronosynaptic organization."}, "outputs": [{"sentence": "Neuroplasticity is a processing short-term, and remodeling of the neuronosynaptic organization."}]}]}
```
kaustubhdhole, line 1: JSON formatting.

transformations/random_deletion/README.md
```diff
@@ -0,0 +1,33 @@
...
+This transformation acts like a perturbation to test robustness. Each word of the given text are randomly deleted. 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc. 
+
+## Previous Work
+1) There has also been some recent work in the contrast sets of the GEM Benchmark (ACL 2021):
+```bibtex
+@article{DBLP:journals/corr/abs-1901-11196,
```
kaustubhdhole, line 18: Might want to cite the paper on word dropout too : https://arxiv.org/pdf/1511.01432.pdf

kaustubhdhole: And maybe this too: https://books.google.co.in/books?id=8wklDwAAQBAJ&pg=PT154&lpg=PT154&dq=%22word+dropout+may+also+be+beneficial+for+preventing+overfitting+and+improving+robustness+by+not+letting+the+model+rely+too+much+on+any+single+word+being+present&source=bl&ots=wFYULSyepA&sig=ACfU3U0YiewFhO-XZVDNt7caKclGUmdqSQ&hl=en&sa=X&ved=2ahUKEwij_aef5YPyAhVpIbcAHX1HAE4Q6AEwDnoECBcQAw#v=onepage&q=%22word%20dropout%20may%20also%20be%20beneficial%20for%20preventing%20overfitting%20and%20improving%20robustness%20by%20not%20letting%20the%20model%20rely%20too%20much%20on%20any%20single%20word%20being%20present&f=false

tuetschek (approved):
Looks good to me. Might be good to remove the test sample, but it's OK.

uyaseen:
@EmileChapuis Unfortunately, `RandomDeletion` tests failed in one of the [builds](https://github.com/GEM-benchmark/NL-Augmenter/pull/178/checks?check_run_id=3328912973). I think tests should always be deterministic and this can only be ensured if the value of probability is close to 1 or 0; for any other value, the tests may just fail any time. I would suggest adjusting the test cases by providing the extreme value of probability as an argument for every individual test case to make sure the test executes in a deterministic fashion.
What do you think @tuetschek @kaustubhdhole?

tuetschek:
Hi @uyaseen sorry for the delayed response. This is strange – with any probability values, the tests should be repeatable if the random seed is set properly. Maybe we need to look into random seeds before applying extreme values? Also, the link for the build you provide is for #178, not #83, so maybe this was supposed to go there?

uyaseen:
@tuetschek yes, you are right, it seems like the random seed is not set properly. Applying extreme values should be the last resort.

PR #84, opened 2021-06-28
Added gender culture diverse name transformation

XudongOliverShen:
This transformation changes a name with another, considering gender and cultural diversity.
Example: Rachel --> Leanne, Phoebe --> Muawiya, Joey --> Brunhilda, Chandler --> Rosalie, Monica --> Badiah

For a recognized name, this transformation randomly samples a (country, gender) pair, and samples a new name from that (country, gender).

Popular male and female names from 141 countries are collected and, on average, 778 names are included for each (country,gender) pair.

implemented both one way and two way.


samsontmr (changes requested):
This is great! Could you include in the README a list of the countries (names and country codes) for which you have data?

You also need to remove the files under `NL_Augmenter.egg-info` from the PR.

transformations/gender_culture_diverse_name/transformation.py
```diff
@@ -0,0 +1,147 @@
...
+            idxs = random.choices(range(len(ret)), k=max_output)
+            return [ret[idx] for idx in idxs], [ret_m[idx] for idx in idxs]
+        else:
+            return ret, ret_m
+                
+
+
+class gender_culture_diverse_name(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
```
samsontmr, line 112: One more comment: One more improvement here and in the other transformation would be to extend this to other languages, probably those that use Latin scripts so your data doesn't have to be modified.

XudongOliverShen: am thinking the same! 
one proposal: define an argument `in_native_language` with default `False`. If False, use an English alphabet dataset (eg, Chinese, Japanese, Korean will be converted to English alphabet using standard translation). If True, non-English names will be in their native alphabets.

Any suggestion?

transformations/gender_culture_diverse_name/transformation.py
```diff
@@ -0,0 +1,147 @@
...
+                        self.name2country[name].add(country)
+
+    def apply(self, doc, n=10, max_output=10, seed=None):
+        """Replace names with another name, considering gender and cultural diversity
+
+        Parameters
+        ----------
+        doc : spacy.token.Doc
+            input
+        meta : bool
```
Nickeilf, line 56: This `meta` does not exist in the function parameters.

transformations/gender_culture_diverse_name/transformation.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+test = gender_culture_diverse_name()
+# sentence = 'Rachel Green, a sheltered but friendly woman, flees her wedding day and wealthy yet unfulfilling life.'
+# sentence = 'Phoebe Buffay is an eccentric masseuse and musician.'
+# sentence = 'Joey has many short-term girlfriends.'
+# sentence = 'Chandler Bing is a sarcastic and self-deprecating IT manager.'
+sentence = 'Monica was overweight as a child.'
+p = test.generate(sentence)
+print(p[0])
```
Nickeilf, line 147: Don't forget to remove these lines or put them in the ifmain, otherwise when importing the transformation they will be executed.

Nickeilf (commented):
Hi @XudongOliverShen, thanks for contributing this interesting transformation!

I've left some comments in the code. Here's my general review:
**Correctness**: All checks have passed - please pull "main" into your branch.
**Specificity**: The change is specific to text with names.
**Novelty**: This transformation is novel and not implemented in NL-Augmenter.
**Test Cases**: 2 test cases have been added.

transformations/gender_culture_diverse_name/transformation.py
Nickeilf: Same issue for the two-way version

transformations/gender_culture_diverse_name/transformation.py
Nickeilf: Also applies to the two-way version

XudongOliverShen: Hi, thanks for the comment! fixed now.

Nickeilf (approved):
Thanks for the changes! I have passed my review.

transformations/gender_culture_diverse_name/transformation.py
samsontmr: hm if you have a reliable way for doing the transliteration that would be good. I would use `latin_script=True` instead of `in_native_language=False` since that could be potentially ambiguous.

samsontmr: To clarify, the change I was referring to in the initial comment is to just extend this list to include the languages/regions your existing name list covers.

XudongOliverShen: Ohh I see. I think it is a miscommunication. The current version has translated all names into English, thus actually didn't cover any other language. So I didn't extend the list.

If the feature that I referred to is implemented in the future, I will update the list accordingly.

samsontmr: Wouldn't the names from languages using latin scripts be largely preserved in English? In that case they would be valid for their own languages, and the names from all latin script languages should be valid in other latin script languages too (this is my hypothesis which needs some verification :P)

XudongOliverShen: updated the list to include all languages covered in this dataset : )

XudongOliverShen:
I noticed the failure of the test is due to several errors in the already accepted PR. I've actually self-tested this transformation and it (should) passes the test.

samsontmr (approved):

kaustubhdhole:
Hi @XudongOliverShen these are wonderful, important and well-written changes! Your README is precisely written and highlights the poor coverage of previously used name substitutions! I am merging your PR now!

transformations/gender_culture_diverse_name/transformation.py
```diff
@@ -0,0 +1,167 @@
...
+        
+        if len(ret) > max_output:
+            idxs = random.choices(range(len(ret)), k=max_output)
+            return [ret[idx] for idx in idxs], [ret_m[idx] for idx in idxs]
+        else:
+            return ret, ret_m
+                
+
+
+class gender_culture_diverse_name(SentenceOperation):
```
kaustubhdhole, line 112: A camel casing name would be better for consistency with other classes.

transformations/gender_culture_diverse_name/transformation.py
```diff
@@ -0,0 +1,167 @@
...
+        'tr', 'si', 'pa', 'sw', 'sl', 
+        'bg', 'lv', 'hr', 'uk', 'mr', 
+        'sk', 'ps', 'bs', 'se', 'he', 
+        'tn', 'it', 'hu', 'cs', 'nl', 
+        'ru', 'et', 'uz', 'gl', 'sq', 
+        'ee', 'fi', 'cy'
+        ]
+    # language code following ISO 639-1 standard
+
+    def __init__(self, n=1, seed=0, max_output=1, retain_gender=False, retain_culture=False, data_path=None):
```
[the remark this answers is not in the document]
XudongOliverShen: You mean to showcase how outputs differ with different values of `retain_gender` and `retain_culture`, at the end of the file `transformation.py`? Or the `test.json` file?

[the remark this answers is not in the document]
kaustubhdhole: In the test.json..

kaustubhdhole:

PR #86, opened 2021-06-29
Added filter selecting strings containing characters outside a given encoding

janpf:


filters/contains-non-ascii/filter.py
```diff
@@ -0,0 +1,15 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class TextContainsNonAsciiFilter(SentenceOperation):
+    tasks = [e for e in TaskType]
+    languages = ["en", "de", "all"]
```
sebastianGehrmann, line 7: @kaustubhdhole correct me if I am wrong, but if `all` is set, the other two do not have to, right?

filters/contains-non-ascii/README.md
```diff
@@ -0,0 +1,12 @@
+## unicode filter
+
+## What type of a filter is this?
+
+This filter filters examples which contain non-ascii characters.
+Author: Jan Pfister
+
+## Why is measuring performance on this split important?
+This filter can be used to find datapoints containing non-ascii unicode characters. Filtering out and testing on examples containing these characters can provide feedback for improving models accordingly as most models are trained on plain english text mostly containing ascii characters. Sometimes non-ascii character are even explicitly stripped away.
```
sebastianGehrmann, line 9: nit: English, not english

Also, since this works for all languages, maybe change this paragraph to address what it would do for other languages?

filters/contains-non-ascii/filter.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    def filter(self, sentence: str = None) -> bool:
+        print(sentence)
```
sebastianGehrmann, line 13: nit: remove.

filters/contains-non-ascii/filter.py
```diff
@@ -0,0 +1,15 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class TextContainsNonAsciiFilter(SentenceOperation):
```
sebastianGehrmann, line 5: Scope: I think this filter style of identifying types of characters could be extended to beyond only ASCII. For example, `containsEmoji`, `containsNonUtf8`, `containsCyrillicCharacters`, etc.

sebastianGehrmann (changes requested):

filters/contains-non-ascii/filter.py
janpf: Thanks for the feedback! Great idea!

filters/encoding/filter.py
```diff
@@ -0,0 +1,15 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class TextEncodingFilter(SentenceOperation):
+    tasks = [e for e in TaskType]
+    languages = ["en", "de"]
```
sebastianGehrmann, line 7: Wouldn't this work for "all" in addition to en and de?

sebastianGehrmann (approved):

filters/encoding/filter.py
janpf: Sure! Didn't know, whether "all" is a valid setting.

filters/contains-non-ascii/filter.py
kaustubhdhole: Yes, "all" represents all languages and the others need not be added.

kaustubhdhole:
@janpf these are great changes: Would you also like to evaluate it using the [prebuilt evalutation engine](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation):

`python evaluate.py -f TextEncodingFilter`

janpf:
> @janpf these are great changes: Would you also like to evaluate it using the [prebuilt evalutation engine](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation):
> 
> `python evaluate.py -f TextEncodingFilter`

I added the results to the Readme

kaustubhdhole:
This looks fine to me overall.

kaustubhdhole (approved):

kaustubhdhole:
Thank you very much for your changes. Your changes are merged. As the field of NLP has largely focussed on high-resource languages and domains, if possible, we also encourage you to add a transformation in your mother tongue and/or a low-resource language which might need more attention from the NLP community. This is not mandatory.

PR #88, opened 2021-06-29
Add "Greetings and Farewells" transformation

paulozip:
Hello everyone. Congratulations on this awesome project. I think it will be a big contribution to the community.

I created this transformation that **basically creates some text variants and perturbations on words and expressions related to greetings and farewells**. For instance, sentences containing words like "Hi!" or "What's up?" will have these words replaced by similar ones, like "Howdy" or "Good morning". The same behavior will be done for farewells and special cases (e.g. "It's nice to meet you" or "How's it going?").

**The main goal of this transformation is to tackle the robustness of text classification/generation models when faced with perturbations**.

Let me know if you need more details about the implementation.

### DIFF
--- a/transformations/greetings_and_farewells/transformation.py
+++ b/transformations/greetings_and_farewells/transformation.py
@@ -1,0 +1,116 @@
+import itertools
+import random
+import re
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+# Setting up a tuple of possible replaceable greetings and farewells
+GREETINGS = (
+    "Hi",
+    "Hello",
+    "Hey",
+    "Howdy",
+    "Greetings",
+    "Good morning",
+    "Good afternoon",
+    "Good evening",
+    "What's up",
+    "Sup",
+)
+
+# Creating a tuple of sentences that can't be simply replaced by a short greeting
+SPECIAL_GREETINGS = (
+    "It's nice to meet you.",
+    "It's great to meet you.",
+    "Pleased to meet you.",
+    "How are you?",
+    "How are you doing?",
+    "How is it going on?",
+    "How is it going?",
+)
+
+FAREWELLS = (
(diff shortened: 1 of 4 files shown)


transformations/greetings_and_farewells/transformation.py
```diff
@@ -0,0 +1,116 @@
...
+    re.compile("have a (great|nice|good) (day|night|week|weekend)", re.IGNORECASE),
+    re.compile("best regards", re.IGNORECASE),
+    re.compile(
+        r"\b(by+e+)+\b", re.IGNORECASE
+    ),  # it matches 'bye', 'bye bye', 'byyyyyyye', 'byeeeee', etc.
+)
+
+
+def greetings_and_farewells(text, seed=0, max_output=1):
+    random.seed(seed)
```
samsontmr, line 77: I believe `random.seed()` needs to be called each time before `random.choice()`. I would recommend generating a random integer as a seed for each `random.choice()` call using this seed to increase diversity in the test examples while remaining reproducible.

samsontmr (changes requested):
Thanks for the contribution! This generally looks good. I noticed that you're located in Brazil; it would be great if you could implement this transformation for any other languages you speak!

kaustubhdhole:
Hi @paulozip, thank you for you changes! Here is my overall review:

Build: passes
Interface: Only tagging won't be applicable since the number of tokens are changing in some cases.
Specificity: This change is specific.
Paraphrasing: I think the coverage is very low.
Description: README describes properly.
Evaluating Robustness: No evaluation has been conducted. 
Test cases: 5 have been added.

I find it fine overall, please remove TEXT_TAGGING and if possible, add more variations to increase coverage.

Thanks for your effort!

samsontmr (approved):
Approving this, the extension to other languages can be done in a separate PR.

kaustubhdhole (approved):

PR #93, opened 2021-06-29
Transformer Contextual Word Replacement

gautierdag:
This PR proposes `TransformerFill`, a transformation that replaces a selection of words with suggestions from a Huggingface transformer model. Selection is done through POS tags and Spacy. It is also possible to pass in a string to prime the model towards certain predictions.


transformations/transformer_replace/test.json
```diff
@@ -0,0 +1,46 @@
+{
+  "type": "transformer_replace",
+  "test_cases": [{
+      "class": "TransformerReplace",
+      "inputs": {
+        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
+      },
+      "outputs": [{
+        "sentence": "I finally returned the French book to Chris Evans I wrote last week"
```
tongshuangwu: Looks like the only sentence changed is this one? All the other outputs are just the same as the inputs? I would suggest at least filtering out un-perturbed sentences?

transformations/transformer_replace/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+The accuracy of a RoBERTa model (fine-tuned on IMDB) (model: "aychang/roberta-base-imdb")
+on a subset of IMDB sentiment dataset = 95.74
+The accuracy of the same model on the perturbed set = 88.26
+
+The average bleu score of a distillbert model (fine-tuned on xsum) (model: "sshleifer/distilbart-xsum-12-6")
+on a subset (10%) of xsum test dataset = 14.9104
+The average bleu score of same model on the pertubed set = 11.9221
+
+## Previous Work
```
tongshuangwu: Some work that I know use LMs to do contextual word replacement -- I'd suggest consider adding refs to them!

[Beyond Accuracy: Behavioral Testing of NLP Models with CheckList](https://homes.cs.washington.edu/~marcotcr/acl20_checklist.pdf)

[Conditional BERT Contextual Augmentation](https://arxiv.org/abs/1812.06705)

[Contextualized Perturbation for Textual Adversarial Attack](https://arxiv.org/pdf/2009.07502.pdf)


I'm also interested in what the author see as the difference between this pull request and https://github.com/GEM-benchmark/NL-Augmenter/pull/81 ? Seems the two share the same spirit as far as I would concern?

transformations/transformer_replace/transformation.py
```diff
@@ -0,0 +1,161 @@
...
+        doc = self.nlp(sentence)
+        masked_texts, original_words = self.get_masked_sentences_from_sentence(doc)
+        if len(masked_texts) > 0:
+            predictions = self.fill_pipeline(masked_texts)
+            new_sentence = self.generate_from_predictions(
+                doc, predictions, original_words
+            )
+            return [new_sentence]
+        else:
+            return [sentence]
```
tongshuangwu: Might be more reasonable if it's just `[]` here so we can filter out unchanged examples.

transformations/transformer_replace/README.md
gautierdag: Hey @tongshuangwu, thank you for the review 🙌 ! 
I can certainly cite these papers - I did cite a few but it seems that you reviewed the first commit on my branch rather than the last? I already cite `Conditional BERT Contextual Augmentation`, but will add the rest.

Definitely a good point as well, when I drafted this PR originally I distinctly remember that #81 was doing things differently, but it seems now that there is indeed a lot overlap with it if you consider the base defaults.

However this approach does **two** main things differently:
1. It offers a way to select the words to replace - through the use of a spacy model and based on POS tokens. Because it uses Spacy it is probably also suitable to use for other languages as long as the spacy model and the huggingface are targeting the same underlying language.

2. It offers a way to _prime_ the model and context. A big problem in these types of augmentation is that meaning can be changed. Giving a way to do so by prepending some context to the sentence to augment can be a way to limit the meaning to drift. 
For instance if I was doing sentiment analysis and have the sentence "We had some delicious food at this restaurant", and that I am replacing every adjective with a suggestion from the model, then I would mask delicious to have: "We had some MASK food at this restaurant". A model could then potentially complete the MASK with "terrible" or "aweful" and completely change the meaning of the sentence. 
However if I prime the context with for instance "POSITIVE : " so that my masked sentence becomes "POSITIVE : We had some MASK food at this restaurant", the model will be less likely to change the meaning of the sentence and thus give an augmentation that might correspond better to the target.

Both of these are in fact points that @kaustubhdhole makes on #81 (https://github.com/GEM-benchmark/NL-Augmenter/pull/81#issuecomment-880671719
) 😄 

This all means that the randomness of the augmentation can be somewhat better controlled through the word selection mechanism and through additional context.

transformations/transformer_fill/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## What are the limitations of this transformation?
+
+There are many limitations to such approach and any use of this augmentation will need to be carefully monitored and guided for the task at hand.
+
+1. It is possible that the new augmented sentence has a different meaning from the original, since it is highly dependent on what words the model is asked to replace.
+
+   For instance the sentence `"This is pretty cool"`. If you want choose to replace adjectives, you might end up asking the model to fill in the blank for the sentence: `"This is pretty <mask>"`. In which case it might decide that the most likely word would make: `"This is pretty bad"`.
+
+   A way to alleviate this is to condition the augmentation or prime the model towards certain words. For instance, by prepending `"Positive review:"` to the aformentioned sentence, we are more likely to obtain predictions that will also align with the prompt or primer.
```
kaustubhdhole, line 62: Hi @gautierdag, thanks for you changes. Have you been able to test this?

tongshuangwu: Similar question as @kaustubhdhole --- I'm not that familiar with distillbert to know if this kind of prefix effect require specific finetuning, but if it's a quantifiable effect (or if there's some paper doing it), then it'd be really cool! other than this, I'm happy to approve the PR.

gautierdag: Thank you both for the feedback 😄! I'm currently travelling but will come back to it and add additional tests in the next two weeks.

kaustubhdhole:
@gautierdag reminder

transformations/transformer_fill/README.md
gautierdag: Hey guys, sorry it took me some time to come back to this. 

I've just added a few extra test cases that show priming the model does impact the replaced words.

In addition, I ran the evaluation code provided to measure the impact of the accuracy drop on the IMDB text classification model (`aychang/roberta-base-imdb`).

I use two `TransformerFill` objects to transform the positive and negative examples differently (with different contexts).

```python
from transformations.transformer_fill import TransformerFill

positive_tf = TransformerFill(context_text="POSITIVE REVIEW:", pos_tokens=["ADJ"])
negative_tf = TransformerFill(context_text="NEGATIVE REVIEW:", pos_tokens=["ADJ"])
```

And obtained the following:

```
Non-transformed dataset:
The accuracy on this subset which has 5000 examples= 95.74
Transformed set accuracy:
The accuracy on this subset which has 4999 examples= 95.79915983196639
```

The accuracy slightly increases on the augmented set because adjectives are being switched out for positive/negative counterparts which might align with the target label. 

In doing so I also noticed that long sentences (outside the model max length) would need to be specially truncated due to the variability of the context and the replacement of words -> `<mask>`. So that is now fixed, and it now works out of the box for long texts as well.

@tongshuangwu As to your question about whether the prefix requires fine tuning: In the Wu et. al. 2019 paper on Conditional Bert Contextual Augmentation, they test with and without fine tuning and while fine tuning a LM for the augmentation was best, the naive BERT augmentation still outperformed more traditional augmentation methods such as synonyms.

Let me know if this is enough for you both, I hope to find the time to run additional experiments using the augmented set to train models in low data settings - because this is where this technique should really shine.

kaustubhdhole:
@tongshuangwu reminder

transformations/transformer_fill/README.md
tongshuangwu: Thanks for the clarification! Happy to approve on my end.

tongshuangwu (approved):

transformations/transformer_fill/test.json
```diff
@@ -0,0 +1,96 @@
...
+      },
+      "outputs": [{
+        "sentence": "This movie was terrible, I'll never watch it again."
+      }]
+    },
+    {
+      "class": "TransformerFill",
+      "args": {
+        "context_text": "POSITIVE REVIEW: ",
+        "pos_tokens": ["ADJ"]
```
kaustubhdhole, line 73: I am approving this PR. The README is nicely written. Besides, I think this is really interesting and deserves more experimentation. I'm not sure of studies like this but I think at a generic level, you can probably make a check if appending a context like this with class descriptions can force the model to pick up relevant words belonging to that class. A set of keywords (in this case: a set of positive and negative) can make it easy to make an empirical assessment.

kaustubhdhole (approved):

PierreColombo:
Hi since you merged our PR can you add this relevant citation ? 
@inproceedings{DBLP:conf/nips/JalalzaiCCGVVS20,
author = {Hamid Jalalzai and
Pierre Colombo and
Chlo{'{e}} Clavel and
{'{E}}ric Gaussier and
Giovanna Varni and
Emmanuel Vignon and
Anne Sabourin},
editor = {Hugo Larochelle and
Marc'Aurelio Ranzato and
Raia Hadsell and
Maria{-}Florina Balcan and
Hsuan{-}Tien Lin},
title = {Heavy-tailed Representations, Text Polarity Classification {&} Data
Augmentation},
booktitle = {Advances in Neural Information Processing Systems 33: Annual Conference
on Neural Information Processing Systems 2020, NeurIPS 2020, December
6-12, 2020, virtual},
year = {2020},
url = {https://proceedings.neurips.cc/paper/2020/hash/2cfa3753d6a524711acb5fce38eeca1a-Abstract.html},
timestamp = {Tue, 19 Jan 2021 00:00:00 +0100},
biburl = {https://dblp.org/rec/conf/nips/JalalzaiCCGVVS20.bib},
bibsource = {dblp computer science bibliography, https://dblp.org}
}

PR #94, opened 2021-06-30
Add Diverse Paraphrase Generation.

ashutoshml:
With support for 4 candidate selection mechanism: DiPS, Diverse Beam, Beam Search, Random Selection

### DIFF
--- a/transformations/diverse_paraphrase/README.md
+++ b/transformations/diverse_paraphrase/README.md
@@ -1,0 +1,59 @@
+# Diverse Paraphrase Generation🦎  + ⌨️ → 🐍
+This transformation produces multiple diverse paraphrases for a given sentence in English.
+
+## What type of a transformation is this?
+This transformation is a multi-output sentence level paraphrase generation model, specifically catered towards generating `num_outputs` (user specified) diverse outputs. 
+
+It has support for 4 candidate selection methods:
+
+a) dips: Based on Kumar et. al. 2019. (See Below)
+b) diverse_beam: Based on Vijaykumar et. al 2018. (See Below)
+c) beam: Selects top `num_outputs` candidates in the beam search.
+d) random: Randomly selects `num_outputs` candidates.
+
+Eg:
+```python
+>>> t = DiverseParaphrase(augmenter='dips', num_outputs=4)
+>>> t.generate('Joe Biden is the President of USA.')
+```
+
+Replace augmenter with any of the above mentioned options. 
+Default: augmenter='dips', num_outputs=4. 
+In most cases, dips should be the preferred choice.
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which need diverse paraphrase candidates for augmentation in tasks like text classification, text generation, etc. 
+
+## Previous Work
+1. DiPS: Original Implementation [here](https://github.com/malllabiisc/DiPS)
+```bibtex
+@inproceedings{dips2019,
+    title = "Submodular Optimization-based Diverse Paraphrasing and its Effectiveness in Data Augmentation",
+    author = "Kumar, Ashutosh  and
+      Bhattamishra, Satwik  and
+      Bhandari, Manik  and
+      Talukdar, Partha",
+    booktitle = "Proceedings of the 2019 Conference of the North {A}merican Chapter of the Association for Computational Linguistics: Human Language Technologies, Volume 1 (Long and Short Papers)",
+    month = jun,
+    year = "2019",
+    address = "Minneapolis, Minnesota",
+    publisher = "Association for Computational Linguistics",
+    url = "https://www.aclweb.org/anthology/N19-1363",
+    pages = "3609--3619"
+}
+```
+
+2. Diverse Beam
+```bibtex
+@paper{AAAI1817329,
+	author = {Ashwin Vijayakumar and Michael Cogswell and Ramprasaath Selvaraju and Qing Sun and Stefan Lee and David Crandall and Dhruv Batra},
+	title = {Diverse Beam Search for Improved Description of Complex Scenes},
+	conference = {AAAI Conference on Artificial Intelligence},
+	year = {2018},
+	keywords = {Recurrent Neural Networks, Beam Search, Diversity},
+	url = {https://www.aaai.org/ocs/index.php/AAAI/AAAI18/paper/view/17329}
+}
+```
+
+## What are the limitations of this transformation?
+The base paraphrasing model used by this transformation is backtranslation (En -> De -> En). If the base model is poor, the candidate outputs will be of low-quality.
--- a/transformations/diverse_paraphrase/submod/submodopt.py
+++ b/transformations/diverse_paraphrase/submod/submodopt.py
@@ -1,0 +1,81 @@
+import numpy as np
+import scipy.linalg as la
+import pdb
+
+from transformations.diverse_paraphrase.submod.submodular_funcs import (
+    distinct_ngrams,
+    ngram_overlap,
+    similarity_func,
+    seq_func,
+    ngram_overlap_unit,
+    similarity_gain,
+    seq_gain,
+)
+
+
+class SubmodularOpt:
+    def __init__(self, V=None, v=None, **kwargs):
+        self.v = v
+        self.V = V
+
+    def initialize_function(self, lam, a1=1.0, a2=1.0, b1=1.0, b2=1.0):
+        self.a1 = a1
+        self.a2 = a2
+        self.b1 = b1
+        self.b2 = b2
+
+        self.noverlap_norm = ngram_overlap(self.v, self.V)
+        self.ndistinct_norm = distinct_ngrams(self.V)
+        self.sim_norm = similarity_func(self.v, self.V)
+        self.edit_norm = np.sqrt(len(self.V))
+        self.lam = lam
+
+    def final_func(self, pos_sets, rem_list, selec_set):
+        distinct_score = (
+            np.array(list(map(distinct_ngrams, pos_sets))) / self.ndistinct_norm
+        )
+
+        base_noverlap_score = ngram_overlap(self.v, selec_set)
+        base_sim_score = similarity_func(self.v, selec_set)
+        base_edit_score = seq_func(self.V, selec_set)
+
+        noverlap_score = []
+        for sent in rem_list:
+            noverlap_score.append(ngram_overlap_unit(self.v, sent, base_noverlap_score))
+        noverlap_score = np.array(noverlap_score) / self.noverlap_norm
+
+        sim_score = []
+        for sent in rem_list:
+            sim_score.append(similarity_gain(self.v, sent, base_sim_score))
+        sim_score = np.array(sim_score) / self.sim_norm
+
+        edit_score = []
+        for sent in rem_list:
+            edit_score.append(seq_gain(self.v, sent, base_edit_score))
+        edit_score = np.array(edit_score) / self.edit_norm
+
+        quality_score = self.a1 * sim_score + self.a2 * noverlap_score
+        diversity_score = self.b1 * distinct_score + self.b2 * edit_score
+
+        final_score = self.lam * quality_score + (1 - self.lam) * diversity_score
+
+        return final_score
+
+    def maximize_func(self, k=5):
+        selec_sents = set()
+        ground_set = set(self.V)
+        selec_set = set(selec_sents)
+        rem_set = ground_set.difference(selec_set)
+        while len(selec_sents) < k:
+
+            rem_list = list(rem_set)
+            pos_sets = [list(selec_set.union({x})) for x in rem_list]
+
+            score_map = self.final_func(pos_sets, rem_list, selec_set)
+            max_idx = np.argmax(score_map)
+
+            selec_sents = pos_sets[max_idx]
+            selec_set = set(selec_sents)
+            rem_set = ground_set.difference(selec_set)
+
+        return selec_sents
--- a/transformations/diverse_paraphrase/submod/submodopt_intent.py
+++ b/transformations/diverse_paraphrase/submod/submodopt_intent.py
@@ -1,0 +1,51 @@
+import numpy as np
+import scipy.linalg as la
+import pdb
+from collections import Counter
+import string
+
+from transformations.diverse_paraphrase.submod.submodular_funcs import distinct_ngrams
+
+
+class SubmodularOpt:
+    def __init__(self, V=None, A=None, A_=None, **kwargs):
+        self.A = A
+        self.A_ = A_
+        self.V = V
+
+    def initialize_function(self, lam=0.5, a1=1.0, a2=1.0, b1=1.0, b2=1.0):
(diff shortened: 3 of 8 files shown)


kaustubhdhole:
I think these changes seem like a wondeful addition to NL-Augmenter - I haven't gone through the entire code yet. But you'll need to shift all the dependencies into a separate requirements.txt in the folder "diverse_paraphrase" to avoid conflicts. 

Also, unfortunately gensim cannot be added here due to a non-permissive license. Can you use any other substitute?

ashutoshml:
Made the necessary changes. 

1. Added requirements.txt in the `diverse_paraphrase` directory
2. Replaced `gensim` word2vec with `torchtext `GloVe

transformations/diverse_paraphrase/README.md
```diff
@@ -0,0 +1,60 @@
...
+	title = {Diverse Beam Search for Improved Description of Complex Scenes},
+	conference = {AAAI Conference on Artificial Intelligence},
+	year = {2018},
+	keywords = {Recurrent Neural Networks, Beam Search, Diversity},
+	url = {https://www.aaai.org/ocs/index.php/AAAI/AAAI18/paper/view/17329}
+}
+```
+
+## What are the limitations of this transformation?
+The base paraphrasing model used by this transformation is backtranslation (En -> De -> En). If the base model is poor, the candidate outputs will be of low-quality.
```
Nickeilf, line 60: The base version seems to be the same as the [back-translation transformation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/back_translation). Therefore, this transformation can be seem as a superset of the existing one. @kaustubhdhole Shall we include both?

transformations/diverse_paraphrase/submod/submodopt_intent.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+import scipy.linalg as la
+import pdb
+from collections import Counter
+import string
+
+from transformations.diverse_paraphrase.submod.submodular_funcs import distinct_ngrams
+
+
+class SubmodularOpt:
```
Nickeilf, line 10: This file seems to be not used. Correct me if I'm wrong

transformations/diverse_paraphrase/submod/submodopt.py
```diff
@@ -0,0 +1,81 @@
...
+    seq_gain,
+)
+
+
+class SubmodularOpt:
+    def __init__(self, V=None, v=None, **kwargs):
+        self.v = v
+        self.V = V
+
+    def initialize_function(self, lam, a1=1.0, a2=1.0, b1=1.0, b2=1.0):
```
Nickeilf, line 21: It would be good to add some comments about what these arguments are.

transformations/diverse_paraphrase/submod/submodopt.py
```diff
@@ -0,0 +1,81 @@
...
+    similarity_func,
+    seq_func,
+    ngram_overlap_unit,
+    similarity_gain,
+    seq_gain,
+)
+
+
+class SubmodularOpt:
+    def __init__(self, V=None, v=None, **kwargs):
```
Nickeilf, line 17: Comments here might be needed. I guess the V and v correspond to the source language and the target language?

transformations/diverse_paraphrase/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# for diverse paraphrase dips
+nltk==3.6.2
+torchtext
```
Nickeilf, line 3: Is there any particular reason for not adding version for torchtext?

Nickeilf (commented):
Hi @ashutoshml, thanks for contributing this interesting transformation!

Here's my general review:
**Correctness**: Everything seems correct and good to me.
**Specificity**: The transformation is not specified to certain types of texts.
**Novelty**: This transformation is novel and has not been implemented in NL-Augmenter. It generates multiple semantically equal paraphrases via round-trip-translation.
**Adding New Libraries**: New libraries added in the folder requirements.txt.

The only suggestion I'd make is to add some more comments in the code, as this will improve the readability of the transformation and be more friendly to end users.

transformations/diverse_paraphrase/submod/submodopt.py
ashutoshml: V is the ground set from which final candidates need to be selected.
v is the original sentence being used for selecting meaning-preserving candidates. 

As suggested, I'll add the comments.

transformations/diverse_paraphrase/requirements.txt
ashutoshml: The main `requirements.txt` file does not have a specific version for `torch`.
`torchtext` depends on the original torch version. 

If the version for the original torch is added, this can be fixed too.

transformations/diverse_paraphrase/submod/submodopt.py
ashutoshml: I'll add the comments

transformations/diverse_paraphrase/submod/submodopt_intent.py
ashutoshml: You are correct, this was being used for some other results in the original paper. 
I'll remove this file.

ashutoshml:
Hi @Nickeilf 

Thanks for the review.
Based on the comments provided, I have updated the code.

transformations/diverse_paraphrase/requirements.txt
Nickeilf: @kaustubhdhole What's your opinion? Shall we specify a torch version in the main requirement file?

transformations/diverse_paraphrase/README.md
kaustubhdhole: Yes for the purpose of this PR, we can keep both.

transformations/diverse_paraphrase/requirements.txt
kaustubhdhole: I avoided specifying the torch version to make it quicker for colab users - Colab has a prebuilt torch installed. (Torch takes a lot of time to download) Its fine to keep torchtext that way. A good idea would be to mention in the README which versions of torchtext are preferred (if there's any such preference).

ashutoshml: There is no preference. 

I just tested the code with torch==1.9.0 and torchtext==0.10.0. Although the useful functions in torchtext are present in all the earlier available versions too, and is unlikely to be removed in the future.

transformations/diverse_paraphrase/transformation.py
```diff
@@ -0,0 +1,151 @@
...
+from transformations.diverse_paraphrase.submod.submodopt import SubmodularOpt
+from transformations.diverse_paraphrase.submod.submodular_funcs import trigger_dips
+
+
+class DiverseParaphrase(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+    heavy = True
+
+    def __init__(self, augmenter="dips", num_outputs=4):
```
kaustubhdhole, line 19: You should pass seed here too.

transformations/diverse_paraphrase/transformation.py
```diff
@@ -0,0 +1,151 @@
...
+
+        seed = 42
+        random.seed(seed)
+        np.random.seed(seed)
+        torch.manual_seed(seed)
+        torch.use_deterministic_algorithms(True)
+        if torch.cuda.is_available():
+            torch.cuda.manual_seed_all(seed)
+
+        assert augmenter in ["dips", "random", "diverse_beam", "beam"]
```
kaustubhdhole, line 30: I think it can be useful for people to see the difference in the outputs of these: You can show 1-2 sentences with different "augmenters" by passing it in args in test.json
Here is an example: https://github.com/GEM-benchmark/NL-Augmenter/blob/main/filters/keywords/test.json

transformations/diverse_paraphrase/transformation.py
ashutoshml: Should I move the seeds from __init__ to class variables?

ashutoshml: As an argument to `__init__`?

kaustubhdhole: Yes!

transformations/diverse_paraphrase/transformation.py
ashutoshml: Sure. The build seems to be failing because of test accuracy evaluation (test.json).
Any suggestions?

ashutoshml: Reminder. Any suggestions on this?

I had earlier set the variable `heavy=True` as in back_translation code.

kaustubhdhole: @aadesh11 @AbinayaM02 any idea of this?

AbinayaM02: @kaustubhdhole : I checked the build failure log. The pytest crashed after it downloaded the necessary libraries (all heavy ones). This looks like an OOM issue on the Github runner, that's my best guess.

The following is the specification of the Github runner,
Hardware specification for Windows and Linux virtual machines:
- 2-core CPU
- 7 GB of RAM memory
- 14 GB of SSD disk space

kaustubhdhole: @ashutoshml can you please try the following in your branch (on your machine):
`pytest -s --t=diverse_paraphrase
`
and share the output.

ashutoshml:
```shell
test/test_main.py:105:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test/test_main.py:83: in execute_test_case_for_transformation
    execute_sentence_operation_test_case(transformation, test)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

transformation = <transformations.diverse_paraphrase.transformation.DiverseParaphrase object at 0x7fdf0499c110>
test = {'args': {'augmenter': 'dips'}, 'class': 'DiverseParaphrase', 'inputs': {'sentence': 'Alice in Wonderland is a 2010 Am...0'}, {'sentence': 'Alice in Wonderland is an American live action / animated dark fantasy adventure movie from 2010'}]}

    def execute_sentence_operation_test_case(transformation, test):
        filter_args = test["inputs"]
        outputs = test["outputs"]
        perturbs = transformation.generate(**filter_args)
        for pred_output, output in zip(perturbs, outputs):
>           assert pred_output == output["sentence"], get_assert_message(
                transformation, output["sentence"], pred_output
            )
E           AssertionError: Mis-match in expected and predicted output for DiverseParaphrase transformation:
E              Expected Output: Alice in Wonderland is an American live-action / animated dark fantasy adventure film from the year 2010
E              Predicted Output: Alice in Wonderland is an American live action / animated dark fantasy adventure movie from 2010
E           assert 'Alice in Won...vie from 2010' == 'Alice in Won...the year 2010'
E             - Alice in Wonderland is an American live-action / animated dark fantasy adventure film from the year 2010
E             ?                                        ^                                         ^ ^^      ---------
E             + Alice in Wonderland is an American live action / animated dark fantasy adventure movie from 2010
E             ?                                        ^                                         ^^^ ^

test/test_main.py:26: AssertionError
=============================================================================================================================== warnings summary ===============================================================================================================================
test/test_main.py::test_operation[diverse_paraphrase-length]
  /scratche/prod/ashutosh/miniconda3/envs/nlgenv/lib/python3.7/site-packages/torch/_tensor.py:575: UserWarning: floor_divide is deprecated, and will be removed in a future version of pytorch. It currently rounds toward 0 (like the 'trunc' function NOT 'floor'). This results in incorrect rounding for negative values.
  To keep the current behavior, use torch.div(a, b, rounding_mode='trunc'), or for actual floor division, use torch.div(a, b, rounding_mode='floor'). (Triggered internally at  /pytorch/aten/src/ATen/native/BinaryOps.cpp:467.)
    return torch.floor_divide(self, other)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================================================================================================================== short test summary info ============================================================================================================================
FAILED test/test_main.py::test_operation[diverse_paraphrase-length] - AssertionError: Mis-match in expected and predicted output for DiverseParaphrase transformation:
=================================================================================================================== 1 failed, 1 warning in 288.53s (0:04:48) ===================================================================================================================
```

kaustubhdhole: Okay, which part is the probabilistic part?

ashutoshml: There are two ways in which probabilistic outputs may happen:

1. When the candidates are generated using beam search, the argmax operation is resolved non-deterministically in the case of two candidates having the same max-scores. 

2. For `dips`, there is a functional maximization operation that occurs resulting in similar non-determinism. The first selected candidate governs how the rest of the candidates are selected. Since the objective is to maximize the lexical-syntactical diversity of the output candidates.

3. In `random` selection, as the name suggests, random candidates are selected from the beam search, which may or may not produce diverse candidates.

transformations/diverse_paraphrase/submod/submodular_funcs.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+model = None
+
+
+def trigger_dips():
+    global model
+    def unk_init(x):
+        return torch.randn_like(x)
+    model = GloVe('6B', unk_init=unk_init)
```
kaustubhdhole, line 30: A lot of things seem to be consuming memory here - this can be reduced to 50 dim instead of the default 300.

ashutoshml: Updated!

AbinayaM02:
Hi @ashutoshml : Since the PR uses some memory heavy libraries, the build is failing. I'm looking at options to resolve this smoothly. It might take while!

PR #95, opened 2021-06-30
feat: a yes or no question filter

wwydmanski:

PR #96, opened 2021-06-30
Suspecting paraphraser

wwydmanski:
This paraphraser transforms a yes/no question into a tag one.
     
 Example: "Did the American National Shipment company really break its own fleet?" 
 -> "The American National Shipment company really broke its own fleet, didn't it?"


transformations/suspecting_paraphraser/README.md
```diff
@@ -0,0 +1,18 @@
...
+Author Affiliation: Jagiellonian University
+
+## What type of a transformation is this?
+This transformation acts like a paraphraser and makes lexical substitutions.
+
+## What tasks does it intend to benefit?
+This transformation would benefit tasks that are focused on conversational question answering.
+
+## What are the limitations of this transformation?
+The pronouns used in the endings aren't always on point - mostly because of ambiguity. As such, this transformation shouldn't be used for tasks that require their correctness.
```
vgtomahawk, line 18: Are you referring to the pleonastic "it" here? Could you explain this through an example so that what you mean by "pronouns used in the endings" will be clarified better [and also the limitation would be overall explained better]

wwydmanski: Yes, what I meant was that while this paraphraser tries to be consistent with the pronouns, there are some cases in which it will give an unexpected result.

Examples:
 - Did he drink that tea? -> He drank that tea, didn't he? (The pronoun gets recognized correctly)
 - Has Tesla finished high school? -> Tesla has finished high school, didn't it? (because the word Tesla gets recognized as an organization entity)
 - Was Sally here yesterday? -> Sally was here yesterday, wasn't [she/he/it]? (Chosen at random, because any pronoun can match here. By default, names that are recognized as strongly masculine or feminine will get 90% chance of selecting corresponding pronoun, but it's configurable).

Do you think that I should include those examples in the readme?

vgtomahawk: Certainly, you should, these will be really helpful! Yeah I get it better why they would now, since finding and assigning both gender and PER vs ORG would be noisy. In addition to the examples, it would be great if you could mention these underlying reasons too.

wwydmanski: OK, I've update readme with those examples. Thank you for the comment!

transformations/suspecting_paraphraser/README.md
```diff
@@ -1,5 +1,8 @@
 # Suspecting Paraphraser 🦎  + ⌨️ → 🐍
-This paraphraser transforms a yes/no question into a subjunctive one, which helps to add more question-specific variety to the dataset.
+This paraphraser transforms a yes/no question into a subjunctive one, which helps to add more question specific informality to the dataset.
```
vgtomahawk, line 2: You've explained it the phenomenon anyway but if you can link in some wiki article or ref for subjunctive it might help more interested readers dig deeper.

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -41,72 +42,78 @@ def __init__(self, seed=0, max_output=1, pronoun_mod=0.9):
...
-                elif noun_gender in ['female', 'mostly_female']:
+                elif noun_gender in ["female", "mostly_female"]:
                     prob = [self._pronoun_alt] * len(self.pronouns)
                     prob[Pronouns.SHE] = self.pronoun_mod
 
             pronoun = np.random.choice(self.pronouns, p=prob)
-        if modal=='have' and (pronoun in ['he', 'she', 'it']):
-            modal = 'has'
+        if modal == "have" and (pronoun in ["he", "she", "it"]):
+            modal = "has"
```
vgtomahawk: Can would, should and could also be potential modals to handle here?

transformations/suspecting_paraphraser/README.md
wwydmanski: I've changed this name into a more fitting one (tag question) and added a source

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,186 @@
...
+        beginning.extend(text[2:verb_position])
+        sentence = nltk.tokenize.treebank.TreebankWordDetokenizer().detokenize(
+            beginning + rest_of_sentence
+        )
+
+        first_verb = doc[verb_position]
+        if modal == "did":
+            demodded = first_verb._.inflect("VBD")
+        else:
+            demodded = modal + " " + str(first_verb)
```
vgtomahawk, line 80: Could you add comments here explaining what the VBD pos tag means, and what this if else does in general [I guess either use modal + raw form or just the inflected form, but might help readers understannd/extend this easily]

vgtomahawk (commented):
Correctness: I think things seem good on this front

Interface: I think the interface chosen here [QuestionAnswerOperation] is apt and pertinent. The tasks also make sense. In theory, this could even be a SentenceOperation, but it applies only to questions which makes the current scheme more sensible.

Specificity: Applies to yes/no questions.

Novelty: Yes, this is novel w.r.t NL-Augmenter.

Test Cases: 9 Test cases are present in `test.json`, sufficient as per requirements.

Adding New Libraries: The new libraries pertain to inflection/gender detection, and have been added into requirements.txt.

Evaluating Robustness:   A robustness evaluation [i.e evaluating b4/after the transformation on the standard dataset-model pair for the chosen interface] has not been done and is pending.

transformations/suspecting_paraphraser/README.md
```diff
@@ -0,0 +1,27 @@
+# Suspecting Paraphraser 🦎  + ⌨️ → 🐍
+This paraphraser transforms a yes/no question into a ![tag question](https://www.englishclub.com/grammar/tag-questions.htm), which helps to add more question specific informality to the dataset. 
+
+Example: "Did the American National Shipment company really break its own fleet?" -> "The American National Shipment company really broke its own fleet, didn't it?"
```
vgtomahawk, line 4: Is there an easy way to detect which qn is a yes/no question , so that we can only apply this transformation in those cases? [Was wondering how we would apply this transformation at scale on a corpus of questions]

requirements.txt
```diff
@@ -8,6 +8,10 @@ torch
 num2words==0.5.10
 word2number==1.1
 
+# for SuspectingParaphraser
+pyinflect==0.5.1
+gender-guesser==0.4.0
+
```
kaustubhdhole, line 14: Hi @wwydmanski, it would be necessary to look for a substitute library to gender-guesser which uses MIT/Apache 2 or a similar license to merge this PR.

transformations/suspecting_paraphraser/README.md
wwydmanski: I made a filter for this - #95

requirements.txt
wwydmanski: Will do, thanks for letting me know

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __init__(self, seed=0, max_output=1, pronoun_mod=0.9):
+        super().__init__(seed)
+        np.random.seed(seed)
+        nltk.download("punkt")
+
+        self.seed = seed
+        self.max_output = max_output
+
+        self.nlp = spacy.load("en_core_web_sm")
```
aadesh11, line 38: Please take master pull, we are loading spacy only once. Hence, this line can be replaced with:

where 'spacy_nlp' is present in 'initialize file.

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_output=1, pronoun_mod=0.9):
+        super().__init__(seed)
+        np.random.seed(seed)
+        nltk.download("punkt")
+
+        self.seed = seed
+        self.max_output = max_output
```
aadesh11, line 36: Duplicate code: Better to use 'max_outputs' present in the parent class and initialize it via super call.

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+        self, context: str, question: str, answers: [str]
+    ) -> List[Tuple[str, str, List[str]]]:
+        if not self._filter_phrase(question):
+            return [(context, question, answers)]
+
+        paraphrased = self._transform(question)
+        return [(context, paraphrased, answers)]
+
+
+if __name__ == "__main__":
```
aadesh11, line 161: Better to comment out this code.

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class Pronouns:
+    HE = 0
+    SHE = 1
+    IT = 2
+    THEY = 3
+
+
+class SuspectingParaphraser(QuestionAnswerOperation):
```
aadesh11, line 25: Can you please write 2-3 lines of comment for this class and also in a function wherever requires? So that any user will get proper clarity instead of going through the whole code.

requirements.txt
```diff
@@ -8,6 +8,10 @@ torch
 num2words==0.5.10
 word2number==1.1
 
+# for SuspectingParaphraser
```
aadesh11, line 11: Instead of directly putting any new package specific to your transformation, better to create a separate 'requirements.txt' file under your transformation folder.

transformations/suspecting_paraphraser/transformation.py
wwydmanski: I've deleted this section altogether as it was redundant

transformations/suspecting_paraphraser/transformation.py
wwydmanski: Done!

transformations/suspecting_paraphraser/transformation.py
wwydmanski: Done, thanks

transformations/suspecting_paraphraser/README.md
vgtomahawk: Thanks very much for addressing this!! Sorry I missed replying on this earlier....

vgtomahawk (approved):
Correctness: I think things seem good on this front

Interface: I think the interface chosen here [QuestionAnswerOperation] is apt and pertinent. The tasks also make sense. In theory, this could even be a SentenceOperation, but it applies only to questions which makes the current scheme more sensible.

Specificity: Applies to yes/no questions. Concomitant filter to check for yes/no questions also added via pull request 95..

Novelty: Yes, this is novel w.r.t NL-Augmenter.

Test Cases: 10 Test cases are present in test.json, more than sufficient as per requirements.

Adding New Libraries: The new libraries pertain to inflection/gender detection, and have been added into requirements.txt.

kaustubhdhole (approved):

transformations/suspecting_paraphraser/transformation.py
```diff
@@ -0,0 +1,203 @@
...
+            "will": "won't {}",
+        }
+
+        self.pronoun_mod = pronoun_mod
+        self._pronoun_alt = (1 - pronoun_mod) / (len(self.pronouns) - 1)
+
+    def _transform(self, question):
+        text = nltk.word_tokenize(question)
+        doc = self.nlp(question)
+        modal = str(doc[0]).lower()
```
kaustubhdhole, line 54: Oh I missed this - why not loop through all docs instead of only the first one?

wwydmanski: It must be a leftover error, thank you for noticing this

wwydmanski: The [0] refers to the first token in the sentence, the name of the variable is a bit misleading here.

kaustubhdhole: Okay! Thanks for the confirmation!

PR #97, opened 2021-06-30
feat: yes/no question to statement paraphraser

wwydmanski:
_Did John drink my tea?_ -> _John drank my tea_

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 checklist==0.0.10
 spacy==2.2.4
+pyinflect==0.5.1
 
 # for back_translation
 torch
--- a/transformations/question_to_statement/transformation.py
+++ b/transformations/question_to_statement/transformation.py
@@ -1,0 +1,119 @@
+from typing import Tuple, List
+import nltk
+from pyinflect import getInflection
+import spacy
+import numpy as np
+
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+
+class QuestionToStatement(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TAGGING,
+    ]
+
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_output=1):
+        super().__init__(seed)
+        np.random.seed(seed)
+        self.seed = seed
+        self.max_output = max_output
+
+        self.nlp = spacy.load("en_core_web_sm")
+
+    def _transform(self, question):
+        text = nltk.word_tokenize(question)
+        doc = self.nlp(question)
+        modal = str(doc[0]).lower()
+        token = doc[0]
+
+        modal_position = token.head.i
+        for tok in token.head.lefts:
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,5 +1,6 @@
 checklist==0.0.10
 spacy==2.2.4
+pyinflect==0.5.1
```
Nickeilf, line 3: We recommend creating a separate requirements.txt in the same transformation folder so that the main requirements.txt is not changed often (which might change the version of the dependency for specific transformation)

transformations/question_to_statement/transformation.py
```diff
@@ -0,0 +1,121 @@
...
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+
+class QuestionToStatement(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TAGGING,
```
Nickeilf, line 20: TEXT_TAGGING might not be suitable here. The Transformation should inherit TaggingOperation if it's for TEXT_TAGGING task.

transformations/question_to_statement/transformation.py
```diff
@@ -0,0 +1,121 @@
...
+            return False
+
+        return True
+
+    def generate(self, sentence: str):
+        if not self._filter_phrase(sentence):
+            return [sentence]
+
+        paraphrased = self._transform(sentence)
+        return [sentence, paraphrased]
```
Nickeilf, line 92: I would recommend only returning the transformed (I would avoid using "paraphrase" for this transformation as it changes the semantics) sentence if can be found.

Nickeilf: Also, don't forget to make changes to the test.json as well.

transformations/question_to_statement/transformation.py
```diff
@@ -0,0 +1,121 @@
...
+        return sentence
+
+    def _filter_phrase(self, question):
+        try:
+            if question.strip()[-1] != "?":
+                return False
+        except IndexError:
+            return False
+
+        if " or " in question:
```
Nickeilf, line 77: Why a sentence with "or" should be avoided?

transformations/question_to_statement/transformation.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            first_verb = doc[modal_position]
+            demodded = modal + " " + str(first_verb)
+        sentence = sentence.replace(str(first_verb), str(demodded)).replace("?", "")
+
+        return sentence
+
+    def _filter_phrase(self, question):
+        try:
+            if question.strip()[-1] != "?":
```
Nickeilf, line 72: I guess this might limit this transformation to only one sentence as an input. In the current implementation, would it fail with this example as the input:
"I met Chris yesterday. Did Sally finally return the french book to Chris?"

Nickeilf: I guess maybe a decent way to fix this is to first split the input into sentences, and then apply the transformation for each sentence, and then concatenate them together.

kaustubhdhole:
@wwydmanski would you like to address the above comments?

kaustubhdhole:
Okay @wwydmanski I am temporarily closing this PR to make tracking PRs easier. Please feel free to reopen the PR!

PR #98, opened 2021-06-30
Normalize to ASCII encoding

PhilipEHausner:


PhilipEHausner:
Hello,
I cannot really figure out why tests were failing for our contribution. Tests are not throwing any errors when executed on my machine. Sadly, I also cannot access the test results anymore, but it was some OS error as far as I remember for which I could not see an immediate solution.
Can you please rerun the tests for the PR?

transformations/normalize_to_ascii/requirements.txt
```diff
@@ -0,0 +1 @@
+text-unidecode==1.3
```
kaustubhdhole, line 1: Hi @PhilipEHausner, thank you for your efforts. This is GPL licensed - please use either an MIT or Apache 2 Licensed.

PhilipEHausner: Hello @kaustubhdhole and thank you for the quick reply. I had a look at https://pypi.org/project/text-unidecode/ which states that the package can be either distributed under the artistic license or GPL, even stating that if GPL licensing is OK within a project, unidecode (instead of text-unidecode) can be used. Since I assumed (after some research) that the artistic license is a permissive license, I assumed it would be also in line with the restrictions for NL-Augmenter. Did I miss something here, or do you consider the artistic license to not be permissive?

kaustubhdhole: Hi @PhilipEHausner, apologies for the delayed reply. The artistic license has restrictions and we would suggest you to use other licensed packages - especially MIT or Apache 2.

kaustubhdhole:
Hi @PhilipEHausner, I am closing this PR for now since we are tracking all of them. Feel free to reopen it later.

PR #100, opened 2021-06-30
Adding close homophones swap

KaizhaoLiang:
This transformation randomly swaps words that sound similar to each other to mimic the online language behaviors among young people or unintentional spelling or dictating mistakes. For example tea <-> tee, pray <-> prey and etc.


transformations/close_homophones_swap/transformation.py
```diff
@@ -0,0 +1,59 @@
...
+from spacy.tokens import Doc
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+
+def close_homophones_swap(text, corrupt_prob, seed=0, max_output=1):
+    random.seed(seed)
+    nlp = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 19: Hi @KaizhaoLiang thank you for the changes!

It might be better if spacy is loaded once (in the constructor).

requirements.txt
```diff
@@ -8,6 +8,9 @@ torch
 num2words==0.5.10
 word2number==1.1
 
+# for CloseHomophonesSwap
+SoundsLike==0.0.11
+
 # for test cases
```
kaustubhdhole, line 14: And now to avoid conflicts, you can add this in a new requirements.txt in the folder of "close_homophones_swap"

KaizhaoLiang: FIxed, thanks for the suggestion.

transformations/close_homophones_swap/transformation.py
KaizhaoLiang: Updated, thanks for the suggestion

transformations/close_homophones_swap/transformation.py
```diff
@@ -0,0 +1,59 @@
...
+from spacy.attrs import LOWER, POS, ENT_TYPE, IS_ALPHA
+from spacy.tokens import Doc
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+
+def close_homophones_swap(text, corrupt_prob, seed=0, max_output=1, nlp = None):
+    random.seed(seed)
```
kaustubhdhole, line 18: I'm not sure but I think random.seed should go inside the for loop so that the seed would work for num_outputs > 1?

kaustubhdhole:
I think this PR looks great overall and satisfies the review criteria- I would suggest one thing though since I see in the test cases some of the homophone pairs are even (a->A.) and (as -> As). I think you could avoid such cases (in which the word is basically the same). 

Thanks for your changes!

transformations/close_homophones_swap/transformation.py
KaizhaoLiang: Hum, I am not sure what the expected behavior should be here? I thought the idea was to have something to set the seed such that the output can be deterministic for the test cases, but we will want something random in the actual generation where num_outputs > 1. If we set the seed every time we generate a new sample, wouldn't reset the random state and output the same thing?

kaustubhdhole: oh my bad! What I meant was random.seed works only once so if you are looping for max_output > 1, you might get different results everytime. I think the correct way would be replace random.choice with random.choices below?

KaizhaoLiang: My understanding of the expected behavior is when max_output > 1, we are supposed to output multiple versions of the augmented sentence from a single source sentence. The current implementation allows it to swap at different locations on the same sentence.

The one you are purposing works, but essentially doing that would make it swap at only a single set of locations for each sentence. so the output could be less diversified.

kaustubhdhole: Okay, but why not choose the tokens using random.choices(tokens) then? I'm trying to ensure that the same results turn up for num_outputs > 1 and one seed.

KaizhaoLiang: I think it does output the same results everytime we run with the same seed, given any max_outputs > 1, since every random function is controlled by random.seed, hence the pesudo random sequence would be the same.

To choose tokens to be swapped with random.choices will definitely look more elegant though. But since we are calling the homophones similarity function on each token, it would be more straightforward to implement it with a for loop, which loops through each token.

KaizhaoLiang: I just ran test with max_output = 3 for multiple times, it looks like it does output same set of results everytime.

kaustubhdhole:
Thank you very much for your changes @KaizhaoLiang and providing clarifications! :)

kaustubhdhole (approved):

transformations/close_homophones_swap/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, 
+text generation, etc. 
+
+## Previous Work
+1) There has been some recent work studying how to create human-like behavior using phonetic information:
+```bibtex
+@article{eger2019text,
+  title={Text processing like humans do: Visually attacking and shielding NLP systems},
```
samsontmr, line 21: This is the right author but not the right paper.

transformations/close_homophones_swap/transformation.py
samsontmr: I had the same question a while ago and it turns out that you only need to call `random.seed()` once. The next `random()` calls return different numbers but it will always be the same sequence of numbers for that seed.

samsontmr (changes requested):
Other than the reference, this looks good!

transformations/close_homophones_swap/README.md
KaizhaoLiang: Hi @samsontmr, the augmentation was indeed inspired by this paper. Maybe one of the citations in this paper is more relevant, where they actually swapped out similarly sounding words to bypass censorship. Do you have another paper in mind?

Chaya Hiruncharoenvate, Zhiyuan Lin, and Eric Gilbert. 2015. Algorithmically bypassing censorship on sina weibo with nondeterministic homophone substitutions. In Proceedings of the Ninth International Conference on Web and Social Media,
ICWSM 2015, University of Oxford, Oxford, UK,
May 26-29, 2015, pages 150–158.

samsontmr: I had this paper in mind actually (by the same author): https://aclanthology.org/2020.aacl-main.79/

KaizhaoLiang: Thanks! this one indeed looks more relevant. Let me update the reference.

samsontmr: Thanks, saw the update. Could you use the ACL bibtex instead of the arxiv? You can get it from the link I sent.

samsontmr: Additionally, if you would like to implement the Chinese version (from the ICWSM paper) that would be great! You can either extend it in this PR or a new one, both ways are fine :)

samsontmr (approved):

kaustubhdhole:
Thank you very much for your changes. Your changes are merged. As the field of NLP has largely focussed on high-resource languages and domains, if possible, we also encourage you to add a transformation in your mother tongue and/or a low-resource language which might need more attention from the NLP community. This is not mandatory.

PR #102, opened 2021-06-30
Adding homophonic transformation for a sentence

SuchitraDubey:
…EADME.md


kaustubhdhole:
You should add why this transformation different from the [close_homophones_swap]( https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/close_homophones_swap)

tuetschek (changes requested):
I'd have three additional points:
* The readme should include your name as the author, plus an example command to use the transformation would be good.
* Isn't there a library of some sorts to provide you with the homophones? I'm not sure that this list of 1600 entries is practical. #143 is using the [jellyfish](https://pypi.org/project/jellyfish/) library – wouldn't that be possible somehow?
* If you do use the included list, it's inefficient to load it from the file on each use. I think that the loader should be called in the constructor.

SuchitraDubey:
> You should add why this transformation different from the [close_homophones_swap](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/close_homophones_swap)

 This transformation is different from "close-homophones_swap" in the following ways:
**1)** It changes the complete sentence (replace almost every word with its homophone).
**Ex:** "Virat Kohli made a big hundred against Australia ." --> "Voyrat kuwlYy meyd ey big hundr'd 'geynst 'streyly' ."
**2)** Those words whose homophone is not present in the homophone dictionary, this transformation breaks the word into two sub-words and generates homophone from those sub-words.
**Ex:** "Virat" => "Vi" + "rat" => "Voy" + "rat" => "Voyrat".
**3)** As we can see in the test.json of close_homophones_swap, some input and output sentences are the same because containing words don't have homophones (in English vocab). This transformation considers only non-vocab words for homophone replacement. Such cases are very few in this transformation.

I have added all the points mentioned above, in the README.md file.

SuchitraDubey:
> I'd have three additional points:
> 
> * The readme should include your name as the author, plus an example command to use the transformation would be good.
> * Isn't there a library of some sorts to provide you with the homophones? I'm not sure that this list of 1600 entries is practical. [Added phonetic match filter #143](https://github.com/GEM-benchmark/NL-Augmenter/pull/143) is using the [jellyfish](https://pypi.org/project/jellyfish/) library – wouldn't that be possible somehow?
> * If you do use the included list, it's inefficient to load it from the file on each use. I think that the loader should be called in the constructor.

@tuetschek 
Hi Ondrej,
I have made the following changes in the PR:
**1)** Added author information, added an example command to use the transformation.
**2)** Completely agree with you, 1600 entries are not much, so I have changed the transformation idea a bit. Now the transformation is replacing each word with its homophone (non-vocab words of English).
**Ex:**  "Virat Kohli made a big hundred against Australia ." --> "Voyrat kuwlYy meyd ey big hundr'd 'geynst 'streyly' ."
**3)** New homophones list is comparatively bigger and contains close to 125K words and sub-words homophones.
**4)** Still if a word doesn't have a homophone present in the homophones list, this transformation breaks that word into two sub-words and generates homophones from those sub-words. This will be useful in the case of rare words, a person's name, organization's name, whose homophones are not present in the list.
**5)** New homophone list is loaded in the constructor.

tuetschek:
@SuchitraDubey Thanks! However, are you sure what you're changing these to are "homophones"? The current version looks more like you've replaced the words with some kind of phonetic representation (i.e. pronunciation). Are you sure this makes sense? This basically completely swaps the English orthography for something else. I think this is very different from the original.

tuetschek:
@SuchitraDubey This does look better but I'm still not sure the transformation looks realistic w.r.t. what you could encounter in real texts. It's not very similar to English orthography, e.g. because capitalization distinguishes different sounds -- `s` is for the `[s]` sound while `S` means `[ʃ]`.

In addition, where does your new 125k-long list come from? Looks like a some kind of phonemic representation, though I can't [recognize it on this list](https://en.wikipedia.org/wiki/Comparison_of_ASCII_encodings_of_the_International_Phonetic_Alphabet).  Would it make sense to find identical or similar pronunciations from it and group them together to get actual homophones? Or was that your original 1.6k list?

kaustubhdhole:
Okay, I am closing this PR since we are tabulating the PRs. Please feel free to reopen it later.

PR #106, opened 2021-07-01
Feature/adding hashtag transformation

ashish3586:


transformations/add_hashtag/transformation.py
```diff
@@ -0,0 +1,118 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+import random
+from spacy import load
+
+be_class_verb = ["is", "am", "are", "was", "were", "will", "shall"]
+subj_obj_list = ["i", "you","we", "they", "he", "she"]
+nlp=load('en_core_web_sm')
```
kaustubhdhole, line 8: This should be placed inside the init function - else it will loaded even when we don't load this generator.

transformations/add_hashtag/transformation.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            transformation_list.append(sentence + "".join(selected_hastag))
+    return transformation_list
+
+
+class HashtagGeneration(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING
```
kaustubhdhole, line 84: This won't work for tagging since more tokens are being added and hence labels/tags would need to change too.

transformations/add_hashtag/transformation.py
```diff
@@ -0,0 +1,118 @@
...
+    return hash_tag_list
+
+
+def get_hash_tags(sentence, seed=0, max_outputs=1):
+    # method for adding hashtags to sentence
+    random.seed(seed)
+    hashtag_list = extract_hashtags(sentence)
+    transformation_list = []
+    for _ in range(max_outputs):
+        selected_hastag = random.sample(hashtag_list, random.randint(1, len(hashtag_list)))
```
kaustubhdhole, line 72: I think you might need to call random.seed before this line.

transformations/add_hashtag/test.json
```diff
@@ -0,0 +1,49 @@
...
+        "sentence": "I love domino's pizza .#Pizza#Domino'SPizza"
+      }]
+    },
+    {
+      "class": "HashtagGeneration",
+      "inputs": {
+        "sentence": "Many people like T20 cricket these days ."
+      },
+      "outputs": [{
+        "sentence": "Many people like T20 cricket these days .#People#Cricket#T20"
```
kaustubhdhole, line 21: Would it be a good idea to tag even hypernyms or some words which are not in the original sentence?

transformations/add_hashtag/test.json
```diff
@@ -0,0 +1,49 @@
+{
+  "type": "hashtag_generation",
```
kaustubhdhole, line 2: this type should match the folder name - in your case: "add_hashtag"

kaustubhdhole:
@ashish3586 please pull main and fix the test cases.

transformations/add_hashtag/test.json
ashish3586: Completed all suggested changes.

PR #107, opened 2021-07-03
Adding numeric filter

SuchitraDubey:


filters/numeric/filter.py
```diff
@@ -0,0 +1,67 @@
...
+class TextContainsNumberFilter(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self, has_digit=True, has_word=True):
+        super().__init__()
+        self.nlp = load("en_core_web_sm")
+        self.has_digit = has_digit
+        self.has_word = has_word
+        self.numbers_in_words = ["one", "two", "three", "four", "five", "six", "seven", "eight",
```
kaustubhdhole, line 30: Isn't this an extension of the "keywords" filter?

filters/numeric/README.md
```diff
@@ -0,0 +1,21 @@
...
+Email: suchitra27288@gmail.com
+Affiliation: Acko
+
+## Why is measuring performance on this split important?
+This filter can be used to create splits. Filtering out and testing on examples 
+contain numeric values can provide feedback for improving training data 
+accordingly.
+
+python evaluate.py -f TextContainsNumericFilter -p=1
+
```
kaustubhdhole, line 16: You will need to run this on your PC.

kaustubhdhole:
@SuchitraDubey @ashish3586

filters/numeric/filter.py
SuchitraDubey: Yes, Kaustubh, you are right. We can see it as an extension of a keywords filter.  But from my understanding, the "Keywords" filter is pretty general in nature and is best suited to filter some special words like "Product name", "Person name" etc (Where filter words are very limited and creating such list is straight forward eg. ["Policy", "Insurance", "Cover"]). On the other hand, in some cases, such lists are relatively long and should be carefully created such that no member word is left out. Now from the NL-aug user's point of view, who need a quick, complete and readymade solution to extract utterances contains numeric data so that he/she can do some future experiments, It could be a useful filter.

filters/numeric/README.md
SuchitraDubey: Working on it.

SuchitraDubey: numeric_filter_eval_results
Added two evaluation results, one with SNLI and another one with MNLI.

kaustubhdhole: Thanks @SuchitraDubey for sharing the results.

kaustubhdhole (approved):

PR #108, opened 2021-07-03
[CI/CD] - Pre commit hooks 

AbinayaM02:
As part of this PR, the following features are added,
- The following hooks are added:
  - flake8
  - isort
  - pytest
  - pre-commit-hooks
      -   trailing-whitespace
      -   check-yaml
      -   check-ast
      -   check-json
      -   check-merge-conflict
      -   detect-private-key
- Added pytest-cov for coverage report generation (can be utilized later)
- Added pyproject.toml to have all the related configs
- Added relevant rules in flake8/.pre-commit-config.yaml to suit the project (can be changed accordingly)
- Added Makefile to make executing commands easy for the user
- Added cache to install libraries the github actions


.github/workflows/nl-action.yml
```diff
@@ -18,12 +18,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.7
+    - uses: actions/cache@v2
```
aadesh11, line 21: Great, with this cache we are able to reduce dependence installation time from 3m 50sec --> 56sec. Hence, reducing overall build run time.

Makefile
```diff
@@ -0,0 +1,19 @@
+# Styling
```
aadesh11, line 1: Maybe we could add a rule for Git hook also, i.e to install pre-commit hook and to run it on all files? That way it will be easier for the users to just use one rule to run all the formatting stuff defined in pre-commit-config.

.github/workflows/nl-action.yml
AbinayaM02: Nice! Works as expected then.

Makefile
AbinayaM02: Sure, let me find a way to do that.

aadesh11: Check this out:
https://gist.github.com/mmphego/dc933ba1bf9630fe57e46e0da5ef7820#file-makefile

kaustubhdhole (approved):

PR #111, opened 2021-07-07
add mix transliteration task

MukundVarmaT:
This transformation converts randomly picks words in the text from their original script to their corresponding English transliteration. It can be used to train/test multilingual models to improve/evaluate their ability to understand complete or partially transliterated text.

Authors: Tanay Dixit (tanay.dixit@smail.iitm.ac.in), Mukund Varma T (mukundvarmat@gmail.com)

### DIFF
--- a/transformations/mix_transliteration/requirements.txt
+++ b/transformations/mix_transliteration/requirements.txt
@@ -1,0 +1,6 @@
+indictrans @ git+https://github.com/libindic/indic-trans.git
+pbr
+six
+future
+cython>=0.24.0a0
+scipy>=0.13.3
--- a/transformations/mix_transliteration/README.md
+++ b/transformations/mix_transliteration/README.md
@@ -1,0 +1,66 @@
+# Mix-Transliteration 🦎  + ⌨️ → 🐍
+
+This transformation converts randomly picks words in the text from their original script to their corresponding english transliteration.
+It can be used to train/test multilingual models to improve/evaluate their ability to understand complete or partially transliterated text.
+
+Author name: Tanay Dixit, Mukund Varma T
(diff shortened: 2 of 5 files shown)


MukundVarmaT:
Hi @kaustubhdhole, 

Thanks for pointing out the licencing issue. We were unable to find a suitable package under the MIT/Apache Licence which also satisfies the requirements of this transform. Therefore would it be alright if we reimplement whatever is required from [indic-trans](https://github.com/libindic/indic-trans) and include them as a script in this transform (essentially all we need to do is include a large lookup table to convert from any Indian language to English, which I believe is pretty straightforward).

MukundVarmaT:
@kaustubhdhole 
Sorry for the delay from our end. Based on our prev conversations, we have updated the scripts as required. We hope it satisfies your concerns.

MukundVarmaT:
@kaustubhdhole 
Update: We ran into [3aransia](https://github.com/3aransia/3aransia) (licenced under Apache 2.0) and we have updated our scripts to use the same. Kindly let us know if the changes will address your concerns. 

Thanks

transformations/mix_transliteration/requirements.txt
```diff
@@ -0,0 +1,8 @@
+aaransia
+numpy==1.18.1
+pandas==1.0.1
+python-dateutil==2.8.1
+pytz==2019.3
+six==1.14.0
+unicode==2.7
+Unidecode==1.1.1
```
kaustubhdhole, line 8: Unidecode is also a GPL one..

MukundVarmaT: Sorry, it's an oversight from my end. The scripts only require 3aransia which is under Apache Licence. I will update the requirements.txt in a bit.

PR #123, opened 2021-07-11
Toxicity Filter

saqibns:
Added a filter to select a subset of sentences based on toxicity.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,6 +29,3 @@ black==21.5b2
 pre-commit==2.13.0
 flake8==3.9.2
 isort==5.9.1
-
-# for toxicity detection
-detoxify==0.2.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,3 +29,6 @@ black==21.5b2
 pre-commit==2.13.0
 flake8==3.9.2
 isort==5.9.1
+
+# for toxicity detection
+detoxify==0.2.2
--- a/filters/toxicity/README.md
+++ b/filters/toxicity/README.md
@@ -1,0 +1,21 @@
+## Toxicity filter
+
+## What type of a filter is this?
+
+This filter filters an example text which has a toxicity value matching a particular threshold.
+Author: Saqib N. Shamsi
+
+Toxicity labels supported:
+* ToxicityTypes.TOXICITY
+* ToxicityTypes.SEVERE_TOXICITY
+* ToxicityTypes.OBSCENE
+* ToxicityTypes.IDENTITY_ATTACK
+* ToxicityTypes.INSULT
+* ToxicityTypes.THREAT
+* ToxicityTypes.SEXUAL_EXPLICIT
+
+## Why is this filter important?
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -29,3 +29,6 @@ black==21.5b2
 pre-commit==2.13.0
 flake8==3.9.2
 isort==5.9.1
+
+# for toxicity detection
```
Nickeilf, line 33: Please make a new `requirements.txt` file in this filter directory and move detoxify to that file. To make it not too heavy, the main requirements should only include dependencies that are used for most transformation/filters.

[Example](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/close_homophones_swap)

Nickeilf (commented):
Very interesting filters! I wonder if it's possible to also implement multiple conditions, such as `TOXICITY within the range of [0.2, 0.8]`, `TOXICITY > 0.9 and INSULT > 0.75`?

requirements.txt
saqibns: Sure thing

saqibns:
Thank you! Yeah, that makes sense. I'll update the code to enable multiple conditions.

saqibns:
@Nickeilf I was wondering if I should implement multiple conditions the way they have been implemented for the [token-amount filter](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/filters/token-amount/filter.py) or should I create another class that composes multiple filters just like [torchvision's API](https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.Compose), which allows users to compose multiple augmentations?

Nickeilf:
> @Nickeilf I was wondering if I should implement multiple conditions the way they have been implemented for the [token-amount filter](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/filters/token-amount/filter.py) or should I create another class that composes multiple filters just like [torchvision's API](https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.Compose), which allows users to compose multiple augmentations?

The idea of torchvision Compose is brilliant! I think it's definitely a good choice to apply multiple filters/transformations for the end-users! I guess the only restriction is that the sequences of filters should be the same class (e.g. all SentenceOperation) to make sure the input/output are the same.

saqibns:
Great! I'll write one for both transformations and filters. Which place would be more appropriate to put the files corresponding to the class?

Nickeilf:
> Great! I'll write one for both transformations and filters. Which place would be more appropriate to put the files corresponding to the class?

I think we can have a separate file to have the torchvision Transform style implementation, and then we can import it in dataset.py when we want to apply multiple transformations/filters at the same time.

@kaustubhdhole here.

kaustubhdhole:
It would be a great idea to have such a mechanism. Will help evaluate examples which feature multiple properties together.

kaustubhdhole:
@Nickeilf

Nickeilf (approved):

kaustubhdhole (approved):

PR #124, opened 2021-07-12
Lost in translation

mgobrain:
`LostInTranslation` is a generalization of the `BackTranslation` transformation that allows for an arbitrary number of encode-decode layers. Languages with [Helsinki-NLP OpusMT](https://huggingface.co/Helsinki-NLP) models are supported.

Example:
```
Input: "Sentences with gapping, such as Paul likes coffee and Mary tea,
lack an overt predicate to indicate the relation between two or more 
arguments."

Layers (5): en -> es -> de -> zh -> fr -> ru -> en

Output: "Because Paul loves coffee and tea, there is a gap in the verdict,
and there is no clear prerequisite for the relationship between two or 
more arguments."
```
LiT can be used to simulate a tailored amount of corruption or information loss to an input sentence by changing the number of layers in the transformation.


transformations/lost_in_translation/transformation.py
```diff
@@ -0,0 +1,110 @@
...
+    # call shell, which applies the layers
+    return [
+            self.shell(sentence,
+                       self.langs, self.how, self.layers,
+                       self.max_length, self.device
+                      )
+            ]
+
+# Sample code to demonstrate usage. Can also assist in adding test cases.
+# You don't need to keep this code in your transformation.
```
kaustubhdhole, line 90: This can be commented.

transformations/lost_in_translation/transformation.py
```diff
@@ -0,0 +1,110 @@
...
+import random
+from transformers import pipeline
+
+class LostInTranslation(SentenceOperation):
+  '''
+  Repeated translation of input text
+  '''
+  tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+  languages = ["en"]
+  heavy = True
```
kaustubhdhole, line 12: you can remove this as of now so that the build can be tested. I will add it later.

transformations/lost_in_translation/requirements.txt
```diff
@@ -0,0 +1 @@
+transformers==4.6.1
```
kaustubhdhole, line 1: Transformers is already present?

transformations/lost_in_translation/transformation.py
```diff
@@ -0,0 +1,110 @@
...
+    self.device = device
+    self.layers = layers
+    self.max_length = max_length
+
+  def encode_decode(self, text, lang, max_length=600, device=-1):
+    # translate and un-translate
+    # using Helsinki-NLP OpusMT models
+    encode = pipeline(
+        'translation_en_to_{}'.format(lang),
+        model="Helsinki-NLP/opus-mt-en-{}".format(lang),
```
kaustubhdhole, line 46: Also, you might want to add in the README why you've chosen these models & their papers if available.

kaustubhdhole:
I think this is an interesting idea. I was actually thinking to add a similar task to BIG-BENCH. Anyways, I think this looks great and since your transformation is going to take up a lot of memory already and additional backtranslations would mostly keep deviating from the original meaning, do you think adding a semantic similarity check (using one of the HF models) would be a good option?

mgobrain:
I'd be happy to add a semantic similarity check. What would be the best way to implement this feature? Should the `generate()` method return a score?

kaustubhdhole:
Hi @mgobrain, sorry for the delayed reply. The generate method would still return the sentences, albeit filtered. So, maybe you generate 15 from this back to back backtranslation and then check how similar all 15 are with the source sentence and keep only those above a certain cut-off. (The similarity threshold can be configurable.) Maybe you could compare the backtranslation outputs at each step with the recent input (rather than source)?

mgobrain:
I added a similarity check using the rouge_score F-measure. This is calculated after each layer is applied and further backtranslation is aborted if the score for the stepwise text versus original input falls below a configurable `min_score`. The threshold defaults to -1 (off).

kaustubhdhole (approved):
Thanks for your effort @mgobrain!

PR #126, opened 2021-07-14
Generate Yes-or-No Question from Statement

cascadianblue:


kaustubhdhole:
Hi @cascadianblue, I think your PR seems a useful one and it is wonderful to see the rules that you've written! I think there is a clear change in meaning from what is a clear fact (or a declarative) to an interrogative. So, I would suggest to append a boolean "Yes" or a "No" to the end to keep the meaning intact.
eg. "Andrew returned the book."  --> "Did Andrew return the book? **Yes**"
"Andrew did not play the sitar." --> "Did Andrew play the sitar? **No**"

transformations/yes_no_question/transformation.py
```diff
@@ -0,0 +1,153 @@
...
+    def __init__(self, seed=0, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.detokenizer = TreebankWordDetokenizer()
+        self.nlp = spacy_nlp if spacy_nlp else spacy.load('en_core_web_sm')
+
+    def generate(self, sentence: str):
+        doc: Doc = self.nlp(sentence)
+
+        # Look for sentence verb head, starting with first token
+        verb_head: Token = doc[0]
```
kaustubhdhole, line 72: Maybe you could loop over all sentences instead of only the first?

transformations/yes_no_question/transformation.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+
+
+def uncapitalize(string: str):
+    if len(string):
+        return string[0].lower() + string[1:]
+    return ''
+
+
+def front_auxiliary(auxiliary: Token) -> str:
```
kaustubhdhole, line 24: Please add a function docstring: https://www.python.org/dev/peps/pep-0257/

cascadianblue:
> I would suggest to append a boolean "Yes" or a "No" to the end to keep the meaning intact.

Hi @kaustubhdhole, I was under the impression that not all transformations had to be paraphrases of the original meaning. Does *this* transformation have to implement the "yes"/"no" at the end of each sentence, or can we make *another* transformation that includes the confirming "yes" or "no"?

kaustubhdhole:
Hi @cascadianblue definitely! Either the meaning has to remain intact or it should be evident what type of example it produces - i.e.  the new class/label should be known if in case the label changes (eg. sentiment goes from positive to negative)! In the former case, it would be useful as a perturbation as well as a paraphraser. In the latter case, it would help in creating more examples for the task. Your particular transformation could add more value if you keep the meaning intact rather than tag it with generic (less useful) labels like declarative & interrogative.

cascadianblue:
On GitHub CI, I am getting some [weird behavior](https://github.com/GEM-benchmark/NL-Augmenter/runs/3335851239#step:6:1347) from `spacy`'s `en_core_web_sm` model, which is causing issues downstream. I left a debugging breadcrumb (see previous link), and it looks like the tokenization is off (e.g. `"isn't"` isn't getting split into `"is"` and `"n't"`, as it should) and `"isn't"` is getting read as a `PROPN` (proper noun) by the POS tagger.

cascadianblue:
> On GitHub CI, I am getting some [weird behavior](https://github.com/GEM-benchmark/NL-Augmenter/runs/3335851239#step:6:1347) from `spacy`'s `en_core_web_sm` model, which is causing issues downstream. I left a debugging breadcrumb (see previous link), and it looks like the tokenization is off (e.g. `"isn't"` isn't getting split into `"is"` and `"n't"`, as it should) and `"isn't"` is getting read as a `PROPN` (proper noun) by the POS tagger.

TL;DR: Spacy's NLP model seems to suddenly have become incapable of splitting up English contractions (e.g. `"isn't"`), and the one token comprising this contraction then gets read as a proper noun, apparently. This bug occurs in the CI run, but I can't seem to replicate it locally.

transformations/yes_no_question/transformation.py
```diff
@@ -0,0 +1,203 @@
...
+    def statement_to_question(self, sentence: Span) -> Union[str, None]:
+        """Given a statement (type: spacy Span), convert to corresponding
+        yes-or-no question.
+        """
+
+        # FIXME: This is a debugging breadcrumb
+        if sentence.text == "Dave isn't going to Seattle.":
+            raise Exception([(token.text_with_ws, token.dep_, token.pos_,
+                              token.morph) for token in sentence])
+
```
AbinayaM02, line 97: Hi @cascadianblue: This exception is causing the build to fail. Please remove it if it's used only  for debugging purpose.

AbinayaM02:
> TL;DR: Spacy's NLP model seems to suddenly have become incapable of splitting up English contractions (e.g. `"isn't"`), and the one token comprising this contraction then gets read as a proper noun, apparently. This bug occurs in the CI run, but I can't seem to replicate it locally.

The reason behind this behavior is because of the issue explained in #209. I have also provided the suggestion for a possible fix. Let me know if you need more information.

cascadianblue:
@AbinayaM02 thank you for the bugfix/advice re: the Spacy tokenizer.

AbinayaM02:
> @AbinayaM02 thank you for the bugfix/advice re: the Spacy tokenizer.

You're welcome. Btw, please add your transformation in the corresponding dictionary (in map_transformation["light"] since yours is a light transformation) in test/mapper.py for the workflow to pick up your transformation for testing.

cascadianblue:
Thanks for your help, @AbinayaM02 . Unfortunately, I am having trouble running the evaluation script (see issue #300). Could someone help to run it locally and add the results? Or resolve the issue?

AbinayaM02:
> Thanks for your help, @AbinayaM02 . Unfortunately, I am having trouble running the evaluation script (see issue #300). Could someone help to run it locally and add the results? Or resolve the issue?

The issue is with torchtext version in the diverse_paraphrase's requirements.txt. You can specify the version of the torchtext to be 0.9.1. Once that's done, also download the en_core_web_sm using the command,
```python
pip install https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
```

You should be able to run the evaluation :) [I tested the diverse paraphrase after specifying the torchtext version, it works]

kaustubhdhole:
I think this PR looks great! You might want to also add keywords and the robustness evaluation too to make it stronger. Besides, it would be really interesting to see whether tasks like Question-Answering actually benefit or falter from such conversion. Also nudging @AbinayaM02 here.

kaustubhdhole (approved):

AbinayaM02 (approved):
Hi @cascadianblue: Thank you for an interesting transformation. I have fixed the pytest failure.

Here is my general observation,
1. Transformation is novel and interesting and it's not implemented already.
2. Test cases are added.
3. New requirements are specified.
4. README is updated with all the necessary information.

I see that docstrings don't have param explanations. Please add that wherever it is missing and we're good to go. Approving the PR!

Edit:
I'm merging this PR since the build is failing on the main branch and the fix for the issue is part of this PR. Docstrings can be added later as separate PR.

PR #128, opened 2021-07-14
Proper noun to noun

simonmeoni:
With @maurice360  we propose an augmenter transforming proper noun to noun.

example: John has a car -> he has a car 

The actual commit doesn't work yet but we already implement the followed basic functionalities : 
1. detection of proper noun with a ner model
2. replace detected entities with the <mask> token
3. use another model to replace <mask> with a pronoun

In this version, we can replace several entities in a sentence. However, the model tends to replace a certain entity with another proper noun.


kaustubhdhole:
reminder @vgtomahawk

vyraun:
Hi, I am one of the assigned reviewers for this PR. This transformation isn't in great shape at the moment. The readme has no relevant information, so please fix that. The test cases are <5, and there are no keywords in the code. Any robustness evaluation (optional: https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation) is also missing. So, overall more clarity is needed before any detailed comments could be provided/addressed.

vgtomahawk:
My general comments/remarks on this transformation, and how it presents advantages over using a mere name-based lexicon, making this transformation valuable...

One advantage is that in this way, it can use strong signals from the context to override the lexicon, which is very cool
 its natural that BERT or any model implicitly also learns the lexicon that traditional lexicons contain i.e in the absence of context or pretty uninformative contexts, they will give consistent, same answers for pronouns like a lexicon lookup

For instance, Nikita is originally a gender neutral name in Russian [Nikita Khruschev who succeeded Stalin as Chairman of USSR], but in Indians and also other subcontinentals to lesser extent its an exclusively female name if the context is Russian [Chechen rebels, RT, duma, oblast, ice-brekers etc etc] then the mask-pronoun-predictor can correctly look for further clues of whether its male or female but if its an Indian setting then it can predict a she using a large pretrained model obviates the use of lexicons in either case.

One objection that can be made is of memory and time complexity w.r.t lexicons, but I think lexicons esp name lexicons can also get heavy not just memory wise, but also time wise in terms of lookup time.

Esp with pervasisve Gpu availability , I think we must seriously think whether distillbert-base is after all better than a name lexicon

requirements.txt
```diff
@@ -13,7 +13,7 @@ pytest==6.2.4
 
 # for evaluation
 datasets==1.7.0
-transformers==4.6.1
+transformers==4.8.2
```
AbinayaM02, line 16: Hi @simonmeoni: Please revert the version of transformer, since it's in the main requirements.txt. You can specify this in the requirements.txt within your pronoun_to_noun_transformation folder.

kaustubhdhole:
Hi @simonmeoni, I am closing this PR since we are tracking all the PRs now. But feel free to reopen it later when you address reviewer comments and we are happy to merge.

PR #131, opened 2021-07-14
Word Noise Transformation

saqibns:
This transformation chooses a set of words at random from the context and the question and forms a sentence out of them. The sentence is then prepended or appended to the context.

It is inspired by the **AddAny** method described in [Adversarial SQUAD](https://arxiv.org/abs/1707.07328).

### DIFF
--- a/transformations/word_noise/transformation.py
+++ b/transformations/word_noise/transformation.py
@@ -1,24 +1,33 @@
-from typing import Tuple, List, Callable
+from random import Random
+from typing import Callable, List, Tuple
+
+import spacy
+
 from interfaces.QuestionAnswerOperation import QuestionAnswerOperation
 from tasks.TaskTypes import TaskType
-import spacy
-from random import Random
 
 
 class WordNoise(QuestionAnswerOperation):
     tasks = [TaskType.QUESTION_ANSWERING]
     languages = ["en"]
 
-    CONCAT_POSITION_APPEND = 'append'
-    CONCAT_POSITION_PREPEND = 'prepend'
-    CONCAT_POSITION_ANY = 'any'
+    CONCAT_POSITION_APPEND = "append"
+    CONCAT_POSITION_PREPEND = "prepend"
+    CONCAT_POSITION_ANY = "any"
 
-    def __init__(self, num_words: int, concat_position: str, remove_stop_words: bool = False, seed:int = 0, max_outputs: int =1):
+    def __init__(
+        self,
+        num_words: int,
+        concat_position: str,
+        remove_stop_words: bool = False,
+        seed: int = 0,
+        max_outputs: int = 1,
+    ):
         super().__init__(seed, max_outputs=max_outputs)
         self.n = num_words
         self.concat_pos = concat_position
         self.concat_function = self.get_concat_function()
-        self.spacy_model = spacy.load('en_core_web_sm')
+        self.spacy_model = spacy.load("en_core_web_sm")
         self.remove_stop_words = remove_stop_words
         self.random = Random(seed)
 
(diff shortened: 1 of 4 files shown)


transformations/word_noise/transformation.py
```diff
@@ -0,0 +1,82 @@
...
+        concat_position: str,
+        remove_stop_words: bool = False,
+        seed: int = 0,
+        max_outputs: int = 1,
+    ):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.n = num_words
+        self.concat_pos = concat_position
+        self.concat_function = self.get_concat_function()
+        self.spacy_model = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 30: `from initialize import spacy_nlp`
`self.spacy_model = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")`

transformations/word_noise/transformation.py
```diff
@@ -0,0 +1,82 @@
...
+        self, context: str, question: str, answers: [str]
+    ) -> List[Tuple[str, str, List[str]]]:
+        context_words = self.extract_words(context)
+        question_words = self.extract_words(question)
+        noise_words = self.random.choices(
+            context_words + question_words, k=self.n
+        )
+        noise = " ".join(noise_words)
+
+        context = self.concat_function(context, noise)
```
kaustubhdhole, line 80: Maybe you could add a third concat function "both" ?

kaustubhdhole:
Hi @saqibns this looks fine to me as it satisfies the review criteria and the code is well written. I've added a few comments.

On the same lines - it might be also be worth exploring appending sentences which make grammatical sense rather than just noise! Maybe replacing the answer entities & question verbs with different values and using them for replacing the answer sentence in the passage? I think the adversarial SQUAD paper do something like that. 

Thank you very much for your efforts!

transformations/word_noise/transformation.py
saqibns: I'll make this change.

transformations/word_noise/transformation.py
saqibns: Yep. That sounds like a good idea!

saqibns:
Hi @kaustubhdhole thank you for your review and comments! I will update the PR according to your comments.

Yes, you are right about Adversarial SQUAD generating sentences that modify the question and append the generated sentence to the context. I checked their code and the script is available [here](https://github.com/robinjia/adversarial-squad/blob/master/src/py/convert_questions.py). The script is rather long, so should I make a new pull request for the same? 

Also, they mention in their paper that the custom rules they've defined to convert a question into a statement are not exhaustive and sometimes they yield sentences that aren't grammatically correct. Not that it should be a problem, but just mentioning it here.

kaustubhdhole:
Okay, sounds good - a new PR would be better as there is another PR too in making here https://github.com/GEM-benchmark/NL-Augmenter/pull/165

saqibns:
Hi @kaustubhdhole I have added the option of both appending and prepending noise to the context. I have also changed the code for `spacy` initialization as per your request.

kaustubhdhole (approved):

transformations/word_noise/transformation.py
```diff
@@ -0,0 +1,101 @@
+from random import Random
```
kaustubhdhole, line 1: I think the test cases in main are failing because of this. Can you use np.random and create a PR to fix that please?

saqibns: Sure thing.

kaustubhdhole: I created a temporary fix here: https://github.com/GEM-benchmark/NL-Augmenter/pull/195/files

saqibns: Thank you!

PR #134, opened 2021-07-16
Add multilingual_code_switch method (IJCAI2020)

yizhen20133868:
Add transformation for CoSDA-ML: Multi-Lingual Code-Switching Data Augmentation for Zero-Shot Cross-Lingual NLP


transformations/multilingual_code_switch/transformation.py
```diff
@@ -0,0 +1,112 @@
...
+    return out.strip()
+
+
+class MultilingualCodeSwitch(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_TAGGING,
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.SENTIMENT_ANALYSIS,
+    ]
+    languages = ["en"]
```
samsontmr, line 87: Are your dictionaries many-to-many or only English-to-everything-else?

yizhen20133868: Hi, samson, is english2everything.

yizhen20133868: @samsontmr

samsontmr (commented):
Generally looks good, though I would rename the perturbation to `MultilingualDictionaryBasedCodeSwitch` since there are other methods to do this. Additionally, could you change the two for the two arxiv bibs to ACL Anthology ones?

transformations/multilingual_code_switch/test.json
```diff
@@ -0,0 +1,60 @@
...
+  "type": "multilingual_code_switch",
+  "test_cases": [
+    {
+      "class": "MultilingualCodeSwitch",
+      "inputs": {
+        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
+      },
+      "outputs": [
+        {
+          "sentence": "Andrew اخيرا returned the French réserver to Chris that I купила 最後 सप्ताह"
```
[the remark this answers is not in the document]
yizhen20133868: @samsontmr  Hi, samson. Thanks for your insightful suggestions. We have updated it according to your advice;
(1) We renamed our method and changed the bib;
(2) We have highlighted the random translation chosen in README.md. This is a very good question. 

We found that though we cannot guarantee that this is the correct word-to-word translation in the context, we can consider it as one of the data augmented strategies for cross-lingual zero-shot tasks. Actually, we also highlighted this in our paper. 

Thanks again!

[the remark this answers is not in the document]
Nickeilf: It seems that similar implementations of code-switching perturbation have existed in NL-Augmenter.

The [multilingual_lexicon_perturbation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/multilingual_lexicon_perturbation) uses a multilingual lexicon dictionary, and randomly replaces words into the corresponding translation in the target language according to the dictionary. This is very similar to this implementation.

The [mixed_language_perturbation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/mixed_language_perturbation) is a heavy version, which uses an off-the-shelf translation model to get the code-switching candidates.

Could you please demonstrate sufficient difference from these existing transformations (especially the first one), otherwise it might seem a bit overlapped.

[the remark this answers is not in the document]
yizhen20133868: @Nickeilf Hi, zhenhao. Thanks for your kind comments. This is a very good question. 

The main difference is that the mentioned transformation can only perform one-to-one mapping, which means that they only consider a pair of source and target languages at a time. By contrast, our transformation performs one-to-many mapping to construct multi-lingual code-switched data in multiple languages. 

By doing this, contextualized embeddings for all the languages can be aligned into the same space while prior work can only align representation in source and one target language for each language. Aligning multiple languages at one time is the core contribution in our work and we have highlighted it in our accepted paper[1].

Thanks for your kind suggestions again!

[1] Qin et al. CoSDA-ML: Multi-Lingual Code-Switching Data Augmentation for Zero-Shot Cross-Lingual NLP. IJCAI2020.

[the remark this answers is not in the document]
Nickeilf: Makes sense. Thanks for the explanation.

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -0,0 +1,112 @@
...
+            ]
+            out += " "
+        else:
+            out += word + " "
+    return out.strip()
+
+
+class MultilingualDictionaryBasedCodeSwitch(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_TAGGING,
```
Nickeilf, line 83: One question regarding code-switching in tagging tasks:
What is the POS tag for the code-switching word? I guess for different languages the linguistic rules might be different, so it is possible that the POS tag for the code-switching word is different from the original word.

If that's the case, I would suggest removing the `TaskType.TEXT_TAGGING` type because `SentenceOperation` assumes that the labels are not changed.

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -0,0 +1,112 @@
...
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.SENTIMENT_ANALYSIS,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_outputs=1, code_switch_rate=0.9):
```
Nickeilf, line 89: The `code_switch_rate` here seems a bit misleading. The default rate is 0.9, but it seems that the actual percentage of perturbed words in the test.json are far less than 90%. I think it is because you first sample from one of the target languages but some words might not exist in the vocab of the sampled language. Therefore, some words are skipped.

Do you have any idea?

transformations/multilingual_dictionary_based_code_switch/transformation.py
yizhen20133868: @Nickeilf hi, zhenhao. Thanks for your good suggestion. Though we proved the method is effective for sequence tagging tasks (e.g., slog filling), we have followed your suggestion to remove the TaskType.TEXT_TAGGING type because SentenceOperation assumes that the labels are not changed.

transformations/multilingual_dictionary_based_code_switch/transformation.py
yizhen20133868: @Nickeilf Yes. Your understanding is right.

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -0,0 +1,111 @@
...
+        super().__init__(seed, max_outputs=max_outputs)
+        self.seed = seed
+        self.code_switch_rate = code_switch_rate
+        self.switch_dict = load_dict()
+
+    def generate(self, sentence: str):
+
+        return [code_switch(sentence, self.switch_dict, self.code_switch_rate, self.seed)]
```
samsontmr, line 96: Just noticed this: could you refactor this such that `code_switch` operates at the word level? In other words, move the loop over the sentence into `generate` and call `code_switch` on each word.

transformations/multilingual_dictionary_based_code_switch/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## What tasks does it intend to benefit?
+This alignment of different language it brought can benefit a massive of multi-lingual tasks.
+As mentioned in the paper, we test this method on tasks including these multi-lingual tasks: 
+Natural Language Inference, Sentiment Classification, Document Classification, Dialogue State Tracking (DST), Spoken Language Understanding
+We perform t-test for all experiments to measure whether the results from the proposed model are significantly better than the baselines(mBERT and XLM).
+
+## What are the limitations of this transformation?
+It is worth noticing that words in the source language can have multiple translations in the target language. In this case, we randomly choose any of the multiple translations as the replacement target language word. Though we cannot guarantee that this is the correct word-to-word translation in the context, we can consider it as one of the data augmented strategy for our tasks. This is verified by our work (CoSDA-ML) [1].
```
samsontmr, line 19: We also mention this in our code-mixing paper for our PolyGloss attack and propose a method to disambiguate translations, but that requires a sentence translation.

samsontmr (commented):
Generally looks good, just some final comments.

transformations/multilingual_dictionary_based_code_switch/transformation.py
yizhen20133868: @samsontmr Samson, thanks for your suggestion. We have changed the code_switch operation from sentence-level to token-level operation:)

transformations/multilingual_dictionary_based_code_switch/README.md
yizhen20133868: @samsontmr Sure, we have added your good work on the section of What are the limitations of this transformation? Thanks again!

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -92,20 +88,26 @@ def __init__(self, seed=0, max_outputs=1, code_switch_rate=0.9):
         self.switch_dict = load_dict()
 
     def generate(self, sentence: str):
-
-        return [code_switch(sentence, self.switch_dict, self.code_switch_rate, self.seed)]
+        words = sentence.split(" ")
+        out = ""
+        for word in words:
+            out += code_switch(word, self.switch_dict, self.code_switch_rate, self.seed)
+        return [out]
```
samsontmr, line 95: You might want to support the `max_outputs` attribute by using another loop. I believe the `random.seed(seed)` call should happen at the top of this function then.

yizhen20133868: @samsontmr Yes, samson. We have fixed it.

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -0,0 +1,120 @@
...
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_outputs=1, code_switch_rate=0.9):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.seed = seed
+        self.code_switch_rate = code_switch_rate
+        self.switch_dict = load_dict()
+
+    def generate(self, sentence: str):
+        random.seed(self.seed)
```
Nickeilf, line 87: Minor suggestions: It's better to put this line into the __init__() function.

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -0,0 +1,120 @@
...
+        ouputs = []
+        for i in range(self.max_outputs):
```
Nickeilf, line 89: A potential issue: The `outputs` list might contain duplicate sentences because of probability. Could you fix this?

Maybe a while loop with a sanity check of duplicate sentences would be better?

Nickeilf (approved):

transformations/multilingual_dictionary_based_code_switch/transformation.py
```diff
@@ -88,11 +84,21 @@ def __init__(self, seed=0, max_outputs=1, code_switch_rate=0.9):
...
     def generate(self, sentence: str):
-        words = sentence.split(" ")
-        out = ""
-        for word in words:
-            out += code_switch(word, self.switch_dict, self.code_switch_rate, self.seed)
-        return [out]
+        random.seed(self.seed)
+        ouputs = []
+        i = 0
+        while i < self.max_outputs:
```
samsontmr, line 90: There's a danger the while loop will not terminate if the space of possible outputs is smaller than `max_outputs`. Since it is `max_outputs`, perhaps you can either use `num_tries` variable to track the number of tries and terminate once it reaches a certain value regardless of the actual outputs generated, or just use `range(max_outputs)`.

yizhen20133868: @samsontmr Got it.  fixed~

samsontmr (approved):

PR #137, opened 2021-07-16
replaced bleu score with rouge score

ashish3586:


kaustubhdhole:
Please resolve conflicts.

evaluation/evaluate_text_generation.py
```diff
@@ -1,14 +1,18 @@
...
 
+scorer = rouge_scorer.RougeScorer(['rouge1'])
+
 
-def sacrebleu_score(hypotheses, references):
-    return corpus_bleu(hypotheses, [references]).score
+def rogue_score(hypotheses, references):
+    total_f1_scores = 0.0
+    for ref, hyp in zip(references, hypotheses):
+        total_f1_scores+=scorer.score(ref, hyp)['rouge1'][2] # [2] for f1 score
```
sebastianGehrmann, line 14: Why ROUGE-1?

evaluation/evaluate_text_generation.py
```diff
@@ -1,14 +1,18 @@
...
+    return total_f1_scores/len(hypotheses)
```
sebastianGehrmann, line 15: The previous code supported multi-reference while this one does not.

sebastianGehrmann (changes requested):

evaluation/evaluate_text_generation.py
ashish3586: rouge1 is matching unigrams between hypothesis and reference. We can also use rouge2(bigram) along with rouge1 if needed.

evaluation/evaluate_text_generation.py
ashish3586: yes, earlier library function "score" from "sacrebleu" is taking a list of hypotheses and references as input, "score" function in "rouge" is taking one hypothesis, reference pair as input. Please let me know if somehow we can make it multi-reference?

evaluation/evaluate_text_generation.py
ashish3586: Hi Sebastian, I have also added "ROUGE-2" and "ROUGE-L" in the evaluation. Please have a look.

evaluation/evaluate_text_generation.py
sebastianGehrmann: Yes, the original ROUGE paper describes how to do this. I implemented this in BIG-bench like this: https://github.com/google/BIG-bench/blob/main/bigbench/api/task_metrics.py#L86

evaluation/evaluate_text_generation.py
sebastianGehrmann: Thanks

ashish3586:
> I don't think my initial comment was addressed that the current implementation does not support multi-reference which needs to be added. It probably would also make sense to write docstrings and add a test to ensure that multi-reference and single-reference both work.

Apologies for the very very ........... very late response.  I promise this will not happen again. I changed the rouge function for multi-reference. added docstring with function. added test cases for single reference and multi-reference (in the main function). Please have a look.

PR #140, opened 2021-07-18
Pinyin Conversion Transformation

cascadianblue:


transformations/pinyin/test.json
```diff
@@ -0,0 +1,41 @@
...
+        "sentence": "wo de 5 G shouji bingbu tai gui 。"
+      }]
+    },
+    {
+      "class": "PinyinTranscription",
+      "inputs": {
+        "sentence": "由于Albert Einstein是犹太裔人，所以尽管身为普鲁士科学院教授，他并没有返回德国。"
+      },
+      "outputs": [{
+        "sentence": "youyu lbert  Einstein shi youtaiyiren ， suoyi jinguan shenwei pulushi kexueyuan jiaoshou ， ta bingmeiyou fanhui deguo 。"
```
Nickeilf, line 37: The "Albert" becomes "lbert" in the output. Do you have any idea what's happening here?

Nickeilf (commented):
The transformation seems interesting and the implementation is generally good.

I have a minor suggestion:
Would it be possible to also generate Pinyin with tones? For example,
_"在国家战略方面"_ will be something like _"zai4 guo2jia1 zhan4lve4 fang1mian4"_.
If that is possible, I guess you could add an argument in the `__init__()` function like `with_tone=True` to determine whether to generate tones in the outputs.

.github/workflows/nl-action.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       run: |
         if [ -f requirements.txt ]; then pip3 install --upgrade --upgrade-strategy eager -r requirements.txt -e . ; fi
         pip3 install https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
+        pip3 install https://github.com/explosion/spacy-models/releases/download/zh_core_web_sm-3.0.0/zh_core_web_sm-3.0.0.tar.gz
```
Nickeilf, line 29: Maybe it's better to remove here and put it as a prerequisite in the README file, otherwise it might add burden for other PR.

transformations/pinyin/test.json
cascadianblue: I think the core character-to-pinyin model (g2pM) sometimes has issues with non-Chinese characters, so I fixed this bug by only sending Chinese characters to g2pM.

.github/workflows/nl-action.yml
cascadianblue: I'm not totally sure I understand what you are suggesting. What I am understanding is: you want me to remove the `zh_core_web_sm` model installation from the CI workflow, and write something about installing it manually in my transformations `README.md`. Won't that cause the PR workflow to fail, though?

Nickeilf: Maybe you could first check the CI results, if it all test cases are passed, then remove it from the workflow file.
Another way is to run `pytest -s --t=pinyin` on your local environment.

Also tagging @kaustubhdhole here.

cascadianblue:
@Nickeilf I uninstalled the `zh_core_web_sm` model as (I believe) you requested. The CI action failed as I anticipated, with the message:

`OSError: [E050] Can't find model 'zh_core_web_sm'. It doesn't seem to be a Python package or a valid path to a data directory.`

What should we do now?

tongshuangwu:
Hi! Could you please try to include the spacy lang package as part of the local requirements.txt? something like this: https://stackoverflow.com/questions/61702357/how-to-add-a-spacy-model-to-a-requirements-txt-file

transformations/pinyin/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Previous Work
+
+The character-to-pinyin converter at the core of this project is a neural 
+model described in this paper:
+
+```bibtex
+@article{DBLP:journals/corr/abs-2004-03136,
+  author    = {Kyubyong Park and
+               Seanie Lee},
+  title     = {g2pM: {A} Neural Grapheme-to-Phoneme Conversion Package for MandarinChinese
```
tongshuangwu, line 32: super minor, but add a space between Madarin/Chinese?

tongshuangwu (approved):
Thanks for the fix! I'm approving this on my end.

cascadianblue:
> Thanks for the fix! I'm approving this on my end.

Great! I guess we are just waiting on review from @Nickeilf now?

Nickeilf (approved):
Sorry for the delay. It looks good now :+1: . I've passed my review.

transformations/pinyin/requirements.txt
```diff
@@ -0,0 +1,2 @@
+g2pM==0.1.2.5
+zh-core-web-sm @ https://github.com/explosion/spacy-models/releases/download/zh_core_web_sm-3.0.0/zh_core_web_sm-3.0.0.tar.gz
```
kaustubhdhole, line 2: I think it would be heavy for this to be downloaded by default for everyone. So, I wouldn't suggest to be put like this. 
Temporarily, you can rename this requirements.txt to requirements-disable.txt.
And add heavy=True in the implementation. 
@Nickeilf @AbinayaM02 @aadesh11

cascadianblue: What does "temporarily" mean in this context? Do you want me to make this change, and then undo it before merging the PR?

kaustubhdhole: No. you can make this change and let it be merged. It just won't install and build by default. You can mention in the README  the procedure to additionally setup this transformation so that users would know that they would need to install this requirements.txt additionally.

PR #142, opened 2021-07-19
Sentence pair tasks interface

kaustubhdhole:
Add interface and a sample transformation for sentence pair tasks (like paraphrase/entailment)

### DIFF
--- a/tasks/TaskTypes.py
+++ b/tasks/TaskTypes.py
@@ -20,3 +20,6 @@ class TaskType(enum.Enum):
     SENTIMENT_ANALYSIS = (
         12,
     )  # This is a specific type of text classification with unique properties
+    PARAPHRASE_DETECTION = (13,)
+    TEXTUAL_ENTAILMENT = (14,)
+
--- a/tasks/TaskTypes.py
+++ b/tasks/TaskTypes.py
@@ -22,4 +22,5 @@ class TaskType(enum.Enum):
     )  # This is a specific type of text classification with unique properties
     PARAPHRASE_DETECTION = (13,)
     TEXTUAL_ENTAILMENT = (14,)
+    QUALITY_ESTIMATION = (15,)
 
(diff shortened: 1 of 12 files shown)


tasks/TaskTypes.py
```diff
@@ -20,3 +20,6 @@ class TaskType(enum.Enum):
     SENTIMENT_ANALYSIS = (
         12,
     )  # This is a specific type of text classification with unique properties
+    PARAPHRASE_DETECTION = (13,)
+    TEXTUAL_ENTAILMENT = (14,)
+
```
Nickeilf, line 25: I think quality estimation can be another task of SentencePairOperation, where the inputs are source and translation and the output is a continuous score.

transformations/lexical_counterfactual_generator/transformation.py
```diff
@@ -0,0 +1,111 @@
...
+        return [sentence]
+
+    def generate(self, sentence1: str, sentence2: str, target: str) -> List[Tuple[str, str, str]]:
+        """
+        This implementation augments data for semantic similarity tasks.
+        This converts pairs which are positive (i.e. paraphrases-> label:1) to non-paraphrases (label:0).
+        """
+        examples = []
+        if target == self.pos_label:
+            for sentence1_n in self.negate_if_single_verb(sentence1):
```
Nickeilf, line 70: It seems that the `negate_if_single_verb` returns a list of only one sentence, then there's no need to do a for loop here.

tasks/TaskTypes.py
kaustubhdhole: Excellent idea! I will add it!

transformations/lexical_counterfactual_generator/transformation.py
kaustubhdhole: The `bt.generate()` function can generate multiple ones actually.. (line 61)

PR #143, opened 2021-07-19
Added phonetic match filter

richplant:
Add a filter that searches for keyword matches using phonetic matching algorithms


requirements.txt
```diff
@@ -8,6 +8,9 @@ torch
 num2words==0.5.10
 word2number==1.1
 
+# for PhoneticMatchFilter
+jellyfish==0.8.2
```
kaustubhdhole, line 12: This should be in a separate requirements.txt in the folder `soundex`

RyanGoslingsBugle: Resolved in 26eed69dba6909b7a04b2d544a3b92834c7b74a2

tuetschek (commented):
Looks mostly good to me! The filter is reasonable and the code looks mostly OK, the required jellyfish library is BSD-licensed. 

I'd have a few suggestions though:
* In the readme, I'd somehow separate the example command from the example output
* Wouldn't this be applicable to other tasks as well (e.g. text tagging)? I don't see why not.
* I don't like the default keywords, IMO they will just cause confusion – I'd suggest you remove them.

tuetschek:
@RyanGoslingsBugle Thanks for the updates! Just a question – are you sure this should be applied to *all* task types?

RyanGoslingsBugle:
No prolem. I don't see which tasks it _wouldn't_ be applicable to - even in say, an RDF triple, you may wish to filter subject entities based on their soundalike quality, if you expect typos in your dataset for instance.

tuetschek (approved):
@RyanGoslingsBugle good point – all good IMO :-)

RyanGoslingsBugle:
@tuetschek cheers!

kaustubhdhole (approved):

PR #144, opened 2021-07-20
Added gender neutral rewrite transformation

tonysun9:


kaustubhdhole:
Pinging @ashish3586 @Saad-Mahamood

transformations/gender_neutral_rewrite/transformation.py
```diff
@@ -0,0 +1,425 @@
...
+    # compromise, allowing computation to proceed much faster while still giving the model a large context to make
+    # predictions at each step.
+
+    encodings = tokenizer(sentence, return_tensors='pt')
+
+    # can adjust stride based on size of input
+    # if (1) input is longer (e.g. document) or (2) we wish to have faster computation, we can set longer stride
+    # reference: https://huggingface.co/transformers/perplexity.html
+
+    if stride == 1:  # if stride is 1, just return average log prob of each token (no need to copy and mask)
```
Saad-Mahamood, line 151: One question. Given the use case of here is at a sentence level. Will there ever be a need to have a bigger `stride` value? If not is there any need for the code underneath this `if` statement?

Saad-Mahamood (changes requested):
Generally well written and documented code. I only have one improvement request and one query before I can approve.

kaustubhdhole:
@tonysun9 would you like to address the above comments ?

transformations/gender_neutral_rewrite/transformation.py
tonysun9: Good observation. Generally, setting `stride` to be 1 is appropriate. However, in certain situations, a larger value may be useful. For example, when compute power is an issue, setting `stride` to be 2, 3, or even 5 can help immensely. From my experience playing around with this parameter, having larger values can significantly improve runtime without noticeable decrease in the quality of rewrites.

transformations/gender_neutral_rewrite/myconstants.py
```diff
@@ -0,0 +1,66 @@
+# pronouns
```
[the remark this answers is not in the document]
tonysun9: Thanks for the suggestion. We certainly would encourage the addition of more occupational words and gender-neutral inflections. We chose to use a `.py` file for the ability to add comments where necessary, since it may not always be immediately obvious why a certain mapping is included. We hope that users can easily add gender or occupation-specific words by following the existing template (perhaps adding an instruction to the README may not be a bad idea as well).

ashish3586 (changes requested):

transformations/gender_neutral_rewrite/transformation.py
```diff
@@ -0,0 +1,425 @@
...
+SIMPLE_REPLACE = EASY_PRONOUNS
+SIMPLE_REPLACE.update(GENDERED_TERMS)
+
+# load SpaCy's "en_core_web_sm" model
+# English multi-task CNN trained on OntoNotes
+# Assigns context-specific token vectors, POS tags, dependency parse and named entities
+# https://spacy.io/models/en
+import en_core_web_sm
+
+nlp = en_core_web_sm.load()
```
kaustubhdhole, line 20: Can you load this using the global spacy load?
```python
from initialize import spacy_nlp
import spacy
```

kaustubhdhole:
@Saad-Mahamood @ashish3586

transformations/gender_neutral_rewrite/transformation.py
```diff
@@ -0,0 +1,425 @@
...
+
+class GenderNeutralRewrite(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_outputs=1):
```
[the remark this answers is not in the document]
tonysun9: Done. Thanks for the suggestion.

transformations/gender_neutral_rewrite/transformation.py
tonysun9: Done. Thanks for the suggestion.

tonysun9:
@kaustubhdhole @Saad-Mahamood I believe I've addressed the comments appropriately. Could you please take a look when you get a chance?

transformations/gender_neutral_rewrite/transformation.py
```diff
@@ -0,0 +1,482 @@
...
+
+
+class GenderNeutralRewrite(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["en"]
+
```
Saad-Mahamood, line 474: Missing the [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) variable here.

Saad-Mahamood (changes requested):
Missing keywords variable in transform.py needed before this can be merged.

transformations/gender_neutral_rewrite/transformation.py
tonysun9: Added.

transformations/gender_neutral_rewrite/README.md
```diff
@@ -0,0 +1,127 @@
+# Gender Neutral Translation
+This perturbation adds noise to all types of text sources (sentence, paragraph, etc.) proportional to noise erupting
```
ashish3586, line 2: The "Gender Neutral Translation" definition seems different from the implementation.

transformations/gender_neutral_rewrite/transformation.py
```diff
@@ -0,0 +1,493 @@
...
+        else:
+            token_texts.append(token.text)
+        if token.whitespace_:  # filter out empty strings
+            token_texts.append(token.whitespace_)
+
+    new_sentence = "".join(token_texts)
+    return new_sentence
+
+
+class GenderNeutralRewrite(SentenceOperation):
```
ashish3586, line 473: Please add docstring for the GenderNeutralRewrite class.

transformations/gender_neutral_rewrite/README.md
tonysun9: Thanks for the catch. I've removed these lines in the README.

transformations/gender_neutral_rewrite/transformation.py
tonysun9: Done. Thanks for the suggestion.

tonysun9:
> @kaustubhdhole @Saad-Mahamood I believe I've addressed the comments appropriately. Could you please take a look when you get a chance?

Would it be possible to get an extension on the 09/30 deadline? I believe I have addressed all the comments appropriately.

Saad-Mahamood (approved):

kaustubhdhole:
@tonysun9 please fix the test cases. Will merge then.

tonysun9:
> @tonysun9 please fix the test cases. Will merge then.

I think the failing test case is not related to my pull request?

kaustubhdhole:
Seems that is the same error as that in main. Okay, I will merge this then. One last thing, please move the spacy loading into your constructor rather than in the import file. Thanks for your hard work and thanks @Saad-Mahamood and @ashish3586 for reviewing the code.

tonysun9:
> Seems that is the same error as that in main. Okay, I will merge this then. One last thing, please move the spacy loading into your constructor rather than in the import file. Thanks for your hard work and thanks @Saad-Mahamood and @ashish3586 for reviewing the code.

Done. Thanks for the suggestion and taking the time to review the code.

Issue #145, opened 2021-07-20
Should we add a global seed for all transformations?


SirRob1997:
Almost all transformations such as, for example, `butter_fingers_perturbation` or `replace_numerical_values` use a `seed` in their constructor. How are we going to handle the global seed? we could easily set one in `initialize.py` that get's imported in each transformation and set as the default, similar to what is currently done for `spacy_nlp`. Otherwise, we can also set it during evaluation, as far as I could tell that is not currently done but I think having a consistent default is a little cleaner.

Happy to make the required changes if that's something we'd want.

kaustubhdhole:
It would be great to have such a facility. I think it's best it is set in `initilialize.py`. Feel free to create a PR! :) 
CC: @aadesh11

PR #146, opened 2021-07-20
Global seed changes

SirRob1997:
This PR is for the changes discussed in #145, I've also used this to fix some remaining `space.load` in some of the filters/transformations so they are loaded only once.

Would be nice if you could test all things to make sure I didn't break anything. 

Draft for now, since I think some tests need to be adjusted now since they used a different seed.

### DIFF
--- a/transformations/change_char_case/transformation.py
+++ b/transformations/change_char_case/transformation.py
@@ -3,9 +3,7 @@ import random
 from interfaces.SentenceOperation import SentenceOperation
 from tasks.TaskTypes import TaskType
 
-
-def change_char_case(text, prob=0.1, seed=0, max_outputs=1):
-    random.seed(seed)
+def change_char_case(text, prob=0.1, max_outputs=1):
     results = []
     for _ in range(max_outputs):
         result = []
@@ -34,11 +32,11 @@ class ChangeCharCase(SentenceOperation):
     ]
     languages = ["en"]
 
-    def __init__(self, seed=0, max_outputs=1):
-        super().__init__(seed, max_outputs=max_outputs)
+    def __init__(self, max_outputs=1):
+        super().__init__(max_outputs=max_outputs)
 
     def generate(self, sentence: str):
         perturbed = change_char_case(
-            text=sentence, prob=0.1, seed=self.seed, max_outputs=self.max_outputs
+            text=sentence, prob=0.1, max_outputs=self.max_outputs
         )
         return perturbed
--- a/filters/quantitative_ques/filter.py
+++ b/filters/quantitative_ques/filter.py
@@ -1,34 +1,50 @@
 import spacy
+
+from initialize import spacy_nlp
 from interfaces.QuestionAnswerOperation import QuestionAnswerOperation
 from tasks.TaskTypes import TaskType
 
 
 class QuantitativeQuestion(QuestionAnswerOperation):
-    '''
+    """
     This filter used to identify the quantitative questions in a dataset
     It identifies the commonly used quant questions based on lexical matching
     (The set of phrases are stored in the self.quant_ques)
-    '''
+    """
+
     tasks = [TaskType.QUESTION_ANSWERING, TaskType.QUESTION_GENERATION]
     languages = ["en"]
 
     def __init__(self):
         super().__init__()
-        self.nlp = spacy.load("en_core_web_sm")
+        self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
         # Covers the broad types of quant questions: distance , age , measurable , un-measurable
-        self.quant_ques = ['many','much',
-                           'close','far',
-                           'young','old',
-                           'short','tall',
-                           'heavy','light',
-                           'narrow','wide',
-                           'deep','shallow',
-                           'broad','thin',
-                           'near','long']
+        self.quant_ques = [
+            "many",
+            "much",
+            "close",
+            "far",
+            "young",
+            "old",
+            "short",
+            "tall",
+            "heavy",
+            "light",
+            "narrow",
+            "wide",
+            "deep",
+            "shallow",
+            "broad",
+            "thin",
+            "near",
+            "long",
+        ]
 
-
-    def filter(self,context:str = None,question: str = None,answers:str = None) -> bool:
+    def filter(
+        self, context: str = None, question: str = None, answers: str = None
+    ) -> bool:
         tokenized = self.nlp(question, disable=["parser", "tagger", "ner"])
-        return ('how' == tokenized[0].text.lower() and tokenized[1].text.lower() in self.quant_ques)
-
-
+        return (
+            "how" == tokenized[0].text.lower()
+            and tokenized[1].text.lower() in self.quant_ques
+        )
--- a/filters/token-amount/filter.py
+++ b/filters/token-amount/filter.py
@@ -1,6 +1,8 @@
 import operator
 import spacy
 
+from initialize import spacy_nlp
+
 from interfaces.SentenceOperation import SentenceOperation
 from tasks.TaskTypes import TaskType
 from collections import defaultdict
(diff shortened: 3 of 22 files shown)


SirRob1997:
@kaustubhdhole can you take a look at this? `random` and `numpy` should now have a global seed of `0` and these should be shared by default across imports. I'm not exactly sure why the `butter_fingers_perturbation` results in a different output, since it was using `0` as a seed before as well. Nevertheless, some tests were using seed `42` and these would most definitely need to be adjusted which I don't want to do without explicit approval.

kaustubhdhole:
Hi @SirRob1997, this looks very promising! Yes, it is okay to change the values of the test cases to reflect the changes off a new seed! I would be happy to re-test it myself too if you need a final confirmation. :)

aadesh11:
Thanks, @SirRob1997 for making the codebase looks cleaner. Please help to make the required changes in the test case of Butter Finger perturbation to pass the GitHub action checks.

SirRob1997:
Yup will adapt things either on Sunday or Monday, depending on when I can scrap a few minutes.

SirRob1997:
I'm having trouble getting three of the transformations to run, so it's hard for me to adjust some of the tests, all others pass now. Does someone know what I need to do to resolve these issues:

`gender_culture_diverse_name` / `gender_culture_diverse_name_two_way`: 

> UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 921887: character maps to <undefined>

`sentence_reordering`:

> allennlp.common.checks.ConfigurationError: coref not in acceptable choices for dataset_reader.type: ['babi', 'conll2003', 'interleaving', 'multitask', 'multitask_shim', 'sequence_tagging', 'sharded', 'text_classification_json']. You should either use the --include-package flag to make sure the correct module is loaded, or use a fully qualified class name in your config file like {"model": "my_module.models.MyModel"} to have it imported automatically.

SirRob1997:
Not exactly sure why the `pytest` GitHub action test results in a different output and hence fails. I get reproducible results on my local machine with the new changes and, for example, for the first failing one `ChangeCharCase` it passes the tests.

SirRob1997:
The `pytest` Github action is failing because it evaluates against the already existing tests and not my changes to them. I don't think there's anything I can do on my end to fix that @kaustubhdhole @aadesh11. My changes result in reproducible results on local and pass all tests.

kaustubhdhole:
For Review @aadesh11 @AbinayaM02

transformations/change_char_case/transformation.py
```diff
@@ -3,9 +3,7 @@
 from interfaces.SentenceOperation import SentenceOperation
 from tasks.TaskTypes import TaskType
 
-
-def change_char_case(text, prob=0.1, seed=0, max_outputs=1):
-    random.seed(seed)
+def change_char_case(text, prob=0.1, max_outputs=1):
```
AbinayaM02, line 6: Hi @SirRob1997 : I went through the build log. The pytest is failing for only ChangeCharCase transformation. The reason is that this transformation uses random function in the logic and a seed was used in the earlier version of the code. 

In the current logic, seed is global and this random function is not seeded. If you fix this, probably the build will go through successfully (or we may encounter more bugs).

AbinayaM02 (commented):
Fix for pytest failure.

transformations/change_char_case/transformation.py
SirRob1997: Hi @AbinayaM02, thanks for taking a look at this! This was a common pattern among many transformations, hence this PR. The random function should in fact be seeded, afaik the global seed gets set for the tests [here](https://github.com/GEM-benchmark/NL-Augmenter/blob/b221b970f946a376d6db867e1c7e143376c1b994/test/test_main.py#L106) and seeding is shared among multiple `import random` in different files. Running the tests also creates reproducible results on local which verifies that `random` is in fact already seeded properly. 

I believe there is a problem with the GitHub action that evaluates the new code against the already existing tests instead of the slightly adapted ones provided in this PR or am I missing something here?

AbinayaM02: This is from the Github Actions log,

AssertionError: Mis-match in expected and predicted output for ChangeCharCase transformation: 
E              Expected Output: Andrew finally returned the French book tO ChriS that I bought Last week. 
E              Predicted Output: AndreW finally returned The French booK To ChRis that I bougHt last week.

I checked the test.json in the PR for this ChangeCharCase transformation. The expected output in the test.json is picked correctly (first line) but the actual execution of the transformation during pytest is producing a different output (second line). So, I thought this has to do with the random.random() call in the code.

SirRob1997: Hmm, yeah you're right that is weird. Any idea why this is happening in the Github action and not on local?

AbinayaM02: I'm not completely sure about why this is happening. Will dig a bit deeper.

There is a strange pattern in all the test.json files. Only 2 or 3 out of 5 sample outputs are changed by the global seed changes. I was expecting all the sentence outputs to have some changes since the new global seed might be different from what was used earlier in the individual modules. Either there should be no changes at all in the test.json files [if the seeds before and after the current PR are same] else all the output samples should have change. 

Is my understanding right?

SirRob1997: Yeah that's right, I noticed that too and opened up issue #162 but I brushed it off as random chance. I haven't found a reason for why that pattern would occur. If I recall correctly, for some of the transformations there were changes in the `test.json` needed even though the `GLOBAL_SEED` matched the previously used one which I found rather odd. However, I couldn't find a root cause for it. I was thinking that I missed a place where seeding was additionally done but a project wide search for "seed" seemed to reject that hypothesis.

AbinayaM02: Let me try to reproduce your fix from my end and see if it happened by chance or is it some bug. Will update soon!

AbinayaM02: Hey @SirRob1997 : I'm getting the exact error on the Github actions log on my local as well. Could you please execute the following command and let me know if you also get the same error or it passes without any issue.

```python
pytest -v --t=light
```

SirRob1997: Yes, that indeed fails for more as well. However, this still passes all the tests for this transformation:
```
pytest -s --t=change_char_case
```

AbinayaM02: Yes, it failed for me too. 

I think I found the issue after breaking my head for so long! The issue is with the random.random() call in the change_char_case transformation.

 I did the following to narrow down the issue,
1. Created new test.json file for the change_char_case transformation after leveraging the global seed [by calling the initialize_model() method only for test file creation].
2. Ran the pytest for change_char_case transformation and all light transformations. In both the cases, it failed for change_char_case transformation.
3. Tried printing the random.random() call. My assumption was that it should use the global seed and give me the same random numbers both during pytest check and while running transforamtion.py separately but it didn't.
4. Finally, I realized that only test case generation is leveraging the global seed but the actual transformation is not using it. So, called set_seed() function in the  change_char_case method. 

```python
from initialize import set_seed

def change_char_case(text, prob=0.1, max_outputs=1):
    set_seed()
    results = []
    for _ in range(max_outputs):
        result = []
        for c in text:
            x = random.random()
            y = random.random()
            print(x, y)
            if c.isupper() and x < prob:
                result.append(c.lower())
            elif c.islower() and y < prob:
                result.append(c.upper())
            else:
                result.append(c)
        result = "".join(result)
        results.append(result)
    return results
```

Once, I made the above change, the testcases for change_char_case passed (both individually and as part of the light transformations). And I landed up in another test case failure for ChangePersonNamedEntities transformation!

So, to keep the global seed changes everywhere, I think you need to check which all transformations use seed, and call set_seed() before performing transformation. Also the test.json files should be recreated for all of them.

SirRob1997: Yeah, good catch for sure. It's still weird to me that `set_seed()` in the `test_main.py` doesn't handle this properly since I believe it should. I wouldn't want to add  `set_seed()` to every transformation since that seems very unclean... Would be nice to figure out why exactly this isn't properly handled from the `test_main.py`.

AbinayaM02 (changes requested):
Please resolve the conflicts and let's see if the pytest goes through without any issue.

PR #154, opened 2021-07-23
Grapheme to phoneme transformation

gokyori:


transformations/grapheme_to_phoneme_transformation/transformation.py
```diff
@@ -0,0 +1,78 @@
...
+def grapheme_phoneme(grapheme):
+    '''
+    converts each word to phonems. If there are multiple pronounciation of a
+    word, only the first prononciation is taken. Stress information from each
+    word is removed.
+    '''
+    phoenems = pronouncing.phones_for_word(grapheme)
+
+    if len(phoenems)> 0:
+        phoenem_without_stress = ''.join([x for x in phoenems[0] if x.isalpha()]).lower()
```
kaustubhdhole, line 21: Thanks for your changes! One comment here - why not use the remaining pronunciations to create multiple paraphrases?

transformations/grapheme_to_phoneme_transformation/transformation.py
```diff
@@ -0,0 +1,78 @@
...
+class PhonemeSubstitution(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=0, prob=0.5, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.spacy_pipeline = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 67: Also, please use the spacy from initialize

transformations/grapheme_to_phoneme_transformation/transformation.py
gokyori: Hi, Thank you for the comments. The reason I considered only one pronunciation is that the differences in pronunciations were mostly because of stress applied in different syllables(phones) rather than syllables themselves in most of the cases.

transformations/grapheme_to_phoneme_transformation/transformation.py
```diff
@@ -0,0 +1,84 @@
...
+
+    return [transformed_sentence]
+
+
+class PhonemeSubstitution(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
```
kaustubhdhole, line 67: Please add the relevant keywords here!

kaustubhdhole:
Here are my overall comments:
This is a novel transformation to NL-Augmenter as far as I can see and can help create noisy training data - especially since real life training data can be itself speech-noisy. I like the idea of cleverly using the pronouncing dictionary too. 
I would suggest you to do the robustness evaluation too to get an estimate of the robustness of your transformation.
If the test cases fail, feel free to remove the test cases for your PR [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/diverse_paraphrase/test.json) and maybe add them in your README.

It looks good to me overall!

msobrevillac:
I agree with @kaustubhdhole. The transformation is novel and so interesting and it is great for me as well.

 I just suggest to add keywords as it was mentioned before.

AbinayaM02:
Hi @gokyori: Please fetch the latest code (fetch & merge main and pull main into your branch). Pytest testcases should pass since many fixes have gone in after your PR. Thanks!

vgtomahawk (approved):
I think this perturbation looks in good shape to me.

The examples seem novel and interesting!

Moreover, any points of contention that could have naturally arisen have been covered in scope of the other reviews...

There are a couple of typos I'd mentioned which it would be nice if they're just fixed , but no big deal..

kaustubhdhole (approved):

PR #160, opened 2021-07-25
Added Synonym insertion

vukosim:


kaustubhdhole:
This is already implemented and is on the verge of merging: #51

vukosim:
@JosephSefara please see

JosephSefara:
@kaustubhdhole This implementation is different from #51 . 

- #51 **replace** a word with its synonym. [if the wrong synonym is used then original context may be lost]
- ours #160 **insert** a synonym next to its original word. This method retains the semantic meaning (no words are being replaced/removed) and somehow increases the context

kaustubhdhole:
Okay, I just have one comment: Do you think it might be better to write synonynms besides the original words so that the sentence is still well-formed?

transformations/synonym_insertion/transformation.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+
+
+class SynonymInsertion(SentenceOperation, ABC):
+    """
+    This class is an implementation of synonym insertion in the sentence. Created by the Authors of TextAugment
+    https://github.com/dsfsi/textaugment
+    """
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
```
kaustubhdhole, line 101: Also, please add keywords here.

kaustubhdhole: Also, I would also recommend adding the robustness evaluation for your PR that can be added to the leaderboard.

kaustubhdhole:
@vukosim @JosephSefara ping!

kaustubhdhole:
Also, one more thing: do you think it would be better to include the synonym within a bracket?

JosephSefara:
> Okay, I just have one comment: Do you think it might be better to write synonynms besides the original words so that the sentence is still well-formed?

@kaustubhdhole, I don't understand you question but we are inserting a synonym next to its original word. Augmentation sometimes is about adding noise to the sentence hence the sentence might not be well-formed but still retains original **context**. E.g. stopwords removal #268 removes stop words, thus the sentence might not be well-formed.

JosephSefara:
> Also, one more thing: do you think it would be better to include the synonym within a bracket?

@kaustubhdhole 
That can be done but not a good idea since most people clean their text before augmentation. 
- Text cleaning includes removal of special characters including **brackets**. 

Why remove special characters? 
- They do not carry any meaning and they create noise depending on the task being done.

transformations/synonym_insertion/transformation.py
JosephSefara: - keywords added
- readme and test.json contains the results of the Robustness Evaluation for 
  - Text Classification
  - Text Generation

PR #163, opened 2021-07-25
add: Grapheme to Phoneme transformation with g2pE

mrshu:
* Extend the transformation in #154 with the `g2pE` package, which
  should cover many more cases than the CMU pronounciation dictionary.

Signed-off-by: mr.Shu <mr@shu.io>


kaustubhdhole:
Hi @mrshu, thanks for your changes. Can you please add what benefit g2pE might provide (as compared to the CMU pronouncing dictionary). Also, it might be easier to add a separate parameter to select between both via the constructor.

kaustubhdhole:
Okay, I am closing this PR temporarily since we are tracking the PRs. feel free to reopen it later with the new changes. :)

mrshu:
Hey @kaustubhdhole,

Sorry, I should have responded way earlier in here. Here is the actual important part from the README that should answer your question directly:

> Note that the `g2pE` package [3] is used for the actual grapheme-to-phoneme
conversion, which combines the CMU Pronouncing dictionary (which is what the
`pronouncing` package does) with a model based on neural network, making it
possible to predict the phonemes even for words that are not part of the
aforementioned dictionary, such as for instance `"disinherited"`.

In principle, the CMU Pronouncing Dictionary is only updated a couple of times per decade (last update was in 2014) and contains about 134k  entries, which means that for pretty much anything non-trivial, you will probably run into out-of-vocabulary issues. The `g2pE` package builds upon the CMU Pronouncing Dictionary and for OOVs it predicts the pronunciation using a pre-trained neural net model.

Hence, I believe it would make sense to include something like this. Personally, I believe using `g2pE` should be a net benefit over doing plain `cmudict` but I'd be happy to add a separate parameter to the constructor that would allow for choosing one or the other :)

kaustubhdhole:
Okay @mrshu thank you very much for the clarification. And yes it would be better to add this as a separate parameter and since g2pE > cmudict, you can probably keep g2pE as the default setting.

PR #164, opened 2021-07-25
Add Emojify transformation

xiaohk:
# Emojify 🔤 → 🦄

This transformation augments the input sentence by swapping words into emojis with similar meanings. For example, it changes word `"movie"` to emoji `"🎬"`.

Author name: Zijie J. Wang

Author email: jayw@gatech.edu

Author Affiliation: Georgia Tech

## What type of a transformation is this?

This transformation acts like a translation to test language models' robustness and generalizability. In this context, we are translating English words into Emoji unicode. The transformed sentence has similar structure and semantics to the source sentence.

Some examples:

```
"Apple is looking at buying U.K. startup for $132 billion."

⬇

"🍎 is 👀 at 🛍️ 🇬🇧 startup for $1️⃣3️⃣2️⃣ billion."
```

```
"The quick brown fox jumps over the lazy dog."

⬇

"The quick 🟤 🦊 jumps over the lazy 🐕."
```

```
"Oh, and their spring rolls and the accompanying peanuts and hot sauces were also delicious."

⬇

"Oh, and their 🌱 🧻 and the accompanying 🥜 and 🌡️ sauces were also 😋."
```


tongshuangwu (commented):
Thanks for the neat PR! I did spot another later PR that similarly tries to insert emojis. I also mentioned in the other one that you two seem to create emoji mappers in different ways, so might be good to merge them? https://github.com/GEM-benchmark/NL-Augmenter/pull/186 @kaustubhdhole thoughts?

xiaohk:
@tongshuangwu Thanks for the comment! It seems #186 only focuses on smileys and smiley emojis with around 60 dictionary keys. My PR is more general and has more than 4k English keywords. I would like to keep two PRs separate.

tongshuangwu (approved):
thanks for the reply, I'm now approving this then!

transformations/emojify/transformation.py
```diff
@@ -0,0 +1,96 @@
...
+        super().__init__(seed, max_outputs=max_outputs)
+
+        # Load the emoji dictionary
+        dict_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), "emoji_dict.json"
+        )
+        self.word_to_emoji = load(open(dict_path, "r"))
+
+        # Load the spacy nlp
+        self.nlp = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 81: Okay one last thing: you might want to use spacy like this to avoid duplicate loading of spacy:
`from initialize import spacy_nlp`
`self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
`

kaustubhdhole (approved):

transformations/emojify/transformation.py
xiaohk: Sounds good! I have just created another PR (https://github.com/GEM-benchmark/NL-Augmenter/pull/208) to address it. Thank you!

xiaohk:
> thanks for the reply, I'm now approving this then!

Thank you! ٩( ᐛ )و

transformations/emojify/transformation.py
xiaohk: I encountered some issues when making the change :(

If you have time, it would be super helpful if you can take a look at #209. Thanks!

PR #165, opened 2021-07-25
Sentence Addition Transformation

timothy22000:
**Notes**

Added a sentence additions transformation to create perturbed outputs based on ideas and results from this paper where adding sentences to the original text to test the robustness of models.

The method that I employed to add sentences is by adding generated sentences from the GPT-2 model to create the perturbations so that can be applied to any settings other than the Q&A setting in original paper. The user can pass in a sentence or paragraph as input text.

More details in README.md

**Changes**

1. Added SentenceAdditionsTransformation
2. Added test cases for transformation

### DIFF
--- a/transformations/sentence_additions/README.md
+++ b/transformations/sentence_additions/README.md
@@ -1,4 +1,4 @@
-# Sentence additions
+# Sentence Additions 🦎  + ⌨️ → 🐍
 This perturbation adds generated sentence to all types of text sources (sentence, paragraph, etc.) by passing the input text to a GPT-2 Text Generation model.
 
 Author name: Timothy Sum Hon Mun (timothy22000@gmail.com)
@@ -7,12 +7,7 @@ Author name: Timothy Sum Hon Mun (timothy22000@gmail.com)
 This transformation will take in an input text and generate additional sentences to perturbe the inputs that have been shown by Jia and Liang, 2017 to impact model performance. This can be used to test model robustness. 
 
 ## What tasks does it intend to benefit?
-This perturbation would benefit all tasks on text classification and generation. 
-
-Benchmark results:
-
-- Sentiment analysis: we run sentiment analysis on a 1% sample of the IMDB dataset. The original accuracy is 956 and the perturbed accuracy is 95.2.
-- Text summarization: we run text summarization on a 1% sample of the xsum dataset. The original BLEU is 15.99 and the perturbed BLEU is 9.75.
+This perturbation would benefit all tasks on text classification and generation.
 
 ## Related work
 
--- a/transformations/sentence_additions/README.md
+++ b/transformations/sentence_additions/README.md
@@ -13,7 +13,6 @@ This transformation uses GPT-2 to generate new grammatical sentence that will be
 Run code below to evaluate:
 ```python evaluate.py -t ButterFingersPerturbation -task TEXT_CLASSIFICATION -m "textattack/roberta-base-SST-2" -d "sst2"```
 
-
 Accuracy of a RoBERTa model tuned on SST-2 (model: "textattack/roberta-base-SST-2") on a subset of the SST-2 dataset = 95.74
 
 Accuracy of the same model on the perturbed set = 89.0
--- a/transformations/sentence_additions/README.md
+++ b/transformations/sentence_additions/README.md
@@ -1,10 +1,10 @@
 # Sentence additions
 This perturbation adds generated sentence to all types of text sources (sentence, paragraph, etc.) by passing the input text to a GPT-2 Text Generation model.
 
-Author name: Timothy Sum Hon Mun (timothy2200@gmail.com)
+Author name: Timothy Sum Hon Mun (timothy22000@gmail.com)
 
 ## What type of a transformation is this?
-This transformation will take in an input text and generate additional sentences to perturbe the inputs that have been show by Jia and Liang, 2017 to be impact model performance. This can be used to test model robustness. 
+This transformation will take in an input text and generate additional sentences to perturbe the inputs that have been shown by Jia and Liang, 2017 to impact model performance. This can be used to test model robustness. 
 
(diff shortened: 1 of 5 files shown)


transformations/sentence_additions/requirements.txt
```diff
@@ -0,0 +1 @@
+transformers==4.7.0
```
kaustubhdhole, line 1: The main requirements.txt already contains a transformations version.

transformations/sentence_additions/README.md
```diff
@@ -0,0 +1,41 @@
...
+Author name: Timothy Sum Hon Mun (timothy22000@gmail.com)
+
+## What type of a transformation is this?
+This transformation will take in an input text and generate additional sentences to perturbe the inputs that have been shown by Jia and Liang, 2017 to impact model performance. This can be used to test model robustness. 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks on text classification and generation.
+
+## Related work
+
```
kaustubhdhole, line 13: Please also take a look at this PR: https://github.com/GEM-benchmark/NL-Augmenter/pull/131/files

transformations/sentence_additions/requirements.txt
timothy22000: I'll remove this. Thanks :)

transformations/sentence_additions/README.md
timothy22000: Hi Kaustubh,

Great to hear from you and I hope you are doing well!

I have taken a look at that PR and can see that it also has been inspired by the same paper but for the AddAny method whereas I was inspired by the AddSent and AddOneSent method in the paper. 

The AddAny method adds a sequence of n arbitrary words to perturbed the input text (eg "offered consistent operational e.g. observations precise conceptual intuitive." from their test case). On the other hand, the AddSent method adds proper grammatical sentences to the input text instead (eg "The man travelled to the past to change history."). 

The AddSent/AddOneSent has also been shown to be able to fool different models with the same generated adversarial example much better in comparison to the AddAny method so the perturbatation has better transferability across models.

Both the PR and the methods in the paper are applied to the Q&A setting. In particular, the AddSent method requires mutation of the original question asked and generating a fake answer based on that. Then, converting them into a statement that serves as the answer as an adversarial example.

I wanted a transformation that can introduce grammatical sentences without needing a question to be mutated and a fake answer to be generated so it can be applied to any text. Therefore, I propose this transformation that is independent of any question asked.

transformations/sentence_additions/transformation.py
```diff
@@ -0,0 +1,39 @@
...
+        self.max_length = max_length
+
+    def generate(self, sentence: str):
+        perturbed = self.sentence_additions(text=sentence)
+        return perturbed
+
+    def sentence_additions(self, text):
+        set_seed(self.seed)
+        generator = pipeline('text-generation', model=self.model)
+        outputs = generator(text, max_length = self.max_length, num_return_sequences = self.max_outputs)
```
kaustubhdhole, line 32: Okay, a couple of things here: Instead of passing direct text, do you think using prompt templates can force the model to generate paraphrases? (like "Paraphrase: ")

timothy22000: Hi @kaustubhdhole,

Sorry I missed your message. That's a very interesting idea! Are you thinking about using prompt templates as inputs to generate paraphrases of the original text? Using a manual or automated approach to create prompt templates? 

I am new to prompt templates so I'll need to look into this further. I came across this article that covers some aspects of prompt engineering:

http://pretrain.nlpedia.ai/data/pdf/template.pdf

Feel free to share any resources that you feel may be useful for me to know as well. :)

kaustubhdhole: No problem! There are some which are interesting which have been studied in the context of GPT-3: https://arxiv.org/pdf/2102.09690.pdf and https://aclanthology.org/2020.emnlp-main.346.pdf

My question was if instead of simply passing the original text, you could test out by appending words like "Sentence: Paraphrase: ", etc. This could be effective later when people decide to replace the model from GPT-2 to say, GPT-3.

transformations/sentence_additions/transformation.py
```diff
@@ -0,0 +1,39 @@
...
+Adds generated sentence into provided sentences or paragraph to create adversarial examples.
+"""
+
+
+class SentenceAdditions(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
```
kaustubhdhole, line 15: Please add [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) too.

transformations/sentence_additions/transformation.py
kaustubhdhole: No problem! You can check these two: https://arxiv.org/pdf/2102.09690.pdf, https://aclanthology.org/2020.emnlp-main.346.pdf. What I meant was that you could append your input with a piece of text (prompt) like "PARAPHRASE: " and try checking how it performs on that. Since few-shot prompting has shown impressive results in GPT-3, later users could pass GPT-3 or another model proven well on prompting. You could probably have a default prompt text defined in the constructor itself.

kaustubhdhole:
This PR looks promising overall. I've added some comments and happy to approve it once resolved.

transformations/sentence_additions/transformation.py
timothy22000: Hi @kaustubhdhole ,

Thanks for sharing those resources! I really appreciate it :D

I'll add a default prompt text ("PARAPHASE: " as per your suggestion) in the constructor which the user can change if they want and also an option to turn the prompting on/off.

kaustubhdhole (approved):

PR #167, opened 2021-07-25
Chinese Pinyin Butter Fingers Perturbation (Character and Word-Level Perturbation)

timothy22000:
**Notes**
This transformation uses a database of 16142 Chinese characters to create perturbation to the input text based on Chinese characters that have similar Pinyin. A user is likely to make a typo when provided with multiple Chinese characters with similar Pinyin as choices commonly seen in Pinyin keyboards for Windows and Mac OS.

Examples of Pinyin keyboard:
Screenshot 2021-07-26 at 19 41 28
Screenshot 2021-07-26 at 19 41 38

More details in the README.md

**Changes 25/07/2021**

1. Added chinese_char.json for 字库 
2. Added ChineseButterFingersPerturbation
3. Added test cases for transformation.

**Changes Update 31/08/2021**

1. Added common_chinese_char.json
2. Added for 3500常用汉字.txt, 四十万汉语大词库.txt and thuocl_all.txt for Chinese words 词库
3. Added Chinese word segmentation
4. Added word-level perturbation
5. Improved character level perturbation to prioritize showing more common characters
6. Added capability to consider Pinyin with or without tones
7. Added capability to switch between word + char level perturbation or only char level perturbation.
8. Improved ChineseButterFingersPerturbation 
9. Modified evaluate_text_classification.py to be able to evaluate Chinese texts against CLUE Roberta and other models from HuggingFace
10. Updated test cases for the latest transformation.

An example:
Screenshot 2021-08-31 at 21 37 57]

**Changes Update 17/09/2021**

1. Added keywords
2. Added Robustness Evaluation in README.md
3. Addressed comments


transformations/chinese_butter_fingers_perturbation/test.json
```diff
@@ -0,0 +1,50 @@
+{
+  "type": "chinese_butter_fingers_perturbation",
+  "test_cases": [
+    {
+      "class": "ChineseButterFingersPerturbation",
+      "inputs": {
+        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
+      },
```
kaustubhdhole, line 8: I am assuming this is still work in progress?

timothy22000: Hi Kaustubh,

Yes, I have an initial version working with a few Chinese characters with a similar approach to the original butter fingers pertubration but that doesn't work well with Chinese characters since there are over 20000 characters compared to 26 alphabets in English.

Therefore, I still working on a version where I can do the perturbation based on a database of Chinese characters. I'll push it as soon as I am done. :)

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,80 @@
...
+Chinese Characters Butter Fingers Perturbation
+"""
+
+class ChineseButterFingersPerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["en"]
```
Nickeilf, line 66: I think the language should be 'zh'.

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,80 @@
...
+        for char_dict in chinese_character_database:
+            if(u.unidecode(pinyin_for_char_to_be_perturbed) == u.unidecode(char_dict['pinyin'])):
+                chars_with_similar_pinyin += char_dict['word']
+    else:
+        print("Keyboard not supported.")
+
+    return chars_with_similar_pinyin
+
+
+def load_chinese_character_data_into_memory():
```
Nickeilf, line 48: It would be better to call this only once in the `__init__()` function of `ChineseButterFingersPerturbation`, otherwise the file reading will be executed every time when `generate()` is called.

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,80 @@
...
+
+"""
+Chinese Characters Butter Fingers Perturbation
+"""
+
+class ChineseButterFingersPerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
```
Nickeilf, line 64: I think `TEXT_TAGGING` might not be suitable for this transformation because the transformation would change the tagging labels (e.g. POS tags).

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,80 @@
...
+
+            butter_text += new_chinese_character
+        perturbed_texts.append(butter_text)
+    return perturbed_texts
+
+def get_characters_with_similar_pinyin(chinese_character, keyboard):
+    chinese_character_database = load_chinese_character_data_into_memory()
+
+    pinyin_for_char_to_be_perturbed = ""
+    if keyboard == "pinyin":
```
Nickeilf, line 33: If the variable `keyboard` only allows one value, it is not necessary to have an if-clause here, and the `keyboard` can be removed as well.

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,80 @@
...
+    random.seed(seed)
+
+    prob_of_typo = prob * 100
+    perturbed_texts = []
+    for _ in itertools.repeat(None, max_outputs):
+        butter_text = ""
+        for chinese_character in text:
+            similar_pinyins = get_characters_with_similar_pinyin(chinese_character, keyboard)
+
+            if random.choice(range(0, 100)) <= prob_of_typo and similar_pinyins != "":
```
Nickeilf, line 20: `random.random() <= prob` would be equivalent, and you don't need to multiply prob by 100.

Nickeilf (commented):
Hi @timothy22000, I think this transformation is very interesting. I've left some comments in the codes, please take a look. 
Regarding my feedback on this transformation, I think there are two points to make this transformation stronger and more realistic:
1. The current implementation is done on the character level, but in Chinese, the homophone error (a wrong word with the same Pinyin) caused by input methods is mostly on word level. _For example, "恰当的运用反义词，可以形成鲜明的对比，把事物(things)的特点表达得更充分，给人留下深刻难忘的印象。" -> "恰当的运用反义词，可以形成鲜明的对比，把食物(food)的特点表达得更充分，给人留下深刻难忘的印象。"_ It would make it more natural if this transformation can do replacements on word level. Then you might need to do word segmentation first, and then having another word dictionary and then match the words with the same Pinyins. In terms of word segmentation, you can take a look at the implementation from [another Pinyin-related PR](https://github.com/GEM-benchmark/NL-Augmenter/pull/140).
2. The replaced characters are randomly sampled as long as they share the same Pinyin with the original one. Therefore, I can see in the test.json file the replaced characters are mostly rare or less common characters. It would be good to have frequent candidates with a higher probability.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Thanks for pointing this out!

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Agreed and Addressed.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Agreed and Addressed.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Agreed and Addressed.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: I had an initial idea about adding the capability to deal with keyboards that inputs Pinyin with tone marks (e.g wǒ ài nǐ) and convert them to Chinese characters (e.g 我爱你) then perturbing them and possibly changing them back. 

Something like this keyboard input method:
https://mandarinhq.com/2020/05/type-pinyin-tone-marks/

Although I believe those sort of text is quite uncommon compared to Chinese characters so I am not sure whether this would be useful or not. I could see the conversion from pinyin tones to Chinese characters being a useful transformation.

timothy22000 (commented):
Hi @Nickeilf,

Thanks for all your comments and suggestions. I really appreciate them and it is nice to meet you. :)

I really like your ideas on improving the transformation further! I was thinking about (2) as well since I noticed that rarer characters are being returned too.

Great point on (1), thanks for the example and for sharing the link for word segmentation. I will take a look at both points.

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,79 @@
...
+        self.prob = prob
+        self.chinese_character_database = load_chinese_character_data_into_memory()
+
+    def generate(self, sentence: str):
+        perturbed_texts = chinese_butter_finger(
+            text=sentence,
+            prob=self.prob,
+            seed=self.seed,
+            max_outputs=self.max_outputs,
+            chinese_character_database = load_chinese_character_data_into_memory()
```
Nickeilf, line 76: I guess this line should be `chinese_character_database = self.chinese_character_database`, right?

transformations/chinese_butter_fingers_perturbation/transformation.py
Nickeilf: The transformation based on Pinyin and tones can be a more constrained version of the current one. I think it would be more natural as it also embeds some phonetic similarities (some people might use voice input while typing). If the tone one is not difficult to implement, I think you could keep both in this transformation. Maybe adding an argument in the `__init__()` function like `consider_tone=False` to limit whether the generated word should have the same Pinyin **and** tone as the original word.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Yes, I noticed this! I've made the change locally but I haven't push it yet. Thanks for pointing it out :)

I'll still be working on this pull request but I won't be able to work on it until after the 15th Aug since I have other commitments. Just in case you don't hear back from me until then.

evaluation/evaluate_text_classification.py
```diff
@@ -135,8 +154,10 @@ def evaluate(
 
 def _get_model_pred(model, examples, batch_size):
     all_preds = []
+    # print(model)
```
Nickeilf, line 157: This can be removed.

evaluation/evaluate_text_classification.py
```diff
@@ -145,12 +166,14 @@ def evaluate_dataset(
     accuracy = 0
     total = 0
     examples = [_get_instance_by_keys(list(raw_text)[:-1]) for raw_text in dataset]
-    #print(examples)
+    # print(examples)
```
Nickeilf, line 169: This can be removed.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Thanks for the great idea! I have implemented this as per your suggestion.

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Apologies, my PhD work took longer than I expected but I should have a working version of the word segmentation, word-level perturbation, and character-level perturbation with more common words now.  :)

evaluation/evaluate_text_classification.py
timothy22000: Addressed.

evaluation/evaluate_text_classification.py
timothy22000: Addressed

timothy22000:
Hi @Nickeilf,

I hope you had a great bank holiday weekend and that the ACL workshop on 6th Aug went smoothly! Thanks for your continued interest and reviews :)

I have made the changes below that include the word segmentation, word-level perturbation, returning common characters for character-level perturbation as per the comments on 26th July 2021. I will appreciate it if you can kindly review them as there are quite a few changes.

I also had some new ideas for transformation while working on this and will try to finish them.


** Changes **
1. Addressed all earlier comments
2. Added common_chinese_char.json
2. Added for 3500常用汉字.txt, 四十万汉语大词库.txt and thuocl_all.txt for Chinese words 词库
3. Added Chinese word segmentation
4. Added word-level perturbation
5. Improved character level perturbation to prioritize showing more common characters
6. Added capability to consider Pinyin with or without tones
7. Added capability to switch between word + char level perturbation or only char level perturbation.
8. Improved ChineseButterFingersPerturbation 
9. Modified evaluate_text_classification.py to be able to evaluate Chinese texts against CLUE Roberta and other models from HuggingFace
10. Updated test cases for the latest transformation. Noticed that the old tests that were passing before failed in the latest workflow although no code changes

**To-Do**

1. Update ReadMe
2. Fix any upcoming issues from workflow run if there are any.

mmeyer39 (changes requested):

evaluation/evaluate_text_classification.py
```diff
@@ -137,6 +156,7 @@ def _get_model_pred(model, examples, batch_size):
     all_preds = []
     with torch.no_grad():
         for e in (range(0, len(examples), batch_size)):
+            print(examples[e:e+batch_size])
```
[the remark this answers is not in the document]
timothy22000: Addressed

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+                                                                     common_chinese_character_database,
+                                                                     consider_tone)
+                if random.random() <= prob and similar_pinyins != "":
+
+                    new_chinese_character = random.choice(similar_pinyins)
+                else:
+                    new_chinese_character = chinese_character
+
+                butter_text += new_chinese_character
+        perturbed_texts.append(butter_text)
```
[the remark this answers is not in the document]
timothy22000: Addressed

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+import shutil
+import jieba
+import pypinyin
+
+
+def chinese_butter_finger(text,
+                          chinese_character_database,
+                          common_chinese_character_database,
+                          chinese_words_database,
+                          prob,
```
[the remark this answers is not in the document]
timothy22000: Hi Maxime,

I kept this as `prob` to keep it consistent with the existing English `butter_fingers_perturbation` since I thought it would be confusing to the user if one transformation used `prob `and another used `replacement_prob` - maybe it would be as a follow-up if we standardize the arguments across different transformations once all the accepted transformations are known?

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+                similar_pinyins_words = dict['similar_pinyin_words']
+                if random.random() <= prob and len(similar_pinyins_words) != 0:
+
+                    new_chinese_character = random.choice(similar_pinyins_words)
+                else:
+                    new_chinese_character = perturb_word
+                butter_text += new_chinese_character
+        else:
+            for chinese_character in text:
+                similar_pinyins = get_characters_with_similar_pinyin(chinese_character,
```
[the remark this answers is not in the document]
timothy22000: I didn't use a list for the char-level transformation since a string with all the characters with the same pinyin is sufficient for me to work with. I can convert them to lists though if you think that would be better?

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+    dirname = os.path.dirname(__file__)
+    words_list = []
+    with open(os.path.join(dirname, 'thuocl/thuocl_all.txt'), mode='r') as file:
+        lines = file.readlines()
+        for line in lines:
+            if(line.rstrip() != ""):
+                dict = {}
+                word_freq = line.split()
+                dict["word"] = word_freq[0]
+                dict["freq"] = word_freq[1]
```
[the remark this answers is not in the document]
timothy22000: If you are referring to `dict["freq"]`, I am including it in the dictionary for future features such as returning the most common words which you can be get for the word frequency. :)

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+                                                                 common_chinese_character_database,
+                                                                 consider_tone)
+            similar_word_pinyin_list = [char for char in similar_pinyins]
+
+        if(perturb_word_len > 1):
+            for i in range(len(chinese_words_database)):
+                word_dict = chinese_words_database[i]
+                word = word_dict['word']
+
+                if (len(word) == perturb_word_len):
```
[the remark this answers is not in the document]
timothy22000: Addressed

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+                                                                 rare_word_prob,
+                                                                 chinese_character_database,
+                                                                 common_chinese_character_database,
+                                                                 consider_tone)
+            similar_word_pinyin_list = [char for char in similar_pinyins]
+
+        if(perturb_word_len > 1):
+            for i in range(len(chinese_words_database)):
+                word_dict = chinese_words_database[i]
+                word = word_dict['word']
```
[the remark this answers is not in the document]
timothy22000: I renamed `perturb_word` to `original_word` and `word `to `candidate_word ` now

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+
+def get_words_with_similar_pinyin(text,
+                                rare_word_prob,
+                                chinese_character_database,
+                                common_chinese_character_database,
+                                chinese_words_database,
+                                consider_tone):
+    words_and_similar_word_dict_list = []
+    for perturb_word in text:
+        words_and_similar_word_dict = {}
```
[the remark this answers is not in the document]
timothy22000: Addressed

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,227 @@
...
+def get_characters_with_similar_pinyin(chinese_character,
+                                       rare_word_prob,
+                                       chinese_character_database,
+                                       common_chinese_character_database,
+                                       consider_tone):
+
+    pinyin_for_char_to_be_perturbed = ""
+    for char_dict in chinese_character_database:
+        if(chinese_character == char_dict['word']):
+            pinyin_for_char_to_be_perturbed = char_dict['pinyin']
```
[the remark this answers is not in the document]
timothy22000: There can be more than one char with the same pinyin (with or without tones/accents) so I would need to loop through the character database to ensure that I retrieve all of them to be saved into the `pinyin_for_char_to_be_perturbed`

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
mmeyer39: ok, fine by me :)

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
mmeyer39: it was to be coherent with the previous function but that is okay as such.

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
mmeyer39: Then I believe that there is an error here. It should be ```pinyin_for_char_to_be_perturbed += char_dict['pinyin']```.
Currently you only assign the last value found, hence my comment.

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
mmeyer39: perfect, thanks

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
mmeyer39: hum, maybe add a comment above stating that this is not used currently but could be use in future work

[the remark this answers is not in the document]
timothy22000: Done!

transformations/chinese_butter_fingers_perturbation/transformation.py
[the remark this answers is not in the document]
timothy22000: Hi @mmeyer39 ,

You are right with your first comment! I am only getting the pinyin for a single character at that part of the code since I am passing each character individually here so there is only one result :D I've updated it with `break` as you suggested

I was thinking about what I was doing in `retrieve_from_database` when I saw your comment last time so I was thinking about the wrong thing.

Nickeilf (commented):
Hi, I've just checked the updates. It seems the file "3500常用汉字.txt" is not used in the code.

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,248 @@
...
+        output = jieba.lcut(text)
+        if(word_level_perturb):
+            words_and_similar_word_dict_list = get_words_with_similar_pinyin(output,
+                                                                            rare_word_prob,
+                                                                            chinese_character_database,
+                                                                            common_chinese_character_database,
+                                                                            chinese_words_database,
+                                                                            consider_tone)
+            for dict in words_and_similar_word_dict_list:
+                perturb_word = dict['perturb_word']
```
Nickeilf, line 36: Since you've changed the `perturb_word` to `original_word` in the `get_similar_word_pinyin_list ` function, could you also change the variable names here to keep consistent?

transformations/chinese_butter_fingers_perturbation/transformation.py
```diff
@@ -0,0 +1,248 @@
...
+
+def get_words_with_similar_pinyin(text,
+                                rare_word_prob,
+                                chinese_character_database,
+                                common_chinese_character_database,
+                                chinese_words_database,
+                                consider_tone):
+    words_and_similar_word_dict_list = []
+    for original_word in text:
+        words_and_similar_word_dict = {"perturb_word": original_word}
```
Nickeilf, line 102: Maybe changing this `"perturb_word"` key to `"original_word"`

timothy22000: Done!

transformations/chinese_butter_fingers_perturbation/transformation.py
timothy22000: Done!

Nickeilf (approved):
Thanks for the changes! The transformation looks pretty good!

asnota:
Thanks for your transformation! Seems like a valuable contribution, however, the code lacks comments - could you please add doc strings to each of your functions as well as to the class, and the comments whenever it is relevant? Another suggestion - could you consider changing the name of the .txt file "四十万汉语大词库" to an English equivalent, since it doesn't tell much to non-Chinese speakers.

asnota:
It would be also nice to have some basic error handling logic, at least when the correct user input is required.

PR #175, opened 2021-07-27
[BUGFIX] (Correctly) Fail on Mismatched Numbers of Expected & Observed Examples

### LINKED ISSUE #127: Tests do not Check that Expected and Generated Outputs have Same Number of Sentences
This issue concerns the following line in the main test script:

https://github.com/GEM-benchmark/NL-Augmenter/blob/27ab1d7163100b133648846a18935cb6f2a137ee/test/test_main.py#L26

[The `zip()` builtin](https://docs.python.org/3/library/functions.html#zip) (which is used in the above-mentioned line to pair up expected sentences with generated sentences) clips the longer of its two inputted iterables to the length of the shorter iterable. E.g.:

```python
>>> list(zip([1,2,3], [6,7,8,9,10]))
[(1, 6), (2, 7), (3, 8)]
```

This means that even if a transformation generates fewer sentences (e.g. 0) than the expected number of sentences, it will still pass and the later expected sentences will not get evaluated. This also makes it impossible to test affirmatively that a transformation does not generate any outputs for a given input.

I would recommend either `assert`ing that the two iterables are of equal length, or replacing `zip()` with [`zip_longest()`](https://docs.python.org/3/library/itertools.html#itertools.zip_longest).

cascadianblue:
Fixes #127 .

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -1,3 +1,5 @@
+from itertools import zip_longest
+
 import pytest
 
 from initialize import initialize_models
@@ -24,7 +26,7 @@ def execute_sentence_operation_test_case(transformation, test):
     filter_args = test["inputs"]
     outputs = test["outputs"]
     perturbs = transformation.generate(**filter_args)
-    for pred_output, output in zip(perturbs, outputs):
+    for pred_output, output in zip_longest(perturbs, outputs):
         assert pred_output == output["sentence"], get_assert_message(
             transformation, output["sentence"], pred_output
         )
--- a/transformations/gender_culture_diverse_name/test.json
+++ b/transformations/gender_culture_diverse_name/test.json
@@ -52,9 +52,7 @@
         "retain_gender": false,
         "retain_culture": false
       },
-      "outputs": [{
-        "sentence": "Rihab Buffay is an eccentric masseuse and musician."
-      }]
+      "outputs": []
     },
     {
       "class": "GenderCultureDiverseName",
@@ -63,9 +61,7 @@
         "retain_gender": false,
         "retain_culture": true
       },
-      "outputs": [{
-        "sentence": "Severus Buffay is an eccentric masseuse and musician."
-      }]
+      "outputs": []
     },
     {
       "class": "GenderCultureDiverseName",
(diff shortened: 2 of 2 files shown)


kaustubhdhole:
Seems the build is failing because of a prediction in GenderCultureDiverseName:  

E           AssertionError: Mis-match in expected and predicted output for GenderCultureDiverseName transformation: 
E              Expected Output: Rihab Buffay is an eccentric masseuse and musician. 
E              Predicted Output: None
E           assert None == 'Rihab Buffay is an eccentric masseuse and musician.'

cascadianblue:
> Seems the build is failing because of a prediction in GenderCultureDiverseName:

@kaustubhdhole I think this is a real example of the bug. The test case is expecting one sentence to be outputted:

`expected = ["Rihab Buffay is an eccentric masseuse and musician."]`

but the actual transformation is outputting no sentences:

`actual = []`

and `pairs = zip(expected, actual) = []`, therefore, when it iterates through those `pairs`, it's just an empty loop.

cascadianblue:
Should we "fix" the test cases of the already existing transformations as part of this PR?

kaustubhdhole:
Thank you for your efforts @cascadianblue! It would be smoother to change the test cases in this PR itself as the build needs to be always kept green :)

cascadianblue:
@kaustubhdhole hmmm, I had assumed that lots of other tests would have this problem, but it looks like `gender_culture_diverse_name` might have been the only one

kaustubhdhole:
Adding @XudongOliverShen once to confirm.

cascadianblue:
@kaustubhdhole did you mean to request @XudongOliverShen for review? I only see you on the reviewers list

XudongOliverShen:
Looks good to me! 

Removing incorrect test cases is okay, so the build is green. After this PR is merged, I will update `GenderCultureDiverseName` to add back proper test cases.

kaustubhdhole:
@cascadianblue I just meant to keep @XudongOliverShen in loop :)

kaustubhdhole (approved):

kaustubhdhole:
Thanks @cascadianblue for making this change!

PR #176, opened 2021-07-27
Universal bias filter

asnota:
The filter returns "True" when the target population is underrepresented. In all other cases, the filter returns "False".
The cases returning "False" are listed below:

1. All sentences are tagged as "neutral", which is when:
   1) they do not contain the keywords neither from target, nor from test arrays,
   2) there are keywords of both - target and test arrays in the same sentence.
2. The amont of sentences tagged as "target" and "test" is equal.
3. The amont of sentences tagged as "test" is less than "target".


mille-s:
Hi, thanks for the perturbation! This will allow for creating subpopulations based on keyworkds defined as belonging to a majority or a minority. I'm wondering if basing the subpopulations on simple (atomic) keywords will allow for getting actually representative subsets. In the provided example, "He is going to make a cake." is labeled as "majority" because it contains "he"; but I would expect the combination of "he" with "baking" to be actually somehow a minority sentence, according to the typical biases (he programs, she cooks, etc.). We can also improve that subpopulation later on and make it more complex, with keyword/meaning combinations. 

Another doubt: what happens if a sentence contains 2 majority items and 2 minority items? Is it counted as neutral?

asnota:
Hi @mille-s ! Thanks for reading the code of the filter. The example you provided is a good fit for the Discrimination filter, also committed to NL-Augmenter repository, which measures whether the discrimination towards minority group takes place, using a group element (major: "he", minor: "she") and the group factor (major: "programming", minor: "backing"), thus, the phrase "He is going to make a cake." will give a "false" for the discrimination presence as the combination of majority group ("he") appears with the minority factor ("backing"). 

So the purpose of the Universal filter is to check the balance in representation according to gender, race, religion, age, etc. - the user just needs to define the lexical attributes needed to measure the kind of balance wanted in his/her language.

And you guessed right - the presence of any number of majority items with any number of minority items in the same sentence will result in a "neutral" tag for a sentence in question.

mille-s:
@asnota , thanks, sounds good!

mille-s (approved):

kaustubhdhole:
Thank you very much for your changes @asnota

One small change before this is merged - Please add the relevant [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) too.

asnota:
@kaustubhdhole I found that "rule-based" for the type of algorithm and "social-reasoning" for the skills required to differentiate between the input and the output are relevant to this filter. Let me know where exactly should I add these keywords.

kaustubhdhole:
The keywords should be added in the transformation class itself like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/55eb7c89f76d4e6414d6ef8869b8ae9db59612ad/transformations/add_hashtags/transformation.py#L92).

asnota:
Thank you for the explanation with and example, @kaustubhdhole !
I've just pushed a new version with keywords. I also added a percentage calculation to ignore small differences in the groups and reviewed the Readme file.

kaustubhdhole (approved):

PR #178, opened 2021-07-29
Added Entity Mention Replacement Transformation

uyaseen:
This transformation randomly swaps an entity mention with another entity mention of the same entity type. It can be used as a data augmentation strategy to improve the underlying sequence (NER) model.

Contributor: Usama Yaseen, Siemens AG (usama.yaseen@siemens.com)


uyaseen:
The build failed due to `RandomDeletion` tests.

tuetschek (commented):
This looks like a reasonable transformation type and the application to NER/tagging is appropriate, but I'm not sure about the actual implementation:
* Do you always need to supply sentences with matching entities as arguments? What happens if you don't? This needs to be described in the readme.
* Wouldn't it make more sense to have lists of random entities of each type, and do the replacement randomly? (Or have this as an alternative?)
* Since other transformations use spaCy already, would it make sense to use it to obtain the entity labels automatically, so you don't need to supply them?

tuetschek:
@uyaseen Thanks, all is clear now! Could you please update the readme so it's clear that you gather entitites of the same type from other sentences, and maybe remove the replacement entities from the test JSON? Or would that break the test?

uyaseen:
@tuetschek please excuse the delayed response. I have included the example and formatting constraint(s) in the readme. I think the replacement entities should be part of the test JSON as otherwise, tests will break.

tuetschek (approved):
@uyaseen No worries, thanks for expanding the readme! Looks good to me now.

csinva (approved):
Hi @**[uyaseen](https://github.com/uyaseen)**, thanks for this interesting contribution! I'm a reviewer assigned to this PR. I think the transformation is extremely clear now, after @tuetschek's review.

Here's my general review:
**Specificity**: The change is specific to anything which takes in text-level inputs and comes with entity type tags.
**Novelty**: Different from other transformations in NL-Augmenter.

**Adding New Libraries**: No new libraries added
**Test Cases**: 5 test cases added.
**Evaluating Robustness**: Robustness evaluation is not yet conducted.

csinva (approved):

uyaseen:
> **Correctness**: All checks have passed - please pull "main" once into your branch.

Hi @csinva many thanks for your review!
I have pulled the latest changes from `main`; strangely `RandomDeletion` was failing the earlier builds, even though it runs smoothly on my local environment, not sure what's wrong :/. I hope it works this time.

csinva:
Thanks, looks good to me!

kaustubhdhole:
Okay @uyaseen thank you for your changes. I have removed the test cases temporarily. Please recreate a separate PR for that. Thank you very much @mille-s and @csinva for the review.

kaustubhdhole (approved):

uyaseen:
> Okay @uyaseen thank you for your changes. I have removed the test cases temporarily. Please recreate a separate PR for that. Thank you very much @mille-s and @csinva for the review.

@kaustubhdhole Thanks! I have created a separate [PR](https://github.com/GEM-benchmark/NL-Augmenter/pull/290) for tests.

PR #180, opened 2021-07-29
Added weekday_month_abbreviation

marco-digio:


transformations/weekday_month_abbreviation/transformation.py
```diff
@@ -0,0 +1,85 @@
+import re
+from typing import List
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def weekday_month_abbreviate(text, max_outputs=1):
+    abbdreviations = {
```
Saad-Mahamood, line 9: This should be hard coded. I would like to see this as a either a JSON or XML config file so that if for example multi-lingual support is added in the future it can be done so by extended your config file.

Saad-Mahamood: Change `abbdreviations` to `abbreviations`.

Saad-Mahamood (changes requested):
Just one minor improvement needed before approval can be given.

transformations/weekday_month_abbreviation/transformation.py
marco-digio: Thank you @Saad-Mahamood for the suggestion. I have hard coded the dictionaries in 2 JSON files in the commit ea5f5d59ad84b451ac882d387eea80c5bde6115e

timothy22000:
Hi marco-digio,

Would it be possible to add the alternative abbreviations for some of the days and months into your JSON files that are not currently present? :)

Tues. - Tuesday
Thur. or Thurs. - Thursday
Sept. - September

marco-digio:
> Hi marco-digio,
> 
> Would it be possible to add the alternative abbreviations for some of the days and months into your JSON files that are not currently present? :)
> 
> Tues. - Tuesday
> Thur. or Thurs. - Thursday
> Sept. - September

Sure @timothy22000 . I added them in 5211414c6eb226599bd2f26b17d7b076494f13f5 as expansions (Tues. -> Tuesday, etc.) 
However, it is a bit more tricky to add them as abbreviations since this transformation now does not involve randomness. Adding more than one possible abbreviation of a word (Tuesday) could be done by selecting randomly one of the possible alternative abbreviations (Tue. or Tues.), but this approach is quite different from the one already implemented and it will also affect the overall speed.

juand-r (approved):
Hi @marco-digio,
I was assigned to review your transformation. I think this should be accepted.

Correctness: The code passed all the tests.
Interface: Looks good.
Applicable Tasks and Keywords: All good.
Specificity: This transformation is very specific.
Novelty: This transformation overlaps with "insert_abbreviation" in NL-Augmenter. However, I think is OK to have this separate, since (1) that transformation only does abbreviations (not expansions), and (2) that transformation uses a large number of abbreviations found on social media.
Adding new libraries: N/A
Description: The README looks good. (But if there are any publications using this kind of transformation, please add them)
Data and code source: Not present. Please add it (stating that you wrote the code and are not using any other code or data).
Paraphrasers and data augmenters: This is a very high-precision transformation.
Test cases: 5 cases, good.
Evaluating robustness: not present.
Languages: English only.  This could be expanded to other languages in the future.
Documentation: no docstrings, but the code is short and straightforward enough.

kaustubhdhole:
Thanks @marco-digio for your changes! Thank you everyone for the reviews.

kaustubhdhole (approved):

PR #182, opened 2021-07-29
Added underscore_trick

marco-digio:


tongshuangwu (approved):

transformations/underscore_trick/transformation.py
```diff
@@ -0,0 +1,42 @@
...
+
+class UnderscoreTrick(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["All"]
+
+    def __init__(self, seed=42, max_outputs=1, prob=0.05):
```
kaustubhdhole, line 31: Can you pass a variable here which takes a character as input? This could be expanded to other characters too like "-", "@", "#", etc. Might not be a good idea to create a separate transformation for each.

marco-digio: Thank you @kaustubhdhole for the suggestion. I have now implemented it in 9b2fc383d10c8bf5e2f9c6a93e8eac8a2dc3c759

transformations/underscore_trick/transformation.py
```diff
@@ -0,0 +1,54 @@
...
+    return perturbed_texts
+
+
+class UnderscoreTrick(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["All"]
```
ahoimarie, line 31: Correct me if I am wrong, but I could not find that "All" is a valid input to the language parameter. Using "All", it does not fail, but:
If you limit it to "en", you will be able to run robustness tests too (fails with "All"). Or, you may also list several or all languages (e.g. from [here](https://spacy.io/usage/models)).

ahoimarie (approved):
Hello @marco-digio,
I was assigned as one of your reviewers in this task too. 
Thanks for this contribution! 

I have left one minor comment in the code, please take a look for some potential suggestions.
Here's my review:
**Correctness**: All checks have passed.
**Interface**: I think you have chosen the correct interface.
**Applicable Tasks & Keywords**: I think you have chosen the relevant task- and keywords. 
**Specificity**: The change is not specific. 
**Novelty**: This transformation is not yet implemented in NL-Augmenter.
**Adding New Libraries**: No new libraries have been added, and there is no requirements.txt file.
**Description**: The README is clear and explains what the transformation is aiming to do (randomly replacing a space with a symbol, by default this symbol is an underscore).
**Data and code source**: Data source missing.
**Evaluating Robustness**: No robustness evaluation has been performed.

Thanks for your changes.

transformations/underscore_trick/transformation.py
```diff
@@ -0,0 +1,54 @@
...
+    keywords = [
+        "morphological",
```
AnanyaSaiB, line 33: Keywords: I'm not sure if this can be considered change can be considered morphological. noise seems appropriate.

transformations/underscore_trick/README.md
```diff
@@ -0,0 +1,20 @@
...
+Author name: Marco Di Giovanni
+Author email: marco.digiovanni@polimi.it
+Author Affiliation: Politecnico di Milano and University of Bologna
+
+## What type of a transformation is this?
+This transformation acts like a perturbation to test robustness.
+It replaces some random spaces with underscores (or even other selected symbols).
+This transformation mimics particular behaviours:
+- names of folders, files, classes, functions (e.g., underscore_trick)
+- trick word counters when there is a limited maximum number of words to insert in an online questionary/quiz.
```
AnanyaSaiB, line 13: Adding on to ahoimarie's point here, along with getting clarity / readability, I also think this application of tricking word limit counters is an interesting point and can be elaborated upon. 
Apart from underscore, you could perhaps provide examples of symbols like '@' or '-' hyphen, etc?

transformations/underscore_trick/README.md
```diff
@@ -0,0 +1,20 @@
...
+Generated transformations display high similarity to the source sentences.
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification, text generation, etc. Especially on tasks related to understanding/generating scripts.
```
AnanyaSaiB, line 17: It'll be interesting to see the performance of various models on the relevant tasks (using the evaluate.py facility)

AnanyaSaiB (changes requested):
Hi Marco, this is a neat trick. I'm curious to see how the models will fare on perturbing their inputs with this transformation.

kaustubhdhole:
Okay, gauging from the interactions above, I concur that this can be merged. Thanks @marco-digio. Also, thank you @tongshuangwu @ahoimarie and @AnanyaSaiB  for the review. @marco-digio I agree with @AnanyaSaiB. It would be great to see robustness evaluation conducted on this PR.

kaustubhdhole (approved):

PR #186, opened 2021-07-30
Added emoji_icon_transformation

marco-digio:


transformations/emoji_icon_transformation/transformation.py
```diff
@@ -0,0 +1,201 @@
...
+                    s, random.choice(dict2[k])
+                )
+    return perturbed_text
+
+
+def emoji2icon(text, seed=42, max_outputs=1, emoji_to_icon=True):
+    random.seed(seed)
+
+    icons = {
+        "smiley": [
```
tongshuangwu, line 22: I would suggest keeping these as a separate json file to make them easily expandable, rather than having these as part of the python script.

tongshuangwu (changes requested):

tongshuangwu:
Thanks for the PR! Though, I spotted an earlier, similar PR that seems to more broadly cover different kinds of emojis: https://github.com/GEM-benchmark/NL-Augmenter/pull/164 
I think you two create the mapping differently, so if you two are open to it, might be worth co-owning a PR. Otherwise, the other PR might be closer to merging, given the time and case coverage.

transformations/emoji_icon_transformation/transformation.py
marco-digio: Thank you @tongshuangwu for the suggestion. I have added the two json files in commit 55cb21bbad2ec536ee40a3648fd4a33151c03402

marco-digio:
> Thanks for the PR! Though, I spotted an earlier, similar PR that seems to more broadly cover different kinds of emojis: #164
> I think you two create the mapping differently, so if you two are open to it, might be worth co-owning a PR. Otherwise, the other PR might be closer to merging, given the time and case coverage.

Yes, I agree that #164 is very similar and I am open to collaborate. However, as spotted by the author of the other PR, there are some substantial differences, including also that my transformation not only translate icons to emojis (as they to) but also emojis to icons (that they do not). Of course this is possible only for a limited amount of icons/emojis being the number of dictionary keys much smaller than the one in the previous PR (e.g., while it is interesting to perturb the word fire with the emoji 🔥, the opposite is usually not safe to perform).

tongshuangwu (approved):
thanks for the change! And the comparisons with the other PR make sense. I'm approving this on my side.

transformations/emoji_icon_transformation/transformation.py
```diff
@@ -0,0 +1,68 @@
...
+class EmojiToIcon(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["All"]
+    keywords = ["rule-based", "visual", "high-precision", "low-coverage"]
+
+    def __init__(self, seed=42, max_outputs=1, emoji_to_icon=False):
```
Filco306, line 44: Perhaps add some type hinting and comments here? Just a small detail but always makes the code better :)

transformations/emoji_icon_transformation/text2icon.json
```diff
@@ -0,0 +1 @@
...], "angry": [">:[", ":{", ":@"], "crying": [":'\u2011(", ":'("], "tears_of_happiness": [":'\u2011)", ":')"], "disgust": ["D\u2011':", "D:<", "D:", "D8", "D;", "D=", "DX"], "surprise": [":\u2011O", ":O", ":\u2011o", ":o", ":-0", "8\u20110", ">:O"], "kiss": [":-*", ":*", ":\u00d7"], "wink": [";\u2011)", ";)", "*-)", "*)", ";\u2011]", ";]", ";^)", ";>", ":\u2011,", ";D"], "tongue": [":\u2011P", ":P", "X\u2011P", "XP", "x\u2011p", "xp", ":\u2011p", ":p", ":\u2011\u00de", ":\u00de", ":\u2011\u00fe", ":\u00fe", ":\u2011b", ":b", "d:", "=p", ">:P"], "skeptical": [":-/", ":/", ":\u2011.", ">:\\", ">:/", ":\\", "=/", "=\\", ":L", "=L", ":S"], "straight_face": [":\u2011|", ":|"], "embarassed": [":$", "://)", "://3"], "sealed_lips": [":\u2011X", ":X", ":\u2011#", ":#", ":\u2011&", ":&"], "angel": ["O:\u2011)", "O:)", "0:\u20113", "0:3", "0:\u2011)", "0:)", "0;^)"], "evil": [">:\u2011)", ":)", "}:\u2011)", "}:)", "3:\u2011)", "3:)", ">;)", ">:3", ";3"], "cool": ["|;\u2011)", "B-)"], "bored": ["|\u2011O"], "tongue_in_cheek": [":\u2011J"], "confused": ["%\u2011)", "%)"], "sick": [":\u2011###..", ":###.."], "disbelief": ["',:-|", "',:-l"], "awkward": [":E"], "skull": ["8-X", "8=X", "x-3", "x=3"]}
```
Filco306, line 1: Is there support for more emojis than these ones perhaps? there are more emojis than these ones, I suppose?

Filco306 (changes requested):
I read the comments of @tongshuangwu and agree that it is very similar to #164, but I think this PR still has value. I just had some small changes requested; otherwise I think the PR looks good :)

transformations/emoji_icon_transformation/transformation.py
marco-digio: Thank you @Filco306 for the suggestion, I have added type hinting and some small comments in 5f610c740a9f32aeaa20c59952e2560240a61dcb. Now I agree that the code is much easier to understand!

transformations/emoji_icon_transformation/text2icon.json
marco-digio: The transformation is written so that new emojis and icons can be easily added just by updating the JSON files. I have used [Wikipedia](https://en.wikipedia.org/wiki/List_of_emoticons) as a reliable source of equivalence between emojis and icons but I agree that many more emojis could be added, if there is at least one icon that is used with the same meaning. I believe that the most used ones are in the JSON files right now but I am available to update them with new ones in the future.

Filco306 (approved):
Changes are good, I approve!

kaustubhdhole:
Thank you @tongshuangwu and @Filco306 for the review. 

Thank you @marco-digio for the changes. I am merging this now but I would suggest you to create a separate PR for the robustness evaluation.

kaustubhdhole (approved):

rteehas:
I saw this listed as one I should review, but it looks like it's already merged in. Regardless, it looks good to me as well

PR #188, opened 2021-07-30
Add Urban Thesaurus transformation

richplant:
Randomly swaps out nouns for related terms drawn from the Urban Dictionary.


RyanGoslingsBugle:
Unit tests won't work for this transform, since the output isn't deterministic.

kaustubhdhole (commented):
Thanks @richplant - This seems to be a great addition to NL-Augmenter!

transformations/urban_dict_swap/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+    http.mount("https://", adapter)
+
+    def __init__(self, seed=0, prob=0.5, min_score=100.0, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.seed = seed
+        self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
+        self.prob = prob
+        self.min_score = min_score
+        self.max_outputs = max_outputs
+        self.url = "https://urbanthesaurus.org/api/related?term="
```
kaustubhdhole, line 43: Aren't there APIs for synonymy relationships? Might be specific as compared to "related" ?

kaustubhdhole:
For unit tests, I would suggest you to add a blank file like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/diverse_paraphrase/test.json).

transformations/urban_dict_swap/transformation.py
kaustubhdhole: Also can the "min_score" be used to reduce the size of the candidates called? UrbanDictionary is a pretty large dictionary.

richplant: Unfortunately, the API is undocumented, and I haven't been able to discover any other filter params. The official Urban Dictionary API doesn't support related terms/synonymy searching, just definitions.

transformations/urban_dict_swap/test.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "urban_dict_swap",
+    "test_cases": [
```
gxywang, line 3: It is recommended to include at least 5 examples here.

gxywang (changes requested):
Hi @richplant, I'm one of the assigned reviewers.

Correctness: Looks good, all checks have passed.
Interface: Looks good as well.
Applicable Tasks & Keywords: Task types and keywords have been added.
Specificity: The task is specific to text classification and generation. Tagging is mentioned in the py file, but I'm not sure if this task is related to tagging specifically.
Novelty: This transformation is not implemented in NL-Augmenter already.
Adding New Libraries: None.
Description: Readme is well organized and thorough.
Data and code source: None. Should add a sentence stating it was fully created by author.
Paraphrasers and Data Augmenters: N/A
Test Cases: No test cases have been added, see comment.
Evaluating Robustness: Evaluation is conducted on text classification only. Perhaps add one for text generation as well?
Languages other than English: N/A
Decent Programming Practise: Docstrings look good.

If anything needs clarification, please let me know. Otherwise, thanks for the contribution!

transformations/urban_dict_swap/transformation.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Iterate through input sequence, obtain related terms for nouns from API
+        @param text: Input sequence
+        @return: List of transformed output sequences
+        """
+        random.seed(self.seed)
+        results = []
+        for _ in range(self.max_outputs):
+            new_text = []
+            for token in self.nlp(text):
```
kaustubhdhole, line 83: Seems you are recomputing the spacy parse multiple times. Maybe this could be done outside the `self.max_outputs` loop

kaustubhdhole:
Also for unit tests, try testing this command on your local: `pytest -s --t=light --f=light`

transformations/urban_dict_swap/test.json
richplant: Hi, thanks for the review. I'm afraid I can't add anything to the test.json, since the build pipeline interprets them as static test cases. Since the transformation is non-deterministic, it'll fail the build.

On the point about tagging, I would argue that swapping out nouns for synonyms is quite a useful perturbation when training a POS tagger.

Since this wholly relies on an API which I did not develop, I am not comfortable claiming ownership over the whole transform. However, I have identified myself as the author in the README.

transformations/urban_dict_swap/transformation.py
richplant: Yep, silly me, very inefficient. Fixed in 2de3222.

transformations/urban_dict_swap/test.json
richplant: I've added a text generation benchmark to the readme in 2de3222

gxywang: Cool, thanks for the changes.

gxywang (approved):
Accept.

transformations/urban_dict_swap/test.json
kaustubhdhole: Okay, if the build is failing, you can add sample examples in the README. It will give an idea of what this transformation generates - both good and bad.

kaustubhdhole (approved):

kaustubhdhole:
Thank you @gxywang for your comments and @richplant for your hard work!

PR #191, opened 2021-08-01
Added polarity filter

maobedkova:


filters/polarity/filter.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__()
+        self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
+        if "spacytextblob" not in self.nlp.pipe_names:
+            self.nlp.add_pipe("spacytextblob")
+        self.diff_allowed = diff_allowed
+
+    def filter(self, sentence1: str = None, sentence2: str = None) -> bool:
+        text1 = self.nlp(sentence1, disable=["parser", "tagger", "ner", 'lemmatizer'])
+        text2 = self.nlp(sentence2, disable=["parser", "tagger", "ner", 'lemmatizer'])
+        if abs(text1._.polarity - text2._.polarity) <= self.diff_allowed:
```
tongshuangwu, line 22: I wonder if you have considered cases on the edge: e.g., a sentence is marginally positive, and after the change it becomes marginally negative --- would there be cases where the label is changed but the diff still satisfy the `self.diff_allowed ` constraint? Should there be a global threshold?

requirements.txt
```diff
@@ -29,3 +29,6 @@ black==21.5b2
 pre-commit==2.13.0
 flake8==3.9.2
 isort==5.9.1
+
+# for sentiment filter
+spacytextblob==3.0.1
```
tongshuangwu: Please create a separate `requirements.txt` file under your folder, so to avoid disturbing the main `requirements.txt` :)

tongshuangwu (changes requested):

filters/polarity/filter.py
maobedkova: The outlined case is possible but in terms of sentiment analysis tasks, I would say it is acceptable. Considering that sentiment scoring tools are not always consistent and cases of near-neutral polarity are mainly problematic for polarity interpretation, we don't want to restrict transformed cases here much. However, potentially this could be a nice argument to introduce: to allow or not to allow any kind of sentiment shift (even from -0.1 to 0.1).

maobedkova: I added another argument that can do a stricter polarity check.

tongshuangwu (approved):
Thanks for the changes! I think the pr is good to go :)

filters/polarity/filter.py
```diff
@@ -0,0 +1,26 @@
...
+from initialize import spacy_nlp
+from spacytextblob.spacytextblob import SpacyTextBlob
+import spacy
+
+
+class TextRetainsPolarity(SentencePairOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self, strict_polarity: bool = False, diff_allowed: float = 0.5):
```
Filco306, line 12: Can you add some comments and documentation for the input arguments?

filters/polarity/test.json
```diff
@@ -0,0 +1,96 @@
+{
```
Filco306, line 1: Could you add some longer examples and check how small changes will change the overall polarity?

Filco306: SImply, make sure that a small change gives a proportionally small polarity change in a long input.

Filco306 (commented):
I think this PR is a very nice one! Simple, efficient and simply a nice transformation. Just some minor comments that could be improved, but well done overall! :D

Filco306:
I feel a bit stupid now; I see that when I reviewed this, it was already merged in. Anyway, good work!

kaustubhdhole:
No problem @Filco306. If possible, it would be great to have comments addressed via a separate PR.

PR #201, opened 2021-08-06
Added contextual_meaning_perturbation

agesb:
This transformation changes the _meaning_ of the sentence while avoiding the introduction of grammar, spelling and logical mistakes. I found it useful for my work with machine translation quality estimation models.

### DIFF
--- a/transformations/contextual_meaning_perturbation/README.md
+++ b/transformations/contextual_meaning_perturbation/README.md
@@ -1,0 +1,45 @@
+# Contextual Meaning Perturbation🦎  + ☎️️ → 🐍
+This transformation changes the meaning of the sentence while avoiding grammar, spelling and logical mistakes.
+
+
+## What type of a transformation is this?
+This transformation's effect is comparable to the <i>"Chinese Whispers"</i> or <i>"Telephone"</i> children's game: The transformed sentence
+appears fluent and somewhat logical, but the meaning of the original sentence is not preserved.
+To achieve logical coherence, a pre-trained language model is used to replace words with alternatives that match the context of the sentence.
+Grammar mistakes are reduced by limiting the type of words considered for changes (based on POS tagging) and replacing adjectives with adjectives,
+nouns with nouns, etc. where possible.
+
+
+## What tasks does it intend to benefit?
+This transformation benefits users who seek perturbations that preserve fluency but not the meaning of the sentence.
+For instance, it can be used in scenarios where the meaning is relevant to the task, but the model shows a tendency
+to over-rely on simpler features such as the grammatical correctness and general coherence of the sentence.
+A real-world example would be the training of quality estimation models for machine translation
+(does the translation maintain the meaning of the source?) or for text summarisation (does the summary capture the content of the source?).
+
+
+
+## What kind of configurations are tweakable?
+- <b>pos_to_change</b>: Determines which tokens are eligible for change based on the part of speech – default: ['ADV','ADJ', 'VERB', 'NOUN', 'PROPN']
+- <b>perturbation_rate</b>: The percentage of eligible tokens that should be changed – default: 0.3
+- <b>language</b>: Currently supported languages include English (en) and German (de) – default: "en"
+- <b>top_k</b>: Determines how many replacement candidates should be proposed – default: 10
+
+
+## Previous Work
+Word substitution with pre-trained language models has been explored in different settings. For example, the augmentation
+library [nlpaug](https://github.com/makcedward/nlpaug "nlpaug") and the adversarial attack library
+[TextAttack](https://github.com/QData/TextAttack "TextAttack") include contextual perturbation methods.
+However, their implementations do not offer control over the type of words that should be perturbed and introduce
+a large number of grammar mistakes. If the aim is to change the sentence's meaning while preserving its fluency,
+this transformation can help to get the same effect with significantly fewer grammatical errors. Li et al. (2021) propose an alternative approach to achieve a similar objective in their paper
+"Contextualized Perturbation for Textual Adversarial Attack".
+
+
+## What are the limitations of this transformation?
+Without using an additional grammar correction tool, it is difficult to avoid grammar mistakes completely.
+With the pre-configured settings, the risk of introducing small grammar mistakes
+is about 5% for English and about 10% for German data for sentences of average length. The meaning of the original sentence is preserved in <1% of the cases (as intended).
+Lowering the perturbation rate increases the risk of paraphrasing the sentences without changing their meaning.
+
+Since this transformation requires the use of a pre-trained language model (via HuggingFace), it is comparatively heavy-weighted.
--- a/transformations/contextual_meaning_perturbation/requirements.txt
+++ b/transformations/contextual_meaning_perturbation/requirements.txt
@@ -1,0 +1,4 @@
+transformers==4.8.2
+
+# for contextual_perturbation (DE) download DE SpaCy model:
+de-core-news-sm==2.2.5
--- a/transformations/contextual_meaning_perturbation/test.json
+++ b/transformations/contextual_meaning_perturbation/test.json
@@ -1,0 +1,52 @@
+{
+  "type": "contextual_meaning_perturbation",
+  "test_cases": [
+    {
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "Andrew finally returned the French book to Chris that I bought last week"
+      },
+      "outputs": [{
+         "sentence": "I finally returned the French book to Chris that I read last year"
+      }]
+    },
+    {
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "The motorized postal line to Hollfeld was abandoned again in the autumn of 1931 because the abutting communities of Zoggendorf and Stücht were not using the service enough for it to be justifiable."
+      },
+      "outputs": [{
+        "sentence": "The first postal line to Hollfeld was closed again in the summer of 1931 because the abutting communities of Zoggendorf and Stücht were not using the line enough for it to be available."
+      }]
+    },
+    {
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "Alice in Wonderland is a 2010 American live-action/animated dark fantasy adventure film"
+      },
+      "outputs": [{
+        "sentence": "Woman in Wonderland is a 2010 American live-action/animated western fantasy adventure series"
+      }]
+    },
+    {
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "Lord Bradley is also a Special Adviser to the President and Vice - Chancellor of the University of Manchester."
+      },
+      "outputs": [
+        {
+          "sentence": "Mr Bradley is also a Special Adviser to the President and Ex - Chancellor of the University of Cambridge."
+        }
+      ]
+    },
+    {
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "She is a strong critic of homeschooling."
+      },
+      "outputs": [{
+        "sentence": "She is a strong advocate of homeschooling."
+      }]
+    }
+  ]
+}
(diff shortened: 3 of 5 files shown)


kaustubhdhole:
Hi @agesb Either the meaning has to remain intact or it should be evident what type of example it produces - i.e. the new class/label should be known if in case the label changes (eg. sentiment goes from positive to negative)! In the former case, it would be useful as a perturbation as well as a paraphraser. In the latter case, it would help in creating more examples for the task. Your particular transformation makes lexical replacements in text and clearly changes label & meaning. I would suggest you to use a task like QuestionAnswering and maybe choose words which are common to both the context and the question and then apply a masking over those words (in the context). You can use the new word(s) to replace in the question too.

transformations/contextual_meaning_perturbation/requirements.txt
```diff
@@ -0,0 +1,4 @@
+transformers==4.8.2
+
+# for contextual_perturbation (DE) download DE SpaCy model:
+de-core-news-sm==2.2.5
```
Nickeilf, line 4: The build fails with the spacy model.

For adding spacy model as a dependency in requirements.txt, you may take a look here:
https://stackoverflow.com/questions/61702357/how-to-add-a-spacy-model-to-a-requirements-txt-file

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+            num_word_swaps = 0
+        else:
+            num_word_swaps = max(
+                1,
+                round(self.percentage * self.count_eligible_tokens(pos_tags)),
+            )
+        if self.verbose:
+            print(num_word_swaps, " tokens will be masked.\n")
+        random.seed(self.seed)
```
Nickeilf, line 109: It would be good to put this line to the __init__() function

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,172 @@
...
+            [
+                token
+                for i, token in enumerate(tokens)
+                if pos_tags[i] in self.pos_to_change
+            ],
+            num_word_swaps,
+        )
+        for token in tokens_to_mask:
+            masked_input = input.replace(token, "<mask>", 1)
+            perturbation_dict[token] = self.unmasker(masked_input)
```
Nickeilf, line 120: A minor suggestion here:

The current implementation only samples the token, but not the position. That is to say, if there are duplicate tokens in the sentence, they might be masked together (even though only one token is sampled).
For example, the input sentence is "I ate an apple and bought an apple phone ." and we only sampled the first "apple", but actually both "apple" will be masked in the for loop.

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,172 @@
...
+        perturbation_dict = self.get_perturbation_candidates(input)
+        tokens, pos_tags, lemmas = self.get_linguistic_features(input)
+
+        replacement_dict = {}
+        for original_token in perturbation_dict:
+            replacement_dict[original_token] = perturbation_dict[
+                original_token
+            ][0]["token_str"]
+
+            # Replace with the first best choice in case no better candidate is found
```
Nickeilf, line 140: Could you implement returning a list of perturbed sentences with the length of self.max_outputs (If there are not that many outputs you can just return those meeting the POS requirements)?

agesb: Good point, thank you. I just added the functionality, the transformation now returns the number of permutations specified in self.max_outputs (or alternatively all valid permutations if that number is smaller than self.max_outputs).

transformations/contextual_meaning_perturbation/transformation.py
agesb: Done!

transformations/contextual_meaning_perturbation/requirements.txt
agesb: Ah I see, that makes sense. I removed it from the requirements.txt since it is not necessary to download the German model to use the transformation in general.

agesb:
> Hi @agesb Either the meaning has to remain intact or it should be evident what type of example it produces - i.e. the new class/label should be known if in case the label changes (eg. sentiment goes from positive to negative)! In the former case, it would be useful as a perturbation as well as a paraphraser. In the latter case, it would help in creating more examples for the task. Your particular transformation makes lexical replacements in text and clearly changes label & meaning. I would suggest you to use a task like QuestionAnswering and maybe choose words which are common to both the context and the question and then apply a masking over those words (in the context). You can use the new word(s) to replace in the question too.

Hi @kaustubhdhole, thanks so much for your input! I really like your ideas but I’m not sure if it’d be possible to implement them while still fulfilling the objective I was aiming for. I developed the transformation to be used in the field of Machine Translation Quality Estimation (MTQE). The QE models tend to pay too much attention to the grammatical correctness of the machine-generated translation, ignoring the fact that the meaning of the source sentence is not preserved. By changing the meaning of the translation and keeping its grammar intact, this transformation helps to create samples that target this specific issue. I absolutely understand that this application might be a bit niche, but hope that others can benefit from the transformation with its current functionality. Maybe it’d be best to have separate transformations for QuestionAnswering and sentiment prediction instead of mixing the objectives in this one?

kaustubhdhole:
Thanks @agesb for the clarification. Yes, I think its fine to have separate transformations in that case. Also, please include QUALITY_ESTIMATION as a type of task in your transformation.

transformations/contextual_meaning_perturbation/README.md
```diff
@@ -0,0 +1,45 @@
+# Contextual Meaning Perturbation🦎  + ☎️️ → 🐍
+This transformation changes the meaning of the sentence while avoiding grammar, spelling and logical mistakes.
+
```
kaustubhdhole, line 3: Also, your name/email/affiliation too

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def get_perturbation_candidates(self, input: str):
+        """
+        Samples tokens that will be replaced.
+        Generates and returns the top k contextual replacement candidates.
+        """
+        perturbation_dict = {}
+        tokens, pos_tags, lemmas = self.get_linguistic_features(input)
+        if self.count_eligible_tokens(pos_tags) == 0:
+            print(
```
tia-e, line 99: why not use logger (with different level) to handle logging?

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,192 @@
...
+        By default, the best fitting candidate is used.
+        If the top k candidate list includes a token with the same POS tag but different word family
+        as the original token, it will be used instead of the default.
+        """
+        perturbation_dict = self.get_perturbation_candidates(input)
+        tokens, pos_tags, lemmas = self.get_linguistic_features(input)
+
+        replacement_dict = defaultdict(list)
+        for original_token in perturbation_dict:
+            #    replacement_dict[original_token] = perturbation_dict[
```
tia-e, line 139: Could u remove commented code?

transformations/contextual_meaning_perturbation/README.md
agesb: Thanks for your feedback @kaustubhdhole, I just added my name, email and affiliation as well as the QUALITY_ESTIMATION task type.

transformations/contextual_meaning_perturbation/transformation.py
agesb: Thanks for the suggestion, I replaced the verbose parameter and prints with Python logging (info & warning). I also deleted the commented code.

transformations/contextual_meaning_perturbation/test.json
```diff
@@ -0,0 +1,52 @@
...
+      "class": "ContextualMeaningPerturbation",
+      "inputs": {
+        "sentence": "She is a strong critic of homeschooling."
+      },
+      "outputs": [{
+        "sentence": "She is a strong advocate of homeschooling."
+      }]
+    }
+  ]
+}
```
mmeyer39, line 52: Hello,
Could you add a new line at the end of the file?

transformations/contextual_meaning_perturbation/README.md
```diff
@@ -0,0 +1,49 @@
...
+(does the translation maintain the meaning of the source?) or for text summarisation (does the summary capture the content of the source?).
+
+
+
+## What kind of configurations are tweakable?
+- <b>pos_to_change</b>: Determines which tokens are eligible for change based on the part of speech – default: ['ADV','ADJ', 'VERB', 'NOUN', 'PROPN']
+- <b>perturbation_rate</b>: The percentage of eligible tokens that should be changed – default: 0.3
+- <b>language</b>: Currently supported languages include English (en) and German (de) – default: "en"
+- <b>top_k</b>: Determines how many replacement candidates should be proposed – default: 10
+
```
mmeyer39, line 27: Could you add the results of evaluate.py on the tasks intended to be addressed by the transformation?

mmeyer39 (changes requested):

transformations/contextual_meaning_perturbation/test.json
agesb: Done, thank you!

transformations/contextual_meaning_perturbation/README.md
agesb: Hi @mmeyer39, my transformation is mostly intended to mitigate partial-input biases for quality estimation tasks, as an addition to a labelled quality estimation dataset. I could evaluate against the sentence classification tasks, but I don't think this would add much value in the README. I added information on the behaviour of the transformation when using the pre-configured parameters – I believe that might be more helpful than reporting benchmark results. What's your opinion on that? Many thanks! :)

mmeyer39: Hi, 
I mentioned it because it was asked in the email from the organizers. However, it is okay as such for me :)

TGessCZ:
@agesb I wanted to ask "NLP model to generate SEO titles". Did not really find linkedin or anything. :) Just this git. If possible send me email to tomxgessxtom@gmail.com Thanks!

transformations/contextual_meaning_perturbation/README.md
```diff
@@ -0,0 +1,47 @@
...
+Word substitution with pre-trained language models has been explored in different settings. For example, the augmentation
+library [nlpaug](https://github.com/makcedward/nlpaug "nlpaug") and the adversarial attack library
+[TextAttack](https://github.com/QData/TextAttack "TextAttack") include contextual perturbation methods.
+However, their implementations do not offer control over the type of words that should be perturbed and introduce
+a large number of grammar mistakes. If the aim is to change the sentence's meaning while preserving its fluency,
+this transformation can help to get the same effect with significantly fewer grammatical errors. Li et al. (2021) propose an alternative approach to achieve a similar objective in their paper
+"Contextualized Perturbation for Textual Adversarial Attack".
+
+
+## What are the limitations of this transformation?
```
Nickeilf, line 45: Could you add a warning that this transformation would change the sentence meaning, therefore when using this transformation, the user should be cautious that the task-specific label for the generated sentences might not be identical to the original sentence? Normally, we assume that SentenceOperation would not change the labels.

transformations/contextual_meaning_perturbation/transformation.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class ContextualMeaningPerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.QUALITY_ESTIMATION,
+    ]
+    languages = ["en", "de"]
+    heavy = True
+
```
Nickeilf, line 20: Could you please add keywords for this transformation?

Check [available keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) and [this example](https://github.com/GEM-benchmark/NL-Augmenter/blob/84db89bce28169bec1eaad240cb70d3d89f4c344/transformations/butter_fingers_perturbation/transformation.py#L82)

kaustubhdhole:
@tia-e could you please get in touch. I am unable to find your contact details.

PR #202, opened 2021-08-07
Token replacement

mnamysl:
Added a transformation that replaces input tokens with their perturbed versions sampled from the lookup tables with plausible lexical replacements. By default, lookup tables that contain OCR-ed and misspelled versions of the original tokens are employed.


mnamysl:
Just realized that my transformation is closely related to the _Replace Spelling Perturbation_ (#133 and #211) but offers better coverage of target words and replacement candidates and offers more elaborate filtering options. In addition to the common misspellings, it also covers OCR errors. 

I've tested the lookup table employed by the former transformation ([spell_errors.json](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/replace_spelling/spell_errors.json)) and it can be utilized out-of-the-box by my method.

Would you consider merging both approaches if my PR will also pass the review process?

mille-s:
Thanks for the perturbation! The proposal is also linked to the [Butterfingers typos perturbation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/butter_fingers_perturbation), and to the [character swap perturbation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/swap_characters), proposed recently. There should be some discussion regarding if/how to merge similar proposals to avoid very similar perturbations, which could result confusing. I like the fact that the proposed perturbation relies on plausibility, and it that includes OCR errors too.

mnamysl:
@mille-s Thank you for your kind review! I agree that some discussion regarding similar proposals is needed.

I have also seen that #213 proposes to correct common misspellings and also employs a (built-in) lookup-table, so it could also be integrated into the proposed approach.

mille-s (approved):

transformations/token_replacement/lookup_table_utils.py
```diff
@@ -0,0 +1,90 @@
...
+Lookup table handling utilities.
+"""
+
+def get_token_replacement(input_token, lookup, max_dist):
+    """
+    Gets a replacement for an input token sampled from a given lookup table.
+    Samples uniformly among all candidates available for an input token.
+    Accepts additional filtering criteria based on the maximum edit distance 
+    between the original and the candidate token.
+    """
```
AbinayaM02, line 21: It would be nice to add param details also like you had added in transformation.py file.

AbinayaM02 (commented):
Hi @mnamysl : Thank you for your contribution. Overall, your PR covers a lot more ground on spelling mistake/typo kind of perturbation. We'll think of combining similar type of transformations later.

 For now, I have provided some minor comments on adding parameter details in all the .py files wherever it is missing in the docstring. Please add them and then will approve the PR.

transformations/token_replacement/lookup_table_utils.py
mnamysl: @AbinayaM02: Thank you for your kind and helpful feedback! 

I've just added missing docstrings and fixed some minor inconsistencies. 
I would be grateful if you could re-review my recent code changes.

AbinayaM02 (changes requested):
Hi @mnamysl: Thank you for addressing the requested changes. Here are my comments,

- The transformation addresses token replacement using OCR versions of the tokens from a lookup table. [This can be merged with other existing typo related transformations.]
- New library is used.
- Test cases have been added.

Have you added the robustness evaluation to the leaderboard in the evaluation folder. If not done, could you please add there too? Also, pull the latest code from the main branch since many PR were merged recently.

mnamysl:
Hi @AbinayaM02: Thank you for your feedback!

I've merged the latest code from the main branch and updated the text classification leaderboard as requested. 

I would be glad to help you merge the typo-related transformations later.

AbinayaM02 (approved):
@mnamysl : Thank you so much for making the requested changes. And sure, I'll keep you in the loop for the merging of all typo-related transformations later, once we finalize the approach. I'm approving the PR now :)

transformations/token_replacement/requirements.txt
```diff
@@ -0,0 +1 @@
+python-Levenshtein>=0.12.1
```
kaustubhdhole, line 1: I am sorry to be the bad guy here but unfortunately this is a library whose license would not be permissible (GNU GPL). Could you please find another library or add a function to implement levenshtein?

The rest of the changes are already approved by two other reviewers, so should be fine to merge. :)

transformations/token_replacement/transformation.py
```diff
@@ -0,0 +1,129 @@
...
+        :param max_outputs: max. number of outputs to generate
+        :param replacement_prob: replacement probability [0.0, 1.0]
+        :param min_length: min. length of token that can be perturbed
+        :param max_dist: max. Levenshtein distance between the original and the
+                         perturbed token
+        :param lookup: dictionary or a list of files with lookup tables
+        """
+        super().__init__(seed, max_outputs=max_outputs)
+        #self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
+        self.nlp = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 41: Also one minor suggestion to reduce memory. This line can be replaced with the global spacy:
`from initialize import spacy_nlp`
`self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")`

transformations/token_replacement/requirements.txt
mnamysl: Hi @kaustubhdhole: thank you for your review and helpful feedback! 

And don't worry, these changes require minimal effort :-) I've just replaced [python-Levenshtein](https://pypi.org/project/python-Levenshtein) with [editdistance](https://pypi.org/project/editdistance) (MIT license). I hope it fits now, otherwise please let me know, I found some other alternatives.

transformations/token_replacement/transformation.py
mnamysl: Thank you @kaustubhdhole for this suggestion!

If I use the global spacy object, the pytest test fails (the same problem as described here: #209). 

I've applied the fix suggested [here](https://github.com/GEM-benchmark/NL-Augmenter/issues/209#issuecomment-900407032) (thanks @AbinayaM02!) to reset the tokenizer of the global object in the generate method, so my transformation can work as expected. Still, I need to restore the old tokenizer afterwards, so other transformations can pass the pytest tests too. I suspect that some test.json files were generated using the already modified tokenizer and they fail when the original tokenizer is used. 

Please look into the transformation.py file for the workaround I've built in. I feel that this issue concerns many other transformation. I hope that @AbinayaM02 comes up with a nice solution for this issue :)

PR #204, opened 2021-08-10
Added transformation to replace nouns with hyponyms/hypernyms

AnanyaSaiB:
I had previously opened a draft pull request for the same transformation. Due to a few glitches, I am opening a new pull request after making the modifications suggested by the reviewers (@kaustubhdhole and @ashish3586 ) on the draft. In particular, 
1) the test cases are corrected in test.json file, 
2) I have added more details about the transformation in the README.md file
3) I've randomised the order of the nouns in the sentence so that every noun gets a fair chance at being picked first, rather than the order in which the nouns appear. 
4) I've removed any arbitrary bits of commented code. Currently only the code used to generate the outputs present in test.json file is present as a multi-line comment at the end of transformation.py.

Thanks for the reviews and inputs. I'll be on the look-out for any more suggestions.

### DIFF
--- a/transformations/replace_with_hyponyms_hypernyms/transformation.py
+++ b/transformations/replace_with_hyponyms_hypernyms/transformation.py
@@ -1,0 +1,114 @@
+import numpy as np
+import spacy
+import random
+from checklist.editor import Editor
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class ReplaceHyponyms(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self, n=1, seed=0, max_output=2):
+        super().__init__(seed)
+        self.nlp = spacy.load("en_core_web_sm")
+        self.n = n
+        self.max_output = max_output
+        self.editor = Editor()
+
+    def generate(self, sentence: str):
+        np.random.seed(self.seed)
+        words = []
+        perturbed_texts = []
+        tokens = self.nlp(sentence)
+        #Shuffle the tokens list so that all noun (and not just the beginning nouns) 
+        #have a fair chance at being picked.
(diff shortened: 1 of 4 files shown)


transformations/replace_with_hyponyms_hypernyms/transformation.py
```diff
@@ -0,0 +1,132 @@
+import numpy as np
+import spacy
+import random
+from checklist.editor import Editor
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class ReplaceHyponyms(SentenceOperation):
```
ashish3586, line 10: Please add DocString for the class ReplaceHyponyms.

transformations/replace_with_hyponyms_hypernyms/transformation.py
```diff
@@ -0,0 +1,132 @@
...
+              if len(perturbed_texts)>=self.max_output:
+                break
+        perturbed_texts = (
+            perturbed_texts[: self.max_output]
+            if len(perturbed_texts) > 0
+            else [sentence]
+        )
+        return perturbed_texts
+
+class ReplaceHypernyms(SentenceOperation):
```
ashish3586, line 56: Please add DocString for the class ReplaceHypernyms.

transformations/replace_with_hyponyms_hypernyms/transformation.py
```diff
@@ -0,0 +1,132 @@
...
+        "external-knowledge-based",
+        "tokenizer-required",
+        "meaning-alteration",
+        "high-precision",
+        "high-generations",
+    ]
+
+    def __init__(self, n=1, seed=0, max_output=2):
+        super().__init__(seed)
+        self.nlp = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 71: You might want to use spacy like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/d39f0e5f57a79937296351996617ac03ee80fb74/transformations/close_homophones_swap/transformation.py#L55).

PR #206, opened 2021-08-13
Ocr perturbation

mnamysl:
This PR introduces the "OCR perturbation" transformation, which directly induces Optical Character Recognition (OCR) errors into the input text. It renders the input sentence as an image and recognizes the rendered text using an OCR engine.


transformations/ocr_perturbation/transformation.py
```diff
@@ -0,0 +1,177 @@
...
+            return spacy.load("fr_core_news_sm")
+        elif self.language == 'es':
+            return spacy.load("es_core_news_sm")
+        elif self.language == 'de':
+            return spacy.load("de_core_news_sm")
+        else:
+            return None
+
+"""
+# Sample code to demonstrate usage. Can also assist in adding test cases.
```
Saad-Mahamood, line 142: I would add the Sample code to a separate file and add a reference to it in the `README.md`.

Saad-Mahamood (changes requested):
Just one minor change required before approval can be given.

transformations/ocr_perturbation/transformation.py
mnamysl: Done. Thank you @Saad-Mahamood for your feedback!

mnamysl:
Any ideas what can be done with the following automatic check error:

`Collecting threadpoolctl>=2.0.0
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out. (read timeout=15)")': /packages/c6/e8/c216b9b60cbba4642d3ca1bae7a53daa0c24426f662e0e3ce3dc7f6caeaa/threadpoolctl-2.2.0-py3-none-any.whl
  Downloading threadpoolctl-2.2.0-py3-none-any.whl (12 kB)
INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. If you want to abort this run, you can press Ctrl + C to do so. To improve how pip performs, tell us what happened here: https://pip.pypa.io/surveys/backtracking
Error: The operation was canceled.` ?

It seems to work fine locally on my computer. Could it be due to some dependency issues between different transformations?

mnamysl:
> Just one minor change required before approval can be given.

@Saad-Mahamood Thank you for your kind review! 

I've addressed the issues that you reported. Is there anything that I could further improve?

Saad-Mahamood (approved):
Looks good to me. Approved.

Saad-Mahamood:
@kaustubhdhole This PR needs a second reviewer can you nominate someone to take a look at this PR?

kaustubhdhole:
Thanks @mnamysl for your changes! I think this is a great PR. Adds OCR noise which is of course a great perturbation idea to test for robustness. The code is well written and conforms to the review criteria except for 
1) the missing [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md). 
2) the failing build due to inability to install tesseract.

mnamysl:
Hi @kaustubhdhole: thank you for your kind review and your feedback!

I've just added the keywords and merged the commits from the main branch.

What would be the best way to handle the second issue? 

README.md already includes all prerequisites (e.g, libtesseract and libleptonica installation, as well as the download and setup of the tessdata files) but is there an easy way to automate this?

mnamysl:
@kaustubhdhole: I've addressed the Tesseract installation issue and included some minor improvements. 

Please consider re-reviewing the PR after these changes :-)

kaustubhdhole:
Thanks for your changes @mnamysl. But since you are making changes to the workflow, let me ask @AbinayaM02's views here.

AbinayaM02:
> Thanks for your changes @mnamysl. But since you are making changes to the workflow, let me ask @AbinayaM02's views here.

Hi @mnamysl : It's good to have the generate_test_cases method in a separate file say test/helper.py. The reason being, there are many transformations using custom json structure for input/output and this function needs to be extended for different task types. For now, move this function to new file as specified above and later we can extend this functionality.

.github/workflows/nl-action.yml
```diff
@@ -24,6 +24,11 @@ jobs:
       with:
         path: ${{ env.pythonLocation }}
         key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('requirements.txt') }}
+    - name: Install Tesseract
+      run: |
+        sudo apt-get update
+        sudo apt-get install tesseract-ocr libtesseract-dev libleptonica-dev pkg-config \
+          tesseract-ocr-eng tesseract-ocr-osd tesseract-ocr-deu tesseract-ocr-fra tesseract-ocr-spa
```
AbinayaM02, line 31: @mnamysl : It's not ideal to do the tesseract installation in the workflow. We're installing only light transformations/filters by default. If this requirement can be specified in the requirements.txt, it should be automatically picked up. Let me find a better way to do this.

mnamysl:
Thank you for your feedback @AbinayaM02.

I moved the generate_test_cases method to test/helper.py so it can be reused by other transformations.

.github/workflows/nl-action.yml
mnamysl: @AbinayaM02 : Thank you for your feedback.

How about a mechanism similar to that used for the pip packages, e.g.:
`if [ -f packages.txt ]; then sudo apt-get update; xargs sudo apt-get install -y < packages.txt; fi` ?

AbinayaM02: The idea is whatever is supposed to be installed for your transformation should go only in your requirements.txt. Let me pull your code and see if this can be done at transformation level through requirements.txt. You can also try to find a way. We'll update once we find something.

mille-s:
Hi, and sorry for jumping in late, but is this tranformation related to https://github.com/GEM-benchmark/NL-Augmenter/pull/202?

mnamysl:
Hi @mille-s : thank you for your feedback!

> is this tranformation related to #202?

A short answer: yes, it is :-) but:
- In this transformation, each input sentence is first rendered as an image and the OCR engine (Tesseract 4) is used to recognize the text depicted in the rendered image to get the output text. The OCR engine is employed directly in this case.
- #202 uses lookup tables with frequent OCR errors (and misspelling) and induces them randomly into the input text. 

The advantage of the transformation proposed in this PR is that the error distribution should be more natural, though. The error are usually not uniformly distributed throughout the text. On the downside, its heavier than and not as flexible as #202 ;-)

.github/workflows/nl-action.yml
AbinayaM02: @mnamysl: Since tesseract can be installed only through apt-get install, it makes sense to do what you're already doing for installation. But, would it be too much if I request you to explore pytesseract (python wrapper of tesseract)? If pytesseract works as expected, then we can switch to that without making changes to the workflow. Let me know what you think.

mnamysl: @AbinayaM02 : Thank you for your feedback!

Unfortunatelly, [pytesseract](https://pypi.org/project/pytesseract/) also [requires](https://github.com/madmaze/pytesseract#installation) Tesseract to be installed in the OS.

I've also tested one open-source neural OCR model but it performed much worse, thus I stick to Tesseract.
Finding a comparable accurate neural model would be a nice extension but I would leave it for future work (another PR).

Generally, it is not bad to have a method that allows us to install system packages (in addition to purely python ones).
Other transformation might use, e.g., another tool or library like spell-checker engine, which is not available in the default OS image.

mille-s:
@mnamysl Ok I see, thanks for your answer! It makes sense to keep these two separated then, it's good to have both options.

.github/workflows/nl-action.yml
AbinayaM02: Oh yeah! Forgot that pytesseract will require Tesseract installation. 

By current design, installing NL-Augmenter will by default install only all the light transformations and filters. The heavy transformations and filters should be installed based on users need. If we make your transformation heavy, we can provide the instructions to install tesseract in the README. What do you suggest? Should we make your transformation as heavy since it uses Tesseract?

mnamysl: @AbinayaM02: Thank you for your input - I definitely agree that it would be the best option.

I reverted the changes in the workflow and flagged _ocr_perturbation_ as heavy in `test/mapper.py`.

I also updated README.md file with the instructions to install Tesseract on Ubuntu and included a link to the instructions for other OS.

AbinayaM02: Great! Thank you for making those changes.

test/helper.py
```diff
@@ -0,0 +1,42 @@
+
+import json, os
+from TestRunner import convert_to_snake_case
+
+
+def generate_test_cases(transformation, src_sentences=[], print_result=False, output_filename=""):
+    """
+    Sample code to demonstrate usage that can also assist in adding test cases.
+    """
```
AbinayaM02, line 9: Minor fix: It would be good to expand the docstring with param details.

transformations/ocr_perturbation/README.md
```diff
@@ -0,0 +1,83 @@
...
+python -m spacy download fr_core_web_sm
+python -m spacy download es_core_web_sm
+python -m spacy download de_core_web_sm
+```
+2. Install [Tesseract](https://github.com/tesseract-ocr/tesseract) (including dependencies and models). See the [installation instructions](https://tesseract-ocr.github.io/tessdoc/#compiling-and-installation) for your operating system. For example, On Debian or Ubuntu, you can simply use its developer tools:
+
+```bash
+apt-get install tesseract-ocr libtesseract-dev libleptonica-dev pkg-config \
+  tesseract-ocr-eng tesseract-ocr-osd tesseract-ocr-deu tesseract-ocr-fra tesseract-ocr-spa
+```
```
AbinayaM02, line 43: Silly question: Is this step enough for the transformation to work or should a user also pip install requirements.txt?

transformations/ocr_perturbation/transformation.py
```diff
@@ -0,0 +1,173 @@
...
+            image_generators = [self._get_image_generator(sent.text) for sent in doc.sents]
+
+            for idx in range(self.max_outputs):
+            
+                perturbed_sentence = ""
+                    
+                for sent, gen in zip(doc.sents, image_generators):
+
+                    img, lbl = gen.next()
+
```
AbinayaM02, line 108: Since lbl is not used, we don't need to store this variable. Replace with underscore.

AbinayaM02 (commented):
@mnamysl : Overall, the transformation looks good. Have added some minor suggestions. Please have a look.

transformations/ocr_perturbation/transformation.py
mnamysl: Done.

transformations/ocr_perturbation/README.md
mnamysl: Python dependencies should also be installed, thank you for spotting this inconsistency!

I updated the instructions in the README.md file accordingly.

test/helper.py
mnamysl: Fixed, thank you for your highlighting this point!

I also fixed a problem with the directory path of the result JSON file, which was set to `os.path.dirname(os.path.realpath(__file__))` inside `generate_test_cases()`. Because this method was moved to another directory, I changed the path to absolute, so the caller have to specify the exact location of the output file and it should  work fine now.

transformations/ocr_perturbation/README.md
AbinayaM02: Great!

AbinayaM02 (approved):
Thank you @mnamysl for making the changes. Here is my general observation,

- The transformation addresses perturbation using OCR engine [Tesseract].  
- New libraries are used and installation instructions are provided in the README.
- Test cases have been added.

I think the evaluation metric is not added to the leaderboard. If not, please add them after you pull the latest code. Approving the PR anyway :)

mnamysl:
@AbinayaM02: Thank you for your kind review and your suggestion that have helped improve this PR!

I added the evaluation results to the leaderboard and to the README.md file.

@kaustubhdhole: Please consider re-reviewing the PR again :-)

PR #212, opened 2021-08-19
My french_conjugation_transformation

Louanes1:
Faced this issue when using pre-commit : 

image

When I went to check the "pre-commit-config.yaml" file, it says that the repo for this hook is local. 

image

Since black, flake8 and isort hooks passed, I commited the code with following command : 

git commit -m "My french_conjugation_transformation" -n

### DIFF
--- a/transformations/french_conjugation_transformation/transformation.py
+++ b/transformations/french_conjugation_transformation/transformation.py
@@ -1,0 +1,385 @@
+import re
+
+import mlconjug
+import nltk
+import spacy
+from spacy.language import Language
+from spacy_lefff import LefffLemmatizer, POSTagger
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+default_conjugator = mlconjug.Conjugator(language="fr")
+
+nlp = spacy.load("fr_core_news_lg")
+nltk.download("wordnet")
+nltk.download("punkt")
+
+# definition of spacy pipeline
+
+
+@Language.factory("french_lemmatizer")
+def create_french_lemmatizer(nlp, name):
+    return LefffLemmatizer()
+
+
+@Language.factory("POSTagger")
+def create_POSTagger(nlp, name):
+    return POSTagger()
+
+
+nlp.add_pipe("POSTagger", name="pos")
+nlp.add_pipe("french_lemmatizer", name="lefff", after="pos")
+
+dicPronouns = {
+    "j'": "1s",
+    "J'": "1s",
+    "je": "1s",
+    "Je": "1s",
+    "tu": "2s",
+    "Tu": "2s",
+    "il": "3s",
+    "Il": "3s",
+    "elle": "3s",
+    "Elle": "3s",
+    "on": "3s",
+    "On": "3s",
+    "'on": "3s",
+    "nous": "1p",
+    "Nous": "1p",
+    "vous": "2p",
+    "Vous": "2p",
+    "ils": "3p",
+    "Ils": "3p",
+    "elles": "3p",
+    "Elles": "3p",
+}
+
+
+# function to update value of a dictionary based on its key
+def update_in_alist(alist, key, value):
+    return [(k, v) if (k != key) else (key, value) for (k, v) in alist]
+
+
+def update_in_alist_inplace(alist, key, value):
+    alist[:] = update_in_alist(alist, key, value)
+
+
+# function to change a dictionary key based on a given value
+def update_in_alist_key(alist, key, value):
+    return [(k, v) if (v != value) else (key, value) for (k, v) in alist]
+
+
+def update_in_alist_inplace_key(alist, key, value):
+    alist[:] = update_in_alist_key(alist, key, value)
+
+
+# Sometimes, the conjugation of a verb will modify the first letter of the verb (ex : avec le verbe ALLER : je vais (présent) devient j'irais (futur)) we therefore need to change the pronoun "Je" to "J'"
+def transform_pronoun_1s(sentence):
+    # regex to find if "je" or "Je" is followed by a vowel
+    regex_je = r"je (a|e|i|o|u|y|é|è|ê)|Je (a|e|i|o|u|y|é|è|ê)"
+    reg = re.search(regex_je, sentence)
+
+    if reg is None:
+        pass
+    else:
+        sentence = sentence.replace("je ", "j'")
+        sentence = sentence.replace("Je ", "J'")
+    # regex to find if "J'"" or "j'" is folowed by a consonant
+    regex_j = r"j'(z|r|t|p|q|s|d|f|g|h|j|k|l|m|w|x|c|v|b|n)|J'(z|r|t|p|q|s|d|f|g|h|j|k|l|m|w|x|c|v|b|n)"
+    regj = re.search(regex_j, sentence)
+    if regj is None:
+        pass
+    else:
+        sentence = sentence.replace("j'", "je ")
+        sentence = sentence.replace("J'", "Je ")
+    return sentence
+
+
+# sometimes, y is used to do a liason between the pronoun an a vowel, it must be deleted if next word starts by y or i
+def y_exception(sentence):
+    regex_y = r"j'y (i|y)|J'y (i|y|)"
+    reg = re.search(regex_y, sentence)
+
+    if reg is None:
+        pass
+    else:
+        sentence = sentence.replace("j'y ", "j'")
+        sentence = sentence.replace("J'y ", "J'")
+    return sentence
+
+
+# the negation is expressed with n' in font of a vowel and ne in front of a consonant
+def n_exception(sentence):
+    regex_n = r"n'(z|r|t|p|q|s|d|f|g|h|j|k|l|m|w|x|c|v|b|n)|N'(z|r|t|p|q|s|d|f|g|h|j|k|l|m|w|x|c|v|b|n)"
+    reg = re.search(regex_n, sentence)
+    if reg is None:
+        pass
+    else:
+        sentence = sentence.replace("n'", "ne ")
+        sentence = sentence.replace("N'", "Ne ")
+
+    regex_ne = r"ne (a|e|i|o|u|y|é|è|ê)|Ne (a|e|i|o|u|y|é|è|ê)"
+    reg = re.search(regex_ne, sentence)
+    if reg is None:
+        pass
+    else:
+        sentence = sentence.replace("ne ", "n'")
+        sentence = sentence.replace("Ne ", "N'")
+    return sentence
+
+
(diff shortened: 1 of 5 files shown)


AbinayaM02:
Hi @Louanes1 : Thanks for your contribution. The pre-commit hook failed because your test case didn't pass. Did you try running the pytest for your transformation locally?

Louanes1:
> Hi @Louanes1 : Thanks for your contribution. The pre-commit hook failed because your test case didn't pass. Did you try running the pytest for your transformation locally?

Thanks for the reply :) Yes when I run the pytest , the test passes : 

image

I've added a new requirements.txt file within my transformation folder. Shoud I use the main requirements.txt file ? (the one in the transformations folder)

AbinayaM02:
> Oh, I think it's because I added a direct link to download the "fr_core_news_lg" in my requirements file, the build in "Checks" fails because of some characters in the link. I'll try to add it differently

Yes, you're right. The build is failing because of that.

Louanes1:
> > Oh, I think it's because I added a direct link to download the "fr_core_news_lg" in my requirements file, the build in "Checks" fails because of some characters in the link. I'll try to add it differently
> 
> Yes, you're right. The build is failing because of that.
> 
> Since your test case passed when you ran it separately, ideally pre-commit test hook shouldn't fail. Try running the pre-commit and see if it throws any specific message.
> 
> [Edit] Are you using ubuntu or windows while committing the code?

I am using windows, and when I run the pre-commit, the test hook still fails :/ 
After removing the fr_core_news_lg, the build throws another error : 
image

It looks like there is a conflict with the nltk version I use in my requirements. I removed the version so pip will attempt to resolve the dependecy conflict.

Louanes1:
Hi @AbinayaM02, 
The dependencies installed correctly ! No more conflict :)
However, I use a spacy package called "fr_core_news_lg" and this one needs to be download with following : python -m spacy download fr_core_news_lg.

The installation is supposed to be similar to the en_core_web_sm  found in initialize.py. I've added a link to download it in my requirements file (its commented otherwise special characters will fail the build : https://github.com/explosion/spacy-models/releases/download/fr_core_news_lg-3.0.0/fr_core_news_lg-3.0.0-py3-none-any.whl)

I have also tried to donwload the fr_core_news_lg, add it to my folder and add a link to that file in my requirements, but obviously it's too large, git won't let me push. 

Do you have any idea on how I am suppose to proceed ? Thanks

AbinayaM02:
> Do you have any idea on how I am suppose to proceed ? Thanks

Try adding your model in the below format in the requirement.txt and uncomment it. (Hopefully, it should work!)
fr-core-news-lg @  https://github.com/explosion/spacy-models/releases/download/fr_core_news_lg-3.0.0/fr_core_news_lg-3.0.0-py3-none-any.whl

richplant:
Seems inefficient to load a second Spacy model that might never get used for more than this single transformation in the initialize script. You should probably import it as a module and call the .load() function directly on it in your script.

You also need to add the appropriate language tags and keywords to your class and a robustness evaluation to the readme (check the evaluate.py script in the main dir).

mille-s:
Hi, nice transformation! Do we have an idea of the accuracy of the substitution? Does it fail sometimes, and if so how often?

Louanes1:
> Hi, nice transformation! Do we have an idea of the accuracy of the substitution? Does it fail sometimes, and if so how often?

Hello @mille-s thank you ! The conjugation of the verbs is quite robust since it is relying on the mlconjug library and their model is trained on the different verb groups

 In french, we have 3 different group of verbs

-  1st group where verbs end with "er" like "manger, écouter"
-  2nd group where they end with "ir" like "finir, courir"
-  3rd group where they end with "re" like "boire, croire"

The conjugation of each group is different whatever the tense.

The french model behind mlconjug is trained to predict the conjugation based on these different groups, so that even when a verb do not exist like "facebooker", it will still consider it as a verb of 1st group and conjugate it accordingly.

However in order to conjugate a verb we first need to transform it into its "indicative" form (he ate --> to **eat**) and send it to the mlconjug function. And we do that with lemmatization, so I guess that if we can not get the lemma of the conjugated verb from the original sentence, we won't be able to conjugate it to a different tense. The spacy lemmatizer is quite good though so I didn't encounter verbs that could trigger these issue. 

The common issues I encountered are more linked to the "pronoun" it should be conjugated to (because the conjugation also differs based on the pronoun used, and it is an entry of the mlconjug function). Right now it can handle sentences where the subject is a pronoun defined in our dictionnary. But if the subject is a group of words ( **the parents**instead of **they**) it won't be able to detect the pronoun, therefore won't know to which person the verb is supposed to be conjugated.

mille-s:
@Louanes1 ok thanks for the answer! Note that there should be some kind of filter applied on the candidate input sentences that returns only sentences with one main verb and a subject pronoun before applying your transformation. Do you have such a filter at hand?

Louanes1:
@mille-s  the input sentence can contain more than one verb and it will be conjugated as long as each verb is preceded by a subject pronoun. If we do not detect a new pronoun for the next verb, it assumes that this verb should also be conjugated to the same pronoun. 

I've changed one of my example test to illustrate this (the sentence contains 4 conjugated verbs and they all change tense according to the pronoun written before them)

I guess the next step should be to detect whenever a group of word represent a specific pronoun and add it to the dictionary of pronouns. 

Unfortunately I don't have a filter at hand that does what you're looking for.

mille-s:
@Louanes1 ok if it's not limited to sentences with one verb then probably no filter is needed, the tranformation will be rather frequent and affect many sentences in a dataset (and thus have a potential impact on the evaluation).

Re "But if the subject is a group of words ( the parents instead of they) it won't be able to detect the pronoun, therefore won't know to which person the verb is supposed to be conjugated.": isn't is safe to assume that if there is no pronoun a verb is in third person? Then it's a matter of finding out if the verb is singular or plural, which should be doable. This could be a future improvement to reach more verbs.

transformations/french_conjugation_transformation/transformation.py
```diff
@@ -0,0 +1,390 @@
...
+        "lexical",
+        "model-based",
+        "highly-meaning-preserving",
+        "high-coverage",
+    ]
+
+    # The transormation tense is specified here
+    def __init__(self, tense):
+        super().__init__()
+        assert tense in ["Futur", "Imparfait"]
```
tuetschek, line 382: Does this correspond to the documentation? You state more tenses there than here?

tuetschek:
> Re "But if the subject is a group of words ( the parents instead of they) it won't be able to detect the pronoun, therefore won't know to which person the verb is supposed to be conjugated.": isn't is safe to assume that if there is no pronoun a verb is in third person? Then it's a matter of finding out if the verb is singular or plural, which should be doable. This could be a future improvement to reach more verbs.
I agree with @mille-s . Otherwise I like the transformation :-) !

mille-s (approved):

transformations/french_conjugation_transformation/transformation.py
```diff
@@ -0,0 +1,390 @@
...
+from spacy_lefff import LefffLemmatizer, POSTagger
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+default_conjugator = mlconjug.Conjugator(language="fr")
+
+nltk.download("wordnet")
+nltk.download("punkt")
+nlp = spacy.load("fr_core_news_md")
```
kaustubhdhole, line 16: It is better to move this into the constructor for consistency.

transformations/french_conjugation_transformation/transformation.py
Louanes1: The tenses specified in the read me can also be passed since they are supported by the mlcnonjug library. I didn't pass them in my function, I had issues with utf-8 encoding when I specify a tense containing an accent (looks like the test json file isn't opened with the right encoding)

Louanes1:
Hey @tuetschek , @mille-s 
> Re  isn't is safe to assume that if there is no pronoun a verb is in third person? Then it's a matter of finding out if the verb is singular or plural, which should be doable. This could be a future improvement to reach more verbs.

Actually, this makes sense for verbs in english. In french the spelling is different when the verb is conjugated with each plural pronoun : 

I ate --> Je mangeai
You ate --> Tu mangeas
He ate --> Il mangea
We ate --> Nous mangâmes
You ate --> Vous mangeâtes
They ate --> Ils mangèrent

There is quite a different ending depending on the pronoun used, that is why we cannot assume one of them in case we didn't find any.
I am  trying to build a classification model where I pass a conjugated verb and I expect the pronoun it is conjugated to, to be returned. 
I've gathered around 3000 verbs; I conjugate them to past, futur and present, then I map each one them with its pronoun. (My dataset have 2 columns : all conjugated verbs, and pronoun) I am still working on it, trying to get the prediction, I will let you know as soon as I find something that could help us solve this issue.
By the way, which classification algorithm, you guys think I should use for this kind of task ?


Meanwhile,  I believe it is better to conjugate verbs to the latest detected pronoun, so that it can handle cases where a subject does more than 1 action, many verbs should be conjugated to that pronoun.

AbinayaM02:
Hi @Louanes1: Please add your transformation name to the test/mapper.py in the right dictionary for the pytest to pick up your test.json. By default, we're testing only light transformations and filters.

mille-s:
@Louanes1 : what I meant was that if there is no pronoun before a verb, this verb will very likely be in third person since it's almost mandatory to have a first or second person pronoun to have a verb conjugated in first or second person (except for the imperative mood, for which the verb is used with no pronouns, but this mood is overall quite unfrequent). So to get the right ending of a verb when there is no pronoun, it boils down to finding the number (third sigular or third plural), which I think can be derived from the original verb form with simple regex in many cases (this needs to be checked with more care). In any case this can be added later as an improvement, no need to do it now.

AbinayaM02:
Hi @Louanes1: Please do not rebase the branch. Follow the below steps to add your changes only,

1. Fetch and merge the main branch of your repository which you have already done on the UI.
image
2. Checkout the main branch and pull the main branch on to your local repository.
```python
git checkout main
git pull origin main
```
3. Switch to your branch where you're making changes and pull the main branch into that.
```python
git checkout french_conjugation_transformation
git pull origin main
```
4. If there are any conflicts, resolve them (mostly there won't be any except for the test/mapper.py where the latest file will have your changes). Add your changes pertaining to french conjugation transformation and commit.
5. Push your changes to the upstream repository.
```python
git push origin french_conjugation_transformation
```
6. Your PR will be automatically updated. 

You can either try to fix your current PR or open a new clean PR with only your changes.

Louanes1:
I've ended up creating a new PR : https://github.com/GEM-benchmark/NL-Augmenter/pull/308
@AbinayaM02 @mille-s

PR #213, opened 2021-08-20
Common misspellings

csinva:
Adds transformation which corrects spellings of commonly misspelled words.


transformations/correct_common_misspellings/transformation.py
```diff
@@ -0,0 +1,4117 @@
...
+        doc = self.nlp(sentence)
+        spaces = [True if tok.whitespace_ else False for tok in doc]
+
+        # replace misspelled words
+        perturbed_text = []
+        for index, token in enumerate(doc):
+            s = token.text
+            if s in self.COMMON_MISSPELLINGS_DICT:
+                s = self.COMMON_MISSPELLINGS_DICT[s]
+            perturbed_text.append(s)
```
RoyRin, line 32: doesn't matter, but:
`perturbed_text = [ self.COMMON_MISSPELLINGS_DICT[s.text] for s in doc if s.text in self.COMMON_MISSPELLINGS_DICT]`

Is likely easier to parse

RoyRin: I'd actually just consider replacing the whole function with

```
perturbed_text = [ self.COMMON_MISSPELLINGS_DICT.get(token.text, token.text) + " " for token in doc 
                              if token.whitespace_ 
                              else self.COMMON_MISSPELLINGS_DICT.get(token.text, token.text) ]                      
```

transformations/correct_common_misspellings/transformation.py
```diff
@@ -0,0 +1,4117 @@
...
+d = np.loadtxt("dict.txt", delimiter="->", dtype=str)
+
+# prune anything with commas
+unique_idxs = np.array([len(x.split()) == 1 for x in d[:, 1]])
+d = d[unique_idxs]
+
+# print as dict
+d = {d[i, 0]: d[i, 1] for i in range(d.shape[0])}
+print(d)
+"""
```
RoyRin, line 4117: I would remove this commented out code

csinva: I'm keeping this since it shows some of the preprocessing that was done to clean the data

transformations/correct_common_misspellings/transformation.py
csinva: changed

transformations/correct_common_misspellings/transformation.py
```diff
@@ -0,0 +1,4112 @@
...
+        perturbed_text = [
+            self.COMMON_MISSPELLINGS_DICT.get(token.text, token.text) + " "
+            if token.whitespace_
+            else self.COMMON_MISSPELLINGS_DICT.get(token.text, token.text)
+            for token in doc
+        ]
+        return ["".join(perturbed_text)]
+
+
+def get_common_misspellings_dict():
```
sebastianGehrmann, line 40: Can this be a separate json file or so that gets loaded here?

sebastianGehrmann (changes requested):
Looking good, thank you! My main comment is to please refactor the dictionary into a separate file so we don't have a 4k line python file.

transformations/correct_common_misspellings/transformation.py
csinva: Thanks good idea -- I just moved the dictionary into a separate json file which gets loaded!

marco-digio:
Hi @csinva 
Thanks for this contribution! 
I like this transformation.

Here's my general review:
Correctness: No checks (1 workflow awaiting approval)
Applicable Tasks & Keywords: I think you have chosen the relevant tasks and keywords.
Specificity: The change is specific to texts with misspellings.
Adding New Libraries: No new libraries appear to be added.
Data and code source: Data and code (with GNU GENERAL PUBLIC LICENSE) sources are referenced in README.
Evaluating Robustness: Robustness evaluation has been performed.

transformations/correct_common_misspellings/README.md
```diff
@@ -0,0 +1,32 @@
...
+It works only for English and operates only at the word-level (i.e. it does not incorporate context around individual words to gauge their spelling).
+
+## Robustness Evaluation
+
+Here is the performance of the model aychang/roberta-base-imdb on the test[:20%] split of the imdb dataset.
+The accuracy on this subset which has 1000 examples = 96.0
+
+1000 examples generated from 1000 original examples, with 35 successfully transformed and 965 unchanged (0.035 perturb rate).
+Here is the performance of the model on the transformed set:
+The accuracy on this subset which has 1000 examples = 96.0
```
kaustubhdhole, line 32: Okay, one last thing - would you be interested to add [robustness evaluation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation) there too.

csinva: just added!

kaustubhdhole (approved):

PR #215, opened 2021-08-21
Implement speech disfluency perturbation

ikb-a:
Simple perturbation to insert English filler words (e.g., "er", "uh", "um", etc...)

### DIFF
--- a/transformations/speech_disfluency_perturbation/README.md
+++ b/transformations/speech_disfluency_perturbation/README.md
@@ -34,7 +34,15 @@ user-provided filler words it will not work on languages without whitespace.
 Also note, inputs that are a single token (where tokens are defined
 merely as being white-space separated), then the output may be ungrammatical due
 to the creation of a new sentence, or due to violating capitalization. For
-example, "Yes." may become "uh Yes." or "Yes. uh". Also worth noting is that the
+example, "Yes." may become "uh Yes." or "Yes. uh".
+
+More generally, for inputs of more than a single token, the disfluency will only
+be inserted between input tokens. However, this transformation does not consider
+capitalization. Therefore, there may be violations to capitalization depending
+on the inputs (e.g., "Yes. The tower is that way" may become "Yes. uh The tower
+is that way").
+
+Also worth noting is that the
 number of possible outputs increases exponentially with the
 number of tokens (i.e., the number of places where disfluencies may be
 inserted). However, short inputs have a limited number of possible transformations
--- a/transformations/speech_disfluency_perturbation/README.md
+++ b/transformations/speech_disfluency_perturbation/README.md
@@ -31,3 +31,11 @@ on other languages. Finally, the current implementation
 inserts disfluencies between words denoted by whitespace, therefore even with
 user-provided filler words it will not work on languages without whitespace.
 
+Also note, inputs that are a single token (where tokens are defined
+merely as being white-space separated), then the output may be ungrammatical due
+to the creation of a new sentence, or due to violating capitalization. For
+example, "Yes." may become "uh Yes." or "Yes. uh". Also worth noting is that the
+number of possible outputs increases exponentially with the
+number of tokens (i.e., the number of places where disfluencies may be
+inserted). However, short inputs have a limited number of possible transformations
+(e.g., inputs of 2 tokens only have 1 possible output).
--- a/transformations/speech_disfluency_perturbation/README.md
+++ b/transformations/speech_disfluency_perturbation/README.md
@@ -1,0 +1,33 @@
+# Speech Disfluency Perturbation 🦎  + ⌨️ → 🐍
+This perturbation randomly inserts speech disfluencies, specifically English filler words.
+
+Author name: Ian Berlot-Attwell
+Author Affiliation: University of Toronto
+
+## What type of a transformation is this?
+This transformation acts like a perturbation to test robustness,
+particularly with respect to spoken language. With independent probability `insert_prob`, a
+speech disfluency (specifically, a randomly selected filler word)
+is inserted between words. The default disfluencies are
+"um", "uh", "erm", "ah", and "er"; they can be alternatively specified
+by the user using `filler_words`. At least one filler word is always
+inserted by the transformation.
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all tasks which have a sentence/paragraph/document as input like text classification,
+text generation, etc.
+
+Performing the default evaluation (i.e., `python evaluate.py -t SpeechDisfluencyPerturbation`),
+ we find a 2% decrease in performance on the TEXT_CLASSIFICATION task.
+Specifically, a degradation from 96.0% to 94.0% for the model `aychang/roberta-base-imdb`
+on the imdb dataset.
+
+## What are the limitations of this transformation?
+The transformation's outputs are simple, and incapable of
+ generating linguistically diverse text. Currently, the transformation
+only contains English filler words, therefore the user must provide
+filler words for other languages should they wish to use this tranformation
+on other languages. Finally, the current implementation
+inserts disfluencies between words denoted by whitespace, therefore even with
+user-provided filler words it will not work on languages without whitespace.
+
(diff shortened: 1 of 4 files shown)


ikb-a:
Updated to include the transformation keywords: ['lexical', 'noise', 'rule-based', 'highly-meaning-preserving', 'high-precision', 'high-coverage', 'high-generations',  'social-reasoning']

I assume lexical is appropriate as the changes are being made at the word-level (even if it's just additions and not replacements). I have tentatively included 'social-reasoning' as the skill that would be required for recognizing speech disfluencies.

marco-digio:
Hi @ikb-a 
Thanks for this contribution! 
I think that this transformation is very simple but still useful in selected experiments.

Here's my general review:
Correctness: No checks (1 workflow awaiting approval)
Applicable Tasks & Keywords: I think you have chosen the relevant tasks and keywords.
Specificity: The change is general but its application could be few. 
Adding New Libraries: No new libraries appear to be added.
Description: The README is clear.
Data and code source: Data and code source missing.
Evaluating Robustness: 2% decrease in performance on the TEXT_CLASSIFICATION task, from 96.0% to 94.0% for the model aychang/roberta-base-imdb on the imdb dataset.

ikb-a:
Hi @marco-digio! Glad to hear the transformation can be of use. As to data & code source, there is no data and the code is my own. Would you like me to add a sentence in the README to that effect?

I can also confirm that there's no new libraries. I think that's everything, but if I've missed something that needs to be adjusted, then please do let me know. Thanks!

transformations/speech_disfluency_perturbation/test.json
```diff
@@ -0,0 +1,60 @@
+{
+  "type": "speech_disfluency_perturbation",
+  "test_cases": [
```
cascadianblue, line 3: I think there should be some unit tests with non-default `filler_words`

transformations/speech_disfluency_perturbation/transformation.py
```diff
@@ -0,0 +1,87 @@
...
+        random.seed(self.seed)
+        perturbed_texts = []
+
+        for _ in range(self.max_outputs):
+            tokens = sentence.split()
+            choices = []
+            while not choices:
+                for i in range(1, len(tokens) - 1):
+                    if random.random() < self.insert_prob:
+                        choices.append(i)
```
cascadianblue, line 50: This goes into an infinite loop if `len(tokens) <= 2`

cascadianblue (changes requested):

transformations/speech_disfluency_perturbation/test.json
ikb-a: Good idea, I've added three test-cases

transformations/speech_disfluency_perturbation/transformation.py
ikb-a: Good catch! I've corrected the infinite loop, adjusted the README to note that the results for 1-token inputs might be a bit odd, and added corresponding tests.

ikb-a:
@cascadianblue @marco-digio  I believe I have address all of the requested changes (and to the best of my understanding, no changes are required in the README as the code & data are my own, without any previous work to the best of my knowledge). Is anything further required?

ikb-a:
> 
> 
> Hi, thanks for the submission! Please assess to what extent this one is different from #194. Could we merge these two?

Hi @mille-s, good question!

I didn't notice #194 when I made my transformation, but nonetheless, the intended use of the default behaviour is different. Mine is focused on text transcribed from audio, hence the inserted fillers are only sounds -- "uh", "err", etc... In comparison, #194's default behaviour is to introduce general linguistic noise by inserting from a much broader range of filler words and phrases, beyond just sounds (specifically, they have some lists of non-meaning preserving fillers, and their choice of _meaning-preserving_ fillters is ["uhm", "umm", "ahh", "err", "actually", "obviously", "naturally", "like", "you know"]). As a result, my default behaviour is safer with respect to preserving meaning -- for example, under the meaning preserving settings, #194 might convert "Jim knows Bob's hair is blonde" to "Jim knows you know Bob's hair is blonde", "Jim knows Bob's hair is naturally blonde", "Jim knows Bob's hair is actually blonde" (implying someone thinks Jim thinks otherwise), or some combinations thereof (e.g., "Jim knows you know Bob's hair is naturally blonde"). In short, the intended use of these transformations is a bit different, and mine's safer with respect to preserving meaning.

Having said all that, the underlying mechanism is exactly the same (randomly insert new token between existing words). #194 uses a fixed list, mine allows for the user to provide a list of fillers. If we want to merge these, it should just be a simple matter of allowing the user to select between our pre-defined lists of fillers, or to provide their own. Defining a constant for each default list and allowing the user to pass in any sum should do the trick. 

Overall, I'd say that based on the default intended purpose, and default degree of meaning-preservation, the transformations merit separate existence. However, since the actual algorithm is pretty much the same, if you disagree then it should be easy to merge these.

mille-s:
@ikb-a : thanks for the clear answer! I also think both transformations should be kept. I like the idea of allowing the users to choose their list or create their own. @kaustubhdhole what's best in this case fro now, keeping two separate transformations or bringing them together?

transformations/speech_disfluency_perturbation/transformation.py
```diff
@@ -0,0 +1,116 @@
...
+
+            choices = choices[::-1]
+            for i in choices:
+                tokens.insert(i, random.choice(self.filler_words))
+
+            if short_input:
+                tokens = tokens[1:-1]
+
+            perturbed_texts.append(" ".join(tokens))
```
cascadianblue, line 66: You don't handle capitalization changes due to introduced words (which is nontrivial; you'd need to figure out what is a proper noun). I don't think it's that big of a deal, since this transformation is more oriented towards transcribed speech, but you should at least mention that in the "Limitations" section.

ikb-a: I've updated the README to reflect the limitation. I can see problems with multi-sentence input (e.g., "Hello. How are you?" -> "Hello. uh How are you?"), but I don't quite follow why proper nouns are relevant ("His name is Bob Barker now" -> "His name is uh Bob Barker now"/"His name is Bob uh Barker now"/"His name is Bob Barker uh now")

cascadianblue: An example of the phenomenon I was thinking of: "I see why you like it here. Ahmed's house is right around the corner. The books are under the table" -> "I see why you like it here. Um **Ahmed's** house is right around the corner. Uh **The** books are under the table.". Notice that "Ahmed" should stay capitalized, while "The" should not.

Sorry for not describing it well enough/providing an example earlier!

transformations/speech_disfluency_perturbation/README.md
```diff
@@ -0,0 +1,49 @@
...
+Also note, inputs that are a single token (where tokens are defined
+merely as being white-space separated), then the output may be ungrammatical due
+to the creation of a new sentence, or due to violating capitalization. For
+example, "Yes." may become "uh Yes." or "Yes. uh".
+
+More generally, for inputs of more than a single token, the disfluency will only
+be inserted between input tokens. However, this transformation does not consider
+capitalization. Therefore, there may be violations to capitalization depending
+on the inputs (e.g., "Yes. The tower is that way" may become "Yes. uh The tower
+is that way").
```
cascadianblue, line 43: Looks good!

cascadianblue (approved):

transformations/speech_disfluency_perturbation/transformation.py
ikb-a: No worries; all makes sense now

kaustubhdhole:
Thank you very much @ikb-a for your work. I agree with @mille-s to keep this PR separate as this one is a specific PR. However, I also feel the scope of the filler words in this PR is very limited. I strongly suggest you to expand the scope of this PR (possibly to more languages or even more tasks?). Nonetheless, it is good to see a 2% drop in performance. I would suggest you to do the same evaluation for the [leaderboard](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation) too. I am merging this one but I hope you would address these changes in a follow-up PR. Besides, I am commenting a few things and removing "social-reasoning" as a keyword from your PR as I think it might not fit well here.

Thanks again for all your changes.

kaustubhdhole (approved):

PR #216, opened 2021-08-21
Spanish gender swap

juand-r:
Swaps the gender of pronouns and gender-bearing nouns from masculine to femenine in Spanish sentences.


juand-r:
The test is failing because it doesn't have the "es_core_news_lg" spacy model.  Should I put this in `initialize.py`?  Or would that not be ideal (because it is only required for the Spanish transformations).

juand-r:
> The test is failing because it doesn't have the "es_core_news_lg" spacy model. Should I put this in `initialize.py`? Or would that not be ideal (because it is only required for the Spanish transformations).


For now I added es_core_news_lg to the requirements.txt file, which should work with pip>=19.1.

mille-s:
This is a very challenging transformation since changing one (pro)noun means adapting some other elements of the sentence too. I think we absolutely need a way to assess the quality of the transformed outputs, because we don't want to create dirty data that would not be usable. On the long term, this transformation could be very powerful (when it is possible to control the amount of (pro)nouns changed and to what they are changed that all agreements are changed correctly too, and that the resulting sentence still makes sense). @juand-r listed some important limitations in his documentation, it is very helpful to have an idea of what will need to be solved. I recommend to accept this transformation so we can start working on it, but I don't expect the resulting data to be fully usable soon; however, we cannot exclude that intermediate versions of the data will already allow for running interesting experiments.

juand-r:
> This is a very challenging transformation since changing one (pro)noun means adapting some other elements of the sentence too. I think we absolutely need a way to assess the quality of the transformed outputs, because we don't want to create dirty data that would not be usable. On the long term, this transformation could be very powerful (when it is possible to control the amount of (pro)nouns changed and to what they are changed that all agreements are changed correctly too, and that the resulting sentence still makes sense). @juand-r listed some important limitations in his documentation, it is very helpful to have an idea of what will need to be solved. I recommend to accept this transformation so we can start working on it, but I don't expect the resulting data to be fully usable soon; however, we cannot exclude that intermediate versions of the data will already allow for running interesting experiments.

Thanks!  I think it is a great idea to extend this (with a little more work) to make it more flexible, so that one could change some (pro)nouns and not others; doing this might cause incorrect changes sometimes (for text to text tasks, such as applying the transformation to both a source and a reference summary), but could be useful for other things like text classification or NER. It wouldn't be hard to also add the reverse direction female -> male, but there would be a little more work dealing with additional complications which would arise (mentioned in the README).

I'll see if I can find some data to help evaluate how well the transformation works.  Also, if there are any experts in Spanish morphosyntax with pointers to grammatical gender phenomena that are not covered here, that would be super helpful. There may be things I haven't considered -- both in terms of syntax (e.g., various sorts of linking verbs, of which I have not found a definitive list), and edge cases like *polícia*. *El polícia* refers to a policeman, while *la polícia* refers to a police force; *un polícia* could be male or female, but to explicitly make it female, one would need to write *la mujer polícia*. The only other examples I can can think of which similar to this are epicene animals such as *búho* (the only to indicate a female owl is *el búho hembra*), but there may be other examples.

Another thing -- some of the gender changes are region-specific (and also evolving over time). For example, here is an interesting article on how *bebé* (baby) is epicene in Spain, but usually not in Latin America (and its usage in Spain is becoming more similar to the Latin American usage): https://cvc.cervantes.es/el_rinconete/anteriores/octubre_15/15102015_01.htm
Unfortunately, unless we can find a list of such region-specific cases, it would be very time consuming to incorporate this distinction into the transformation.

msobrevillac:
> The test is failing because it doesn't have the "es_core_news_lg" spacy model. Should I put this in `initialize.py`? Or would that not be ideal (because it is only required for the Spanish transformations).

I think yes.

msobrevillac:
This transformation is interesting and seems to be useful, mainly for bias mitigation. Besides, the code is well-documented. I recommend to approve this transformation.

msobrevillac (approved):

juand-r:
> This transformation is interesting and seems to be useful, mainly for bias mitigation. Besides, the code is well-documented. I recommend to approve this transformation.

Thanks!

mille-s (approved):

PR #217, opened 2021-08-23
Commit of augmentation files for word stemmer

davidwanguoft:
Title.

Adding files for a word stemmer augmentation. See README for more details.

### DIFF
--- a/transformations/word_stemmer/test.json
+++ b/transformations/word_stemmer/test.json
@@ -1,0 +1,50 @@
+{
+    "type": "simple_word_stemmer",
+    "test_cases": [
+      {
+        "class": "SimpleWordStemmer",
+        "inputs": {
+          "sentence": "Where did he learn to dance like that?"
+        },
+        "outputs": [{
+           "sentence": "where did he learn to danc like that?"
+        }]
+      },
+      {
+        "class": "SimpleWordStemmer",
+        "inputs": {
+          "sentence": "His eyes were dancing with humor."
+        },
+        "outputs": [{
+          "sentence": "hi eye were danc with humor."
+        }]
+      },
+      {
+        "class": "SimpleWordStemmer",
+        "inputs": {
+          "sentence": "She shook her head and danced away"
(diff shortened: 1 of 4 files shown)


mille-s:
Hi, thanks for the submission! Could you please describe a scenario in which we could use the stemming to better test models? I'm not sure how the created data could be used for training or testing purposes in the context of NL-Augmenter.

kaustubhdhole:
I agree with @mille-s though I like the idea of using stemmed words. It might be useful for your case to point out some related work or motivation about stemming eg. is the meaning fully preserved at least in the eyes of the human? Here are a few comments:

transformations/word_stemmer/transformation.py
```diff
@@ -0,0 +1,75 @@
...
+    
+    Attributes:
+
+    """
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+    ]
+    languages = ["en"]
+    heavy = True
```
kaustubhdhole, line 42: Do you need heavy to be true here? Also, please add the keywords

transformations/word_stemmer/test.json
```diff
@@ -0,0 +1,50 @@
...
+          "sentence": "hi eye were danc with humor."
+        }]
+      },
+      {
+        "class": "SimpleWordStemmer",
+        "inputs": {
+          "sentence": "She shook her head and danced away"
+        },
+        "outputs": [{
+          "sentence": "she shook her head and danc away"
```
kaustubhdhole, line 28: If you could maintain the casing in the output, that might help make the transformation specific?

kaustubhdhole:
Thanks @davidwanguoft, this transformation looks good for NL Augmenter. Please address the above comments and I am happy to merge. If possible, please perform the robustness evaluation too.

kaustubhdhole:
@davidwanguoft ping!

PR #218, opened 2021-08-24
Added use_acronyms

Sotwi:
Change a group of words to their corresponding acronym


transformations/use_acronyms/transformation.py
```diff
@@ -0,0 +1,72 @@
...
+Use Acronyms.
+    Change groups of words for their equivalent acronym.
+    The acronyms were scraped from Wikipedia's List of Acronyms and naively filtered.
+"""
+
+
+def transformation(sentence, acronyms):
+    new_sentence = sentence
+    for key in acronyms.keys():
+        if key in new_sentence:
```
ashish3586, line 14: One suggestion, if we add "key.lower() in new_sentence.lower()" in the if statement, we can also cover cases like "massachusetts institute of technology", "Massachusetts institute of technology" and "massachusetts institute of Technology" etc.

Sotwi: I was weary of dealing with casing because of how it might affect the meaning between input and output. For example, a sentence like: "They visited the new York stadium" will end up becoming "They visited the NY stadium" which is not the same. By sticking to the _standard_ capitalization the chance of these miss changes diminishes.

I guess it is a matter of preference as well as how relevant it is to keep the meaning of the original sentence for the specific task for which the data is being augmented.

In any case, I added support for lowercasing as a class parameter.

transformations/use_acronyms/__init__.py
```diff
@@ -0,0 +1 @@
+from .transformation import *
```
ahoimarie, line 1: This seems unnecessary.

transformations/use_acronyms/transformation.py
```diff
@@ -0,0 +1,94 @@
...
+        "low-coverage",
+        "low-generations",
+    ]
+
+    def __init__(
+        self,
+        seed=0,
+        max_outputs=1,
+        lowercase=False,
+        acronyms_file="transformations/use_acronyms/acronyms.tsv",
```
ahoimarie, line 48: This could be improved to make it more robust and less dependent on your working directory, e.g. by using a file path instead.  

E.g.:
At the top `import os`
and your input `acronyms_file="acronyms.tsv"`
And then accordingly:
       ` acronyms_path=os.path.join(
                os.path.dirname(os.path.abspath(__file__)),
                acronyms_file)`

And then later:
`with open(acronyms_path, "r", encoding=encoding) as file:`

ahoimarie (approved):
Hello @Sotwi,
I was assigned as one of your reviewers. Thanks for this contribution! I think this could come in handy in writing documents (such as theses) where you can input a list of acronyms and it then uses the acronyms throughout. 

I have left some minor comments in the code, please take a look for some potential suggestions.
Here's my general review:
**Correctness**: All checks have passed. 
**Interface**: I think you have chosen the correct interface.
**Applicable Tasks & Keywords**: I think you have chosen the relevant task- and keywords. However, you specified the keyword `tokenizer-required` but I did not see the use of one? 
**Specificity**: The change is specific to text classification and generation (if an input text contains a key from the accompanying acronyms dictionary).
**Novelty**: This transformation is not yet implemented in NL-Augmenter.
**Adding New Libraries**: No new libraries appear to be added, and there is no requirements.txt file.
**Description**: The README is clear and explains what the transformation is aiming to do.
**Data and code source**: Data source missing.
**Evaluating Robustness**: One robustness evaluation has been performed.

transformations/use_acronyms/transformation.py
Sotwi: I wanted to make it possible for the user to use their own list of acronyms (that way it can work on a very specific set f acronyms or with data in other languages). But I will apply this suggestion to make sure that when loading the default file it will be done properly.

ahoimarie: Yes, it sounds reasonable. I am happy for you to leave it as is.

Sotwi: Perfect, now the default acronyms file is being loaded using the file path as suggested while still allowing for user defined pths for use defined files.

I also added a bit of code to make sure the dictionary keys are ordered from larger to shorter, to prevent smaller Acronyms to override larger ones (ie. prevent "New York University" from being turned into "NY University" instead of "NYU").

transformations/use_acronyms/transformation.py
```diff
@@ -0,0 +1,102 @@
...
+            acronyms[k] = temp_acronyms[k]
+        self.acronyms = acronyms
+
+    def generate(self, sentence: str):
+        return [transformation(sentence, self.lowercase, self.acronyms)]
+
+
+# """
+# Sample code to demonstrate usage. Can also assist in adding test cases.
+# You don't need to keep this code in your transformation.
```
kaustubhdhole, line 75: You might want to remove/comment this.

transformations/use_acronyms/transformation.py
```diff
@@ -0,0 +1,102 @@
...
+                new_sentence[:key_index]
+                + acronyms[key]
+                + new_sentence[key_end:]
+            )
+
+    return new_sentence
+
+
+class UseAcronyms(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
```
ashish3586, line 33: you can add tasks PARAPHRASE_DETECTION, TEXTUAL_ENTAILMENT, SENTIMENT_ANALYSIS

mille-s (approved):

kaustubhdhole:
Hi @Sotwi, you don't need to comment the test section in your case. There was an error in a different test case which Abhinaya fixed here - https://github.com/GEM-benchmark/NL-Augmenter/pull/306

transformations/use_acronyms/test.json
```diff
@@ -0,0 +1,61 @@
...
+      ]
+    },
+    {
+      "class": "UseAcronyms",
+      "inputs": {
+        "sentence": "A strong ElectroMagnetic Pulse will damage nearby electronic devices."
+      },
+      "outputs": [
+        {
+          "sentence": "A strong EMP will damage nearby electronic devices."
```
kaustubhdhole, line 56: Why not do the reverse also since you have access to this large dictionary? Or set a default parameter in the constructor to do so selectively.

Sotwi: Hi, the problem is that an acronym might represent different sets of words (eg. IMRI can be "Information Management Research Institute" ,"International Market Research Information", or "Intraoperative Magnetic Resonance Imaging"). Because of that the transformation Acronym -> Set of Words is not trivial. Not even sure it's solvable as of right now.

kaustubhdhole: Aah! I guessed that would be the case! Cool, I will merge this! Thanks for the clarification!

kaustubhdhole (approved):

PR #220, opened 2021-08-25
simple ciphers transformation

Sohl-Dickstein:
This transformation modifies the input text in ways that a human could rapidly decipher, but which make the input sequences almost completely unlike typical input sequences used during model training.


transformations/simple_ciphers/transformation.py
```diff
@@ -0,0 +1,132 @@
...
+  """
+
+  def __init__(self, seed=0, max_outputs=8):
+    self.rng = random.Random(seed)
+    super().__init__(seed, max_outputs=max_outputs)
+
+  
+  tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+  languages = ["en"]
+
```
ashish3586, line 103: Please add keywords.

mille-s (approved):

kaustubhdhole:
This is a wonderful transformation. This also emphasizes one of the larger goals of the transformation - which we want our models to be able to give high scores to pairs which might not be generally natural language but say, naturally understood by us. I would also suggest @Sohl-Dickstein to perform a robustness testing on the same to be added to the leaderboard.

PR #229, opened 2021-08-29
add gender randomizer

tk-sugumar:


transformations/gender_randomizer/README.md
```diff
@@ -0,0 +1,45 @@
+# Gender Randomizer 🦎  + ⌨️ → 🐍
+
+Author name: Tabitha Sugumar
+Author email: __
+Author Affiliation: __
+
```
kaustubhdhole, line 6: Thanks for your changes @tk-sugumar . Please add your email and affiliation.

tk-sugumar: Added!

transformations/gender_randomizer/README.md
```diff
@@ -0,0 +1,45 @@
...
+This uses the coreferee library (https://github.com/msg-systems/coreferee). The code is downloaded and included locally, to allow for slight modifications to the setup file to allow for installation in python 3.7, as required for this transformations (the coreferee library was designed/tested in python 3.8).
+
+The names directory comes from https://www.kaggle.com/nltkdata/names. The README within the directory provides more detail.
+
+## What are the limitations of this transformation?
+This transformation does not handle gendered words such as actor/actess, waiter/waitress, etc. The handling of pronouns is limited to what the coreferee library can identify, and, in this version is limited to the segment of text fed in one iteration (ie: if a text is separated and fed in batches, the pronouns/names will not be consistent accross the text).
+
+## Examples of this transformation
+
+Because this is a randomized transformation, in both the selection of gender and selection of name, test examples are impossible -- the output for a single sentence is expected to be different in each successive run. Instead I've provided some example sentences and outputs for reference.
```
timothy22000, line 23: I believe you can use a default seed in the argument in `init` of your GenderRandomizer transformation so you can generate consistent results for your test cases so you can include them in your test.json

Quite a few of the PRs use this approach for test cases.

See for example:
https://github.com/GEM-benchmark/NL-Augmenter/pull/164/files

transformations/gender_randomizer/README.md
```diff
@@ -0,0 +1,45 @@
+# Gender Randomizer 🦎  + ⌨️ → 🐍
+
+Author name: Tabitha Sugumar
+Author email: t.sugumar@elsevier.com
+Author Affiliation: Elsevier
+
+## What type of a transformation is this?
+This transformation changes names in English texts, randomizing selection so there's an even chance of male and female names. It modifies pronouns to match the selected name.
```
sebastianGehrmann, line 8: Please add an acknowledgement that names are not deterministic identifiers of someones pronouns/gender :)

transformations/gender_randomizer/README.md
tk-sugumar: Thanks Timothy! When I tried this, the same name was predicted for each sentence, so for use as intended the user would have to modify the code after downloading. Should I still go ahead and do this?

transformations/gender_randomizer/README.md
tk-sugumar: Added!

transformations/gender_randomizer/transformation.py
```diff
@@ -0,0 +1,175 @@
...
+import spacy
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+"""
+Randomizes names in text for a 50/50 gender breakdown. Handles pronouns.
+"""
+nlp = spacy.load("en_core_web_sm", disable=["lemmatizer"])
+nlp.add_pipe("coreferee")
```
msobrevillac, line 14: You might want to use spacy like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/d39f0e5f57a79937296351996617ac03ee80fb74/transformations/close_homophones_swap/transformation.py#L55).

transformations/gender_randomizer/transformation.py
```diff
@@ -0,0 +1,175 @@
...
+    if pronoun_text == 'herself':
+        return 'himself'
+    
+    return None
+
+
+class GenderRandomizer(SentenceOperation):
+    tasks = [TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
```
msobrevillac, line 89: Please, add some [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) here.

transformations/gender_randomizer/README.md
tk-sugumar: Hi Timothy, I added in the seed in the initializer, the name names does get predicted each time though, I hope it's ok! Test cases are also added in the test.json

transformations/gender_randomizer/transformation.py
tk-sugumar: Modified as given in example

transformations/gender_randomizer/transformation.py
tk-sugumar: Added

mille-s (approved):

transformations/gender_randomizer/README.md
```diff
@@ -0,0 +1,45 @@
...
+Author email: t.sugumar@elsevier.com
+Author Affiliation: Elsevier
+
+## What type of a transformation is this?
+This transformation changes names in English texts, randomizing selection so that there's an even chance of male and female names. Acknowledging that names are not deterministic identifiers of someone's gender/pronouns, here pronouns are modified to she/her/herself if the selected name is in names/female.txt and to he/his/him/himself if the selected name is in names/male.txt. These files can be modified or replaced as desired.
+
+## What tasks does it intend to benefit?
+This is intended to avoid gender bias in natural language processing models. Run this transformation on text data prior to using it to train a model.
+
+## Previous Work
```
kaustubhdhole, line 13: Importantly please add a Data and Code Provenance section to your transformation. Also, seems you've added about a 109 files which are hard to evaluate. I would suggest moving this into a separate pip project out of this and then adding it to the requirements.txt.

tk-sugumar: Thanks! I've expanded on the data and code provenance, and put the description in a Data and Code Provenance section in the Readme.

On the 109 files -- most of them come from the coreferee directory -- this actually already exists as a library installable by pip, but when I was working on this was only installable in python 3.8 and the current version requires python 3.9. Since these transformations are required to be compatible with python 3.7, I downloaded here to make it installable in python 3.7.

AbinayaM02:
> Hi @tk-sugumar, it won't be a good idea to merge all of these in the repository. It would be better to make a pip library out of it in a separate repository and call only the relevant parts here. @AbinayaM02 thoughts

Agreed. Like @kaustubhdhole mentioned, you should be installing the library (specify it in the reuirements.txt) and use it for your transformation @tk-sugumar. You can check if the library works fine for python 3.7.

PR #230, opened 2021-08-29
Add writing system replacement transformation

RomanPlusPlus:
It works as follows:

1. There is an input sentence (e.g. "I love potatoes")
2. A new random writing system is generated. For example, a logographic system where the word "love" is written as "蘋"
3. The output is the text in the new writing system. In out example, it's '蚴蘋䗑'

Currently, this transformation can convert the input into the following writing systems: alphabet, syllabary, partial phonemic, logographic.

### DIFF
--- a/transformations/writing_system_replacement/README.md
+++ b/transformations/writing_system_replacement/README.md
@@ -1,0 +1,56 @@
+# Writing System Replacement ❤️ → 心
+This transformation replaces the writing system of the input with another writing system.
+
+Author: Roman Sitelew (sitelewr at gmail dot com)
+
+A few snippets of the code are based on the code from the 
+[whitespace_perturbation transformation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/whitespace_perturbation), 
+authored by Xinyi Wu, and released under the same license (MIT) as the rest of the NL-Augmenter project.
+
+## What type of a transformation is this?
+It's a novel transformation that works as follows:
+
+1. There is an input sentence (e.g. "I love potatoes")
+2. A new random writing system is generated. For example, a logographic system where the word "love" is written as "蘋"
+3. The output is the text in the new writing system. In out example, it's '蚴蘋䗑'
+
+We use [CJK Unified Ideographs](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs) 
+as the source of characters for the generated writing systems. 
+
+Thanks to the enormous diversity of the CJK ideographs (tens of thousands of characters), 
+we can generate thousands of different writing systems, without running out of characters. 
+
+Currently, this transformation can convert the input into the following writing systems:
+
+|writing system type | some real-world languages that use it | "I love potatoes" (a sample output in a generated writing system)      |
+|---------------  | ------------------------------------  | -------------------------------------- |                                
+|alphabet         | English, Spanish, Russian             | 之 笓䒉㘔䆇 躓䒉蝲討蝲䒉䆇䁣             |
+|syllabary        | Mycenaean Greek, Japanese             | 䃓 熵鏘 㚐料                           |
+|partial phonemic | Hebrew,  Arabic                       | 阠㚶乍 渓绌敿                          |
+|logographic      | Ancient Egyptian, Oracle bone script  | 驿掩㑇㕶誨                             |
+
+## What tasks does it intend to benefit?
+This transformation could benefit text classification tasks (especially - language identification tasks).
+
+Trained humans are able to read texts that are written in different writing systems.
+For example, Japanese speakers can easily identify these texts as Japanese, and understand their meaning 
+(all 3 means "hiragana"):
+
+    平仮名
+    ひらがな
+    ヒラガナ
+
+The ability of ML models to identify and understand a language written in various writing systems could
+improve the quality of language identification in general, 
+and the quality of identification of low-resource languages in particular.
+
+As a useful side effect, it could help to decipher the [Voynich manuscript](https://en.wikipedia.org/wiki/Voynich_manuscript),
+the [Phaistos Disc](https://en.wikipedia.org/wiki/Phaistos_Disc), and other undeciphered artifacts.
+
+## What are the limitations of this transformation?
+
+If the input text is too short, the transformed text will not contain enough information 
+to identify the language of the original text. 
+For example, there is not enough information to decipher '驿掩㑇㕶誨', 
+where each word of the original English input was replaced with a random CJK ideograph.
+Thus, for language identification purposes, we would suggest inputs of 1000 characters or more. 
--- a/transformations/writing_system_replacement/README.md
+++ b/transformations/writing_system_replacement/README.md
@@ -47,10 +47,70 @@ and the quality of identification of low-resource languages in particular.
 As a useful side effect, it could help to decipher the [Voynich manuscript](https://en.wikipedia.org/wiki/Voynich_manuscript),
 the [Phaistos Disc](https://en.wikipedia.org/wiki/Phaistos_Disc), and other undeciphered artifacts.
 
+## Robustness Evaluation
+
+We've evaluated this transformation by running the build-in evaluate.py script as follows:
+`python3 evaluate.py -t WritingSystemReplacement`
+
+Note: it seems that there is a bug in 
+[this line](https://github.com/GEM-benchmark/NL-Augmenter/blob/f0111c1587cfa36cd4bd2c9739744e59c2796c26/TestRunner.py#L132)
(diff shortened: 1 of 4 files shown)


tia-e:
The idea is really great in the sense that this transformation will increase diversity of dataset in terms of writing systems.

However I was wondering how would it behave in scenarios  where the meaning matters?
One limitation I see is the complete lost of meaning of the original sentence. Does it even keep the output syntactically valid? (I don't think so).
Do you think it will have any impact in the task at hand?

It is not a generic system replacement, it is specifically a one way replacement model, right?
I think it would be beneficial to provide a way to support transformation in the both directions. Even if only one way is implemented. (Through the use of a source and target system variable for instance).

Overall, good idea!

transformations/writing_system_replacement/transformation.py
```diff
@@ -0,0 +1,502 @@
...
+    '䵹䵺䵻䵼䵽䵾䵿䶀䶁䶂䶃䶄䶅䶆䶇䶈䶉䶊䶋䶌䶍䶎䶏䶐䶑䶒䶓䶔䶕䶖䶗䶘䶙䶚䶛䶜䶝䶞䶟䶠䶡䶢䶣䶤䶥䶦䶧䶨䶩䶪䶫䶬䶭䶮䶯䶰䶱䶲䶳䶴'
+    """
+    cjk_codes = """
+    4E00–9FDF
+    3400–4DB5
+    """
+    """You use the ranges below too, to greatly increase the 
+    number of available characters.
+    But many systems can't render them correctly yet (as of 24 jul 2021).
+    Thus, enable them only if you're sure that characters are supported
```
tia-e, line 38: wouldn't it be interesting to control that with a boolean argument.

RomanPlusPlus:
Hi @tia-e, many thanks for your review! 
I've implemented your suggestions in the latest commit (ca03bbf).

Please find my answers to the questions below.

> One limitation I see is the complete lost of meaning of the original sentence. Does it even keep the output syntactically valid? (I don't think so).

You're right, the syntax is often not fully preserved. The meaning is preserved in an encrypted form, but only if the text is long enough to preserve enough information. 

For example, the transformation could generate the same text in a new alphabet. A sufficiently advanced model could be able to decrypt the text and extract the meaning.

I expanded the Limitations section to address this issue in detail.

> It is not a generic system replacement, it is specifically an one way replacement model, right?

I agree, in many cases it is indeed an one way replacement. For example, in partial-phonemic writing systems, some information is not encoded in the writing (i.e. the vowels are removed). We simulate such systems by removing a certain percentage of the characters, which can't be recovered afterwards.

transformations/writing_system_replacement/transformation.py
```diff
@@ -0,0 +1,502 @@
...
+import string
+import numpy as np
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+"""
+"""
+
+replace_white_spaces7 = False # if True, spaces will be considered as regular characters
```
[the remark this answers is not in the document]
RomanPlusPlus: fixed, thanks!

transformations/writing_system_replacement/transformation.py
RomanPlusPlus: Good idea! Added the argument

transformations/writing_system_replacement/README.md
```diff
@@ -0,0 +1,116 @@
...
+The accuracy on this subset which has 1000 examples = 96.0
+Applying transformation:
+100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1000/1000 [00:19<00:00, 51.83it/s]
+Finished transformation! 1000 examples generated from 1000 original examples, with 1000 successfully transformed and 0 unchanged (1.0 perturb rate)
+Here is the performance of the model on the transformed set
+The accuracy on this subset which has 1000 examples = 100.0
+
+```
+
+The performance is surprisingly good, considering the dramatic changes made by our transformation.
```
Sotwi, line 75: Consider adding the 4 suggested tests for Text Classification in the [evaluation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation)  section. You will notice that in the 3 non-default tests the performance falls significantly, as expected:

- `roberta-base-SST-2`: 94.0 -> 51.0
- `bert-base-uncased-QQP`: 92.0 -> 67.0
- `roberta-large-mnli`: 91.0 -> 43.0

It is still very interesting how the performance improves to 100.0 for `roberta-base-imdb`. It might be worth looking into that.

transformations/writing_system_replacement/transformation.py
```diff
@@ -0,0 +1,519 @@
...
+        "morphological",
+        "lexical",
+        "syntactic",
+        "rule-based",
+        "unnaturally-written",
+        "possible-meaning-alteration",
+        "high-coverage",
+    ]
+
+    def __init__(self, seed=0, max_outputs=1):
```
Sotwi, line 498: Is there a reason why many of the function parameters are hardcoded instead of passed as class parameters here and saved as class attributes? For example `replace_white_spaces`, `use_all_codes`, `exclude_white_spaces`, `main_rnd_generator`, `spaces_frequency`

Similarly, is there a reason for `chinese_chars` to be a global variable instead of a class attribute?

Sotwi (commented):
Other than those two comments, it looks very well to me.
The code runs, the right interface is used, tasks and keywords are indicated, it is a novel transformation, and enough test cases are provided.
Very interesting!

RomanPlusPlus:
Hi @Sotwi, thank you very much for your review! 

I've implemented your suggestions in the latest commit (d73b377).

transformations/writing_system_replacement/README.md
RomanPlusPlus: Thank you for the performance evaluation! Added the results to the readme.

Regarding the imdb anomaly, I'll open an issue about it.

transformations/writing_system_replacement/transformation.py
RomanPlusPlus: Good point! I've re-implemented them as class attributes.

Sotwi (approved):
As I said before, this is a very interesting transformation that surely can help dive into some interesting behaviors of current neural models. The whole situation with the default evaluation is a good proof of that.

Now that we've discussed the feedback and it's been applied I'm more than happy to accept it. Excellent work!

Quetzalcohuatl:
Just an outside observer, but I don't really understand what this transformation does. Is it translation into Chinese? Or translation into random different language?

RomanPlusPlus:
Hi @Quetzalcohuatl,

thanks for the question!

the transformation keeps the language, but replaces the writing system. 

For example, `I love potatoes` could be transformed into another alphabet like this: `了 乚凸丫丯 尸凸亍人亍凸丯乙...`, or into yet another alphabet like this: `之 笓䒉㘔䆇 躓䒉蝲討蝲䒉䆇䁣` (here, each letter was replaced with a randomly selected new character, e.g. `o` -> `䒉`).

Besides alphabets, it can transform into other writing systems. For example, into a logographic system like this: `I love potatoes` -> `驿掩㑇㕶誨`, where the entire word `potatoes` is written as a single character `誨`.

For the output, we use [CJK Unified Ideographs](https://en.wikipedia.org/wiki/CJK_Unified_Ideographs) (which are mostly Chinese characters), because t has enough characters to generate many thousands of different writing systems.

mille-s (approved):

Quetzalcohuatl:
So it is a substitution cipher? Like A is replaced by 去, B is replaced by 看, ..., Z is replaced by 就, but there is no actual meaning/reason for the replacement, it’s just a substitution cipher into CJK alphabet?

RomanPlusPlus:
> So it is a substitution cipher? Like A is replaced by 去, B is replaced by 看, ..., Z is replaced by 就, but there is no actual meaning/reason for the replacement, it’s just a substitution cipher into CJK alphabet?

It supports several mappings, not only alphabet -> alphabet. But I agree, basically it is a cipher. 

My reasoning is as follows:

These days, It is trivial to automatically identify `I love potatoes` as English, and `я люблю картошку` as Russian. But I'm not aware of a tech that can identify the language _without relying on the text's writing system_. 

As far as I know, ML can't do that yet, but humans can. 

For example, one can write `I love potatoes` as `aɪ lʌv pəˈteɪtəʊz`. If the text is long enough, a trained linguist can easily identify the text as English, in spite of the different writing system. 

Such an ability has several practical applications, including identification of the language of a text written in an unknown writing system (as in the case of the [Voynich manuscript](https://en.wikipedia.org/wiki/Voynich_manuscript)).

PR #231, opened 2021-08-29
Visual Attack Letters

Quetzalcohuatl:
perturbs each character with a visually similar letter. Very relevant for profanity filters. Some future ideas: we could mix lowercase and uppercase letters. And also expand beyond the standard english alphabet.

### DIFF
--- a/evaluation/README.md
+++ b/evaluation/README.md
@@ -56,6 +56,7 @@ Here, we provide a leaderboards for each default task, by executing transformati
 | ReplaceNumericalValues      | 94.0->94.0 (0.0)     | 92.0->92.0 (0.0)        | 91.0->90.0 (-1.0)    | 95.0->95.0 (0.0)    |
 | SentenceReordering          | 94.0->95.0 (1.0)     | 92.0->93.0 (1.0)        | nan                  | 95.0->94.0 (-1.0)   |
 | TokenReplacement            | 94.0->89.0 (-5.0)    | 92.0->84.0 (-8.0)       | 91.0->89.0 (-2.0)    | 95.0->92.0 (-3.0)   |
+| VisualAttackLetters         | 94.0->56.0 (-38.0)   | 92.0->69.0 (-23.0)      | 91.0->47.0 (-44.0)   | 95.0->96.0 (1.0)    |
 
 
 Default models and datasets:
--- a/transformations/visual_attack_letters/README.md
+++ b/transformations/visual_attack_letters/README.md
@@ -1,0 +1,57 @@
+# Visual Attack Letter
+This perturbation replaces letters with visually similar, but different, letters.
+Every letter was embedded into 576-dimensions. The nearest neighbors are obtained through cosine distance.
+How was the embedding obtained? The letter was resized into a 24x24 image, then flattened and scaled.
+This follows the Image Based Character Embedding (ICES) described in this paper: https://arxiv.org/pdf/1903.11508v1.pdf
+
+The top neighbors from each letter are chosen. Some were removed by judgment (e.g. the nearest neighbors for 'v' are many variations of the letter 'y') which did not qualify from the image embedding.
+You should be able to add more typefaces as they are created. (e.g. 🅰🅱🅲🅳🅴, 𝕒𝕓𝕔𝕕𝕖) if you want. For this version, I focused only on visual similarity.
+
+Author name: Corey James Levinson
+Author email: thecoreylevinson@gmail.com
+Author Affiliation: None
+
+## What type of a transformation is this?
+This transformation acts like a perturbation to test robustness and visual intuition of letters. 
+Transformations are chosen according to their ICES embedding. The embeddings were downloaded here: https://public.ukp.informatik.tu-darmstadt.de/naacl2019-like-humans-visual-attacks/
+Generated transformations display high similarity to the source sentences i.e. the code outputs highly precise generations. 
+
+## What tasks does it intend to benefit?
+This perturbation would benefit tasks which measure effectiveness against visual attacks. For example, it can be used to test robustness of profanity filters.
+
+```python evaluate.py -t VisualAttackLetters -task TEXT_CLASSIFICATION```
+```model_name = "aychang/roberta-base-imdb"```
+
+
+## Previous Work
+
+```bibtex
+@article{DBLP:journals/corr/abs-1903-11508,
+  author    = {Steffen Eger and
+               G{\"{o}}zde G{\"{u}}l Sahin and
+               Andreas R{\"{u}}ckl{\'{e}} and
+               Ji{-}Ung Lee and
+               Claudia Schulz and
+               Mohsen Mesgar and
+               Krishnkant Swarnkar and
+               Edwin Simpson and
+               Iryna Gurevych},
+  title     = {Text Processing Like Humans Do: Visually Attacking and Shielding {NLP}
+               Systems},
+  journal   = {CoRR},
+  volume    = {abs/1903.11508},
+  year      = {2019},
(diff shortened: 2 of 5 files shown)


transformations/visual_attack_letters/README.md
```diff
@@ -0,0 +1,56 @@
...
+  year      = {2019},
+  url       = {http://arxiv.org/abs/1903.11508},
+  archivePrefix = {arXiv},
+  eprint    = {1903.11508},
+  timestamp = {Tue, 02 Apr 2019 11:16:55 +0200},
+  biburl    = {https://dblp.org/rec/journals/corr/abs-1903-11508.bib},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+```
+
```
Sotwi, line 52: Add a Robustness Evaluation section as shown in the [evaluation](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation) section. Try to get the 4 model + dataset suggested for Text Classification.

transformations/visual_attack_letters/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+  'X': ['Х', 'Χ', 'ⵝ', 'Ⅹ', 'Ӿ', 'ⵅ', 'ⴴ', 'Ẋ', 'ⴳ', 'Ẍ', 'Ӽ', 'Ҳ', '✘', '☓', 'ⴵ'],
+  'Y': ['Ү', 'Υ', 'Ỵ', 'Ẏ', 'Ƴ', 'Ỷ', 'Ý', 'Ῠ', 'Ỳ', 'Ɏ', 'Ỹ', 'Ÿ', 'Ϋ', 'Ŷ', 'Ȳ', 'Ῡ', 'ϒ'],
+  'Z': ['Ζ', 'Ẓ', 'Ẕ', 'Ƶ', 'Ż', 'Ȥ', 'Ź', 'ƻ', 'Ẑ', 'Ž']
+}
+
+
+class VisualAttackLetters(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION, TaskType.TEXT_TAGGING]
+    languages = ["en"]
+
```
Sotwi, line 90: Add a keywords attribute following the [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) section recommendations.

Sotwi (commented):
Besides some missing details the transformation looks great. The code runs and acts as expected, the right interface is used, there are specific details about the use of the transformation (ie. profanity filters), and it is a new transformation different from the previews PR.

Quetzalcohuatl:
Like honestly I would love to make it so that a letter from any set is swapped with another letter from that set. E.g. instead of just replacing `a` with `ạ`, it would be nice to also be able to transform `ạ` into `a`. That way, the languages isn't only [en] but it could be all languages that use the Latin alphabet (e.g. Portuguese).

I can't think of a good way to code it though. If you have ideas, let me know.

Otherwise, I will try adding the Robustness Evaluation and keywords stuff you recommended.

Thanks @Sotwi

Sotwi:
@Quetzalcohuatl That is a great idea and would really increase the coverage of the transformation. I have two ideas to on how you could do that, depending on what you prefer. Probably there are better ways to do it but that's what comes to mind right now.

**List of sets instead of dictionary**

This approach will be less efficient since you'll have to be constantly iterating over `letter_mappings`.
First, define `letter_mappings` as a list of lists instead of a dictionary (eg. substitute `'f': ['ƒ', 'ẜ', 'ḟ', 'ẝ', 'ʄ', 'ϝ']` with `{'ƒ', 'ẜ', 'ḟ', 'ẝ', 'ʄ', 'ϝ'}`).
Second, change the final part
> ```
> if letter in letter_mappings and random.random() < self.perturb_pct:
>     sentence_list[idx] = random.choice(letter_mappings[letter])
> ```
for something like this:
```
if random.random() < self.perturb_pct: # checking first lets you skip the rest some times
    for letter_family in letter_mappings: # iterate over every "letter family", that's what makes it less efficient
        if letter in letter_family : # if your letter belongs to that "letter family"
            sentence_list[idx] = random.choice(tuple(letter_family - {letter}) # Perturb the letter excluding the current letter
            break # to prevent going over all the other "letter families"
```

**Longer Dictionary**
This approach consist on hardcoding in the dictionary all the possible inputs, you can computer the complete mappings before and then just print it and copy and paste it in the code  or you can run code for that once in the initialization. In any case, you can code that with something like this:
```
complete_mappings = {}
for letter_family in letter_mappings.values():
    for letter in letter_family:
        complete_mappings [letter] = letter_family
letter_mappings = temp_mappings
```

Quetzalcohuatl:
Thanks for the feedback @Sotwi , I made some of the changes that you suggested.

The one that I couldn't figure out was the `evaluate.py` ... I tried running the example with ButterFingers, but my terminal just ends with this:

`Here is the performance of the model textattack/roberta-base-imdb on the test[:20%] split of the imdb dataset`

And it just hangs there, forever! I tried it on my local laptop and also tried a fresh git clone & evaluate.py on a cloud machine. It always hangs. Got any ideas? O_o

Sotwi:
Thanks for addressing those details and improving the transformation!

As soon as the robustness evaluation is available I'd be happy to approve the PR.
I run it on my end (on colab) just to make sure everything is working, here are my results in case you want to double check:

| Model and Dataset flags                                            | Change |
|:---------------------------------------------------------|:----------|
| `-m "textattack/roberta-base-SST-2" -d "sst2"`         | 94 -> 56 |
| `-m "textattack/bert-base-uncased-QQP" -d "qqp"` | 92 -> 69 |
| `-m "roberta-large-mnli" -d "multi_nli"`                     | 91 -> 47 |
| ` -m "textattack/roberta-base-imdb" -d "imdb"`       | 95 -> 96 |

Quetzalcohuatl:
@Sotwi Rebased the PR, and I added the results in the evaluate section. I also added a little section talking about tokenization and Unicode normalization.

Sotwi (approved):
As I said before, this is a very interesting and useful transformation, even more so now that I can cover so many other languages. All the feedback was addressed so I'm happy to approve it. Good work!

evaluation/README.md
```diff
@@ -56,6 +56,7 @@ Here, we provide a leaderboards for each default task, by executing transformati
 | ReplaceNumericalValues      | 94.0->94.0 (0.0)     | 92.0->92.0 (0.0)        | 91.0->90.0 (-1.0)    | 95.0->95.0 (0.0)    |
 | SentenceReordering          | 94.0->95.0 (1.0)     | 92.0->93.0 (1.0)        | nan                  | 95.0->94.0 (-1.0)   |
 | TokenReplacement            | 94.0->89.0 (-5.0)    | 92.0->84.0 (-8.0)       | 91.0->89.0 (-2.0)    | 95.0->92.0 (-3.0)   |
+| VisualAttackLetters         | 94.0->56.0 (-38.0)   | 92.0->69.0 (-23.0)      | 91.0->47.0 (-44.0)   | 95.0->96.0 (1.0)    |
```
kaustubhdhole, line 59: This is the largest drop so far I guess among the transformations (which probably is close to the input sentence).. :)

kaustubhdhole:
Thank you @Quetzalcohuatl for these changes. This satisfies the review criteria and your code is neat and easily editable due to the use of the dictionaries. The drops in numbers are surprising and great to know how brittle these models are.

And thank you very much @Sotwi for providing a thoughtful review.

kaustubhdhole (approved):

PR #233, opened 2021-08-30
Spelling mis-correction transform

nick11roberts:
This transformation is a mechanism for data augmentation which applies (neural) spelling correction to sentences whose words have been so heavily misspelled that spelling correction fails to recover the original word. This is inspired by text messages I often receive from certain anonymous sources. 

@vinayprabhu @rteehas @jjangsangy

### DIFF
--- a/transformations/spelling_mis_correction/transformation.py
+++ b/transformations/spelling_mis_correction/transformation.py
@@ -1,0 +1,118 @@
+import itertools
+import random
+import re
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+from neuspell import BertChecker 
+# TODO remember to add neuspell to requirements.txt
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+def modified_butter_finger(text, prob=0.1, keyboard="querty", seed=0, 
+        max_outputs=1):
+    random.seed(seed)
+    key_approx = {}
+
+    if keyboard == "querty":
+        key_approx["q"] = "qwasedzx"
+        key_approx["w"] = "wqesadrfcx"
+        key_approx["e"] = "ewrsfdqazxcvgt"
+        key_approx["r"] = "retdgfwsxcvgt"
+        key_approx["t"] = "tryfhgedcvbnju"
+        key_approx["y"] = "ytugjhrfvbnji"
+        key_approx["u"] = "uyihkjtgbnmlo"
+        key_approx["i"] = "iuojlkyhnmlp"
+        key_approx["o"] = "oipklujm"
+        key_approx["p"] = "ploik"
+
+        key_approx["a"] = "aqszwxwdce"
+        key_approx["s"] = "swxadrfv"
+        key_approx["d"] = "decsfaqgbv"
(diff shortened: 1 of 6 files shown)


Mayukhga83:
Hi @nick11roberts,
I was assigned to review your transformation. Thank you for the contribution. I would brief my findings

Correctness: All checks failed
Interface: I think correct.

Description: The README is not clear about the goal of the transformation. Firstly it says that it applies (neural) spelling correction to sentences whose words have been so heavily misspelled, but actually it does the opposite. The test cases shows that it takes a sentence and changes the structure in a nonsensical manner

Applicable Tasks & Keywords: Relevant
Specificity: transforms sentence
Novelty: ok
Adding New Libraries: No
Data and code source: Data and code well documented
Test Cases: 10 have been added (I think its good)
Evaluating Robustness: Not performed yet

Note> The README should be edited. Moreover I think the transformation has nothing to do with mis-spellings, it just modifies the structure in a nonsensical way. Therefore the nomenclature is not appropriate

nick11roberts:
Thank you for the feedback, and I think this indicates that the readme as it stands failed to properly communicate the contribution of this task. The idea of this transformation is to force the spelling correction algorithm to make mistakes - in other words, misspellings are artificially made to be so bad that the spelling correction algorithm chooses the **wrong** correction. Hence the nomenclature - "spelling _mis_-correction" as opposed to "spelling correction." Spelling correction is used as a subroutine here - the goal of the task is not to perform spelling correction on misspelled sentences. 

Said another way, the source sentence is counterfactually modified to contain heavy enough misspellings to cause errors in spelling correction. 

Indeed, the nonsensical output of the transformation is _intentional_. These types of spelling-correction errors occur in real life, and would therefore reasonably occur in real-world datasets. Furthermore, the structure of the sentence is not modified. The modification occurs on a per-word basis. 

I would be happy to add clarifying examples of the process by which these spelling correction mistakes occur in the readme.

transformations/spelling_mis_correction/transformation.py
```diff
@@ -0,0 +1,118 @@
...
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+from neuspell import BertChecker 
+# TODO remember to add neuspell to requirements.txt
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+def modified_butter_finger(text, prob=0.1, keyboard="querty", seed=0,
```
samsontmr, line 14: Could you mention what was modified?

samsontmr (commented):
This is a very nice idea! The build appears to be failing though.

Additionally, could you refactor the functions so that `spelling_mis_correction` operates on words instead of the entire sentence? In other words, he loop over the words in the sentence should be in `generate`.

AbinayaM02:
Hi @nick11roberts : To fix the build failure, please pull the latest code and add your transformation name in the test/mapper.py file in the appropriate dictionary. Since yours is a heavy transformation, please set the heavy flag to True in your transforamtion.py file and run the pytest separately and confirm if it passes [The workflow checks the test cases for only light transformations/filters].

kaustubhdhole:
Feel free to reopen after the issues have been fixed.

PR #234, opened 2021-08-30
Added tag_subsequence_substitution

ExplorerFreda:
Add subsequence substitution for sequence tagging augmentation: https://aclanthology.org/2021.findings-acl.307.pdf

### DIFF
--- a/transformations/tag_subsequence_substitution/README.md
+++ b/transformations/tag_subsequence_substitution/README.md
@@ -1,5 +1,5 @@
 # Subsequence substitution for sequence tagging augmentation 🦎  + 🦎 → 🐍
-
+Contributor: Haoyue "Freda" Shi ([freda@ttic.edu](mailto:freda@ttic.edu)), Toyota Technological Institute at Chicago
 
 ## What type of a transformation is this?
 This transformation performs same-label subsequence substitution for the task of sequence tagging, which replaces a subsequence of the input tokens with another one that has the same sequence of tags.
--- a/transformations/tag_subsequence_substitution/README.md
+++ b/transformations/tag_subsequence_substitution/README.md
@@ -1,0 +1,61 @@
+# Subsequence substitution for sequence tagging augmentation 🦎  + 🦎 → 🐍
+
+
+## What type of a transformation is this?
+This transformation performs same-label subsequence substitution for the task of sequence tagging, which replaces a subsequence of the input tokens with another one that has the same sequence of tags.
+
+### Example
+
+**Input**
+
+tokens: `['EU', 'rejects', 'German', 'call', 'to', 'boycott', 'British', 'lamb', '.']`
+
+(POS) tags: `['NNP', 'VBZ', 'JJ', 'NN', 'TO', 'VB', 'JJ', 'NN', '.']`
+
+**Drawn subsequence from input**
+
+tokens: `['British', 'lamb']`
+
+(POS) tags: `['JJ', 'NN']`
+
+**Reference (randomly drawn from datset w.r.t. subsequence tags)**
+
+tokens: `['Germany', 'imported', '47,600', 'sheep', 'from', 'Britain', 'last', 'year', ',', 'nearly', 'half', 'of', 'total', 'imports', '.']`
+
+(POS) tags: `['NNP', 'VBD', 'CD', 'NN', 'IN', 'NNP', 'JJ', 'NN', ',', 'RB', 'NN', 'IN', 'JJ', 'NNS', '.']`
+
+**Substitution content** (with the same label subsequence as the drawn subsequence from input)
+
+tokens: `['last', 'year']`
+
+(POS) tags: `['JJ', 'NN']`
+
+**Output**
+
+tokens: `['EU', 'rejects', 'German', 'call', 'to', 'boycott', 'last', 'year', '.']`
+
+(POS) tags: `['NNP', 'VBZ', 'JJ', 'NN', 'TO', 'VB', 'JJ', 'NN', '.']`
+
+See more details in the paper: [Substructure Substitution: Structured Data Augmentation for NLP](https://aclanthology.org/2021.findings-acl.307).
+
+## What tasks does it intend to benefit?
+This perturbation would benefit all sequence labeling tasks, and is originally tested on part-of-speech (POS) tagging.
+
+## What are the limitations of this transformation?
+One would need to specify a dataset to operate with when using this pertubation (see the argument `base_dataset: List[Tuple[List[str], List[str]]]` of the `TagSubsequenceSubstitution` class, which should consists of a list of (tokens, tags) tuples); however, this should be easy to obtain in most supervised learning settings, by setting it as the training set.
+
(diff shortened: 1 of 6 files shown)


transformations/tag_subsequence_substitution/README.md
```diff
@@ -0,0 +1,61 @@
+# Subsequence substitution for sequence tagging augmentation 🦎  + 🦎 → 🐍
+
```
kaustubhdhole, line 2: Please add your name/email/affiliation.

ExplorerFreda: Thanks, added.

kaustubhdhole:
Thanks @ExplorerFreda for the wonderful changes! Here is my overall review:

transformations/tag_subsequence_substitution/data.py
```diff
@@ -0,0 +1,11 @@
+from datasets import load_dataset
```
kaustubhdhole, line 1: I guess you've kept this code here for the benefit of the readers. It'll definitely be helpful! Maybe you want to comment it?

transformations/tag_subsequence_substitution/transformation.py
```diff
@@ -0,0 +1,159 @@
...
+            for ss_l in range(
+                min_n_gram, max_n_gram + 1
+            ):  # subsequence_length
+                for i in range(length - ss_l):
+                    subsequence_sentence = tuple(sentence[slice(i, i + ss_l)])
+                    subsequence_tags = tuple(tags[slice(i, i + ss_l)])
+                    stats[subsequence_tags][subsequence_sentence] += 1
+        return stats
+
+    def generate(
```
kaustubhdhole, line 74: You might want to add comments in the style of the PEP 257 Docstring Conventions.

transformations/tag_subsequence_substitution/data.py
kaustubhdhole: Okay, I just noticed that "example_dataset" is being used. So you can ignore my previous comment. This looks good! :)

transformations/tag_subsequence_substitution/transformation.py
```diff
@@ -0,0 +1,159 @@
...
+    (POS) tags: ['JJ', 'NN']
+Output:
+    tokens: ['EU', 'rejects', 'German', 'call', 'to', 'boycott', 'last', 'year', '.']
+    (POS) tags: ['NNP', 'VBZ', 'JJ', 'NN', 'TO', 'VB', 'JJ', 'NN', '.']
+"""
+
+
+class TagSubsequenceSubstitution(TaggingOperation):
+    tasks = [TaskType.TEXT_TAGGING]
+    languages = "All"
```
kaustubhdhole, line 39: You might want to add the relevant [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md).

transformations/tag_subsequence_substitution/transformation.py
```diff
@@ -0,0 +1,159 @@
...
+        ), "base_dataset should be a list of (tokens, tags) tuples"
+        self.base_stats = self.extract_stats(
+            base_dataset, min_n_gram, max_n_gram
+        )
+        self.max_n_gram = max_n_gram
+        self.min_n_gram = min_n_gram
+
+    @staticmethod
+    def extract_stats(dataset, min_n_gram, max_n_gram):
+        stats = collections.defaultdict(collections.Counter)
```
kaustubhdhole, line 61: Do you think it makes sense to store these stats in a file here?

kaustubhdhole:
Overall this satisfies the review criteria with the exception of keywords and docstring comments. There are evaluations conducted in the paper which are very relevant and this is a great addition to NL-Augmenter. The robustness evaluations can be skipped for this PR as there is no default model.
Thanks for your efforts!

transformations/tag_subsequence_substitution/transformation.py
ExplorerFreda: Thanks, added docstrings for the functions.

transformations/tag_subsequence_substitution/transformation.py
ExplorerFreda: Thanks for the reminder! Added.

transformations/tag_subsequence_substitution/transformation.py
ExplorerFreda: Sure, I've added saving stats for potential reuse.

mille-s:
Thanks for the submission! I like how generic this transformation is, but I'm not sure I understand what would come out of it. Is it to see if a PoS tagger is still able to predict the same sequence with a different (although potentially implausible) vocabulary?

I'm also wondering to what extent this is applicable to other languages beyond English, and if it is possible to control the plausibility of the ouptut sentence, which would be two very interesting features.

ExplorerFreda:
Thanks for your comments!

> Thanks for the submission! I like how generic this transformation is, but I'm not sure I understand what would come out of it. Is it to see if a PoS tagger is still able to predict the same sequence with a different (although potentially implausible) vocabulary?

Some clarification: as stated in our paper (https://aclanthology.org/2021.findings-acl.307.pdf), it generates "sentences" (which could be implausible of course) with pseudo part-of-speech tags---training on the combination of real sentences (original dataset) and the generated sentences can improve over state-of-the-art PoS tagging performance (Tables 1, 2). It is interesting to see that we can have SotA NLP models without any semantic controls but tag subsequence substitution. 

Also, my personal random thoughts: I suspect that unnatural sentences (such as the production of this augmenter) would be beneficial to NLP models, and may especially work for improving models' compositionally. 

> I'm also wondering to what extent this is applicable to other languages beyond English, 

This can definitely be applied to any (natural or program) language other than English, as long as it has a corresponding sequence of tags. 

> if it is possible to control the plausibility of the ouptut sentence, which would be two very interesting features.

Controlling plausibility would an interesting feature, where I would probably suggest using a language model (e.g., GPT-2) for scoring and leaving it for future work.

mille-s:
@ExplorerFreda : thanks for the detailed answer! I understand how your data can be useful at training time to improve the models, I'm just trying to make sure it can be used at test time too, which is the case of all the other transformations. Can you elaborate briefly on this particular aspect?

ExplorerFreda:
@mille-s : While it's not designed for test-time usage, I can imagine that this transformation might support "colorless green ideas sleep furiously"--style test. A more recent reference on that the current community is still interested in such test: https://arxiv.org/pdf/1803.11138 

> which is the case of all the other transformations
By the way, I'm not sure if I agree on this. To my best, automatic synonym substitution(https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/synonym_substitution) may also generate implausible sentences, and visual attack(https://github.com/GEM-benchmark/NL-Augmenter/tree/main/transformations/visual_attack_letters) introduces extra difficulty almost all the time.

mille-s:
@ExplorerFreda : about test time, my mistake, sorry for the confusion!

mille-s (approved):

kaustubhdhole:
Thank you @ExplorerFreda for your effort and thanks @mille-s for your review. I am merging this now.

kaustubhdhole (approved):

PR #242, opened 2021-08-30
Concat

KentonMurray:
This is a transformation that concats training data examples to create longer examples. This has been shown to be effective for Low-Resource Machine Translation by Nguyen et al., 2021

### DIFF
--- a/transformations/concat/README.md
+++ b/transformations/concat/README.md
@@ -25,3 +25,7 @@ This concatenation would benefit all text tasks that use a transformer (and like
 
 ## What are the limitations of this transformation?
 The transformation has only appeared in publications in Low Resource Settings and for Machine Translation. It is unclear what the benefits will be in a higher resource settting or in a monolingual task.
+
+## Robustness Evaluation 
+Undefined task type, switching to default task %s TEXT_CLASSIFICATION
+No default evaluation model exists for the interface <class 'interfaces.SentenceOperation.SentenceAndTargetOperation'> in the locale en.It's okay to skip the evaluation for the purpose of the PR. If you are interested to evaluate your perturbation on a task and a dataset, the right place to do it would to add a new class in the evaluation folder and call it from execute_model. That's it!
--- a/transformations/concat/README.md
+++ b/transformations/concat/README.md
@@ -1,0 +1,27 @@
+# Concatenate Two Random Sentences
+This transformation concats two random sentences as explained in [Nguyen et al., 2021](https://arxiv.org/abs/2105.01691).
+
+## What type of a transformation is this?
+This transformation creates longer data sequences.
+
+## What tasks does it intend to benefit?
+This concatenation would benefit all text tasks that use a transformer (and likely other sequence-to-sequence architectures). Previously published work has shown a large gain in performance of low-resource machine translation using this method. In particular, the learned model is stronger due to being able to see training data that has context diversity, length diversity, and (to a lesser extent) position shifting.  
+
+## Previous Work and References
+1) The transformation is from ["Data Augmentation by Concatenation for Low-Resource Translation: A Mystery and a Solution" (Nguyen et al., 2021)](https://arxiv.org/abs/2105.01691). 
+```bibtex
+@inproceedings{nguyen2021data,
+  title={Data Augmentation by Concatenation for Low-Resource Translation: A Mystery and a Solution},
+  author={Nguyen, Toan Q and Murray, Kenton and Chiang, David},
+  booktitle = "Proceedings of the International Workshop on Spoken Language Translation",
+  month = jul,
+  year = "2021",
+  address = "Online",
+  publisher = "Association for Computational Linguistics",
+  url = "https://arxiv.org/abs/2105.01691",
+  abstract = "In this paper, we investigate the driving factors behind concatenation, a simple but effective data augmentation method for low-resource neural machine translation. Our experiments suggest that discourse context is unlikely the cause for the improvement of about +1 BLEU across four language pairs. Instead, we demonstrate that the improvement comes from three other factors unrelated to discourse: context diversity, length diversity, and (to a lesser extent) position shifting.",
+}
+```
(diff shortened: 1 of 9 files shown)


transformations/concat/README.md
```diff
@@ -0,0 +1,31 @@
+# Concatenate Two Random Sentences
+This transformation concats two random sentences as explained in [Nguyen et al., 2021](https://arxiv.org/abs/2105.01691). This is the bilingual version. A monolingual version is available under concat monlingual.
+
```
Nickeilf, line 3: Could you please add your information here, name/contact/institution, etc.?

transformations/concat_monolingual/transformation.py
```diff
@@ -0,0 +1,79 @@
+import re
+
+from checklist.perturb import Perturb
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class ConcatMonolingual(SentenceOperation):
```
Nickeilf, line 9: I'm not sure about the goal of having a monolingual concat transformation. Is it intended for classification tasks?

I can understand that concating random source + translation might work as an augmentation method where model can learn from more combinations of data. However, for the monolingual concat, if only the source is concatenated with a random sentence, but the target (either a label in the classification task or a sentence in translation task) remains unchanged, how would it help the model? I would imagine there can be a meaning mismatch in this way.

Nickeilf (commented):
Hi, @KentonMurray! This transformation looks good and interesting. However, it works on corpus-level rather than sentence-level (which means that the most appropriate interface is not found at the moment). Therefore, the current implementation uses two variables to cache the previous sample and add it to the current sample. I think the best way to do it is to first load a parallel corpus, and then randomly sample from the corpus to do the concatenation. That might introduce more diverse results than simply taking the previous sample.

We currently have a PR under review, which added the corpus-level interface to this repo. I'd suggest let's wait for that PR to be merged, and then you can change this transformation from `SentenceOperation` to the new corpus-level operation.

@kaustubhdhole What's your opinion?

KentonMurray:
It's true that this operates on a corpus level - however previously published results (which this is based off of, Nguyen et al., 2021) have shown that concatenating two adjacent sentences works just as well as random. That paper actually surprised people because concatenating consecutive sentences had been claimed to use discourse information. However it didn't do any better than random. The benefit comes from just seeing variable length sentences, differing contexts, and positional information. All of these actually exist with adjacent sentences as well.

Nickeilf:
> It's true that this operates on a corpus level - however previously published results (which this is based off of, Nguyen et al., 2021) have shown that concatenating two adjacent sentences works just as well as random. That paper actually surprised people because concatenating consecutive sentences had been claimed to use discourse information. However it didn't do any better than random. The benefit comes from just seeing variable length sentences, differing contexts, and positional information. All of these actually exist with adjacent sentences as well.

Thanks very interesting, and I have no more questions about this transformation. Could you please add your name/email/affiliation in the README?

transformations/concat/README.md
KentonMurray: Done. Thanks!

KentonMurray:
@kaustubhdhole I only see one reviewer has been assigned. Does this need a second?

mille-s (approved):

KentonMurray:
I've attached the error message from the build. This is something new and appears to deal with "transformation.ChangeDateFormat" which is not a transformation that I created. Basically, expected 2019 vs. predicted 2020, in the output. Is this test causing errors for anyone else? It seems to be independent of my PR


`=================================== FAILURES ===================================
_________________________ test_operation[light-light] __________________________

transformation_name = 'light', filter_name = 'light'

    def test_operation(transformation_name, filter_name):
        initialize_models()
>       execute_test_case_for_transformation(transformation_name)

test/test_main.py:154: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/test_main.py:125: in execute_test_case_for_transformation
    execute_sentence_operation_test_case(transformation, test)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

transformation = <transformations.change_date_format.transformation.ChangeDateFormat object at 0x7f0fb42fdb50>
test = {'class': 'ChangeDateFormat', 'inputs': {'sentence': 'The first known case of COVID-19 was identified in Wuhan, China ...ber 2019.'}, 'outputs': [{'sentence': 'The first known case of COVID-19 was identified in Wuhan, China in Dec 2019.'}]}

    def execute_sentence_operation_test_case(transformation, test):
        filter_args = test["inputs"]
        outputs = test["outputs"]
        perturbs = transformation.generate(**filter_args)
        for pred_output, output in zip_longest(perturbs, outputs):
>           assert pred_output == output["sentence"], get_assert_message(
                transformation, output["sentence"], pred_output
            )
E           AssertionError: Mis-match in expected and predicted output for ChangeDateFormat transformation: 
E              Expected Output: The first known case of COVID-19 was identified in Wuhan, China in Dec 2019. 
E              Predicted Output: The first known case of COVID-19 was identified in Wuhan, China in Dec 2020.
E           assert 'The first kn... in Dec 2020.' == 'The first kn... in Dec 2019.'
E             Skipping 63 identical leading characters in diff, use -v to show
E             -  in Dec 2019.
E             ?           ^^
E             +  in Dec 2020.
E             ?           ^^

test/test_main.py:33: AssertionError`

kaustubhdhole:
Okay, I think @AbinayaM02 might have a better clue for that. I will let this be merged meanwhile since the build in main is also failing due to ChangeDate

kaustubhdhole (approved):

KentonMurray:
Got it. Thanks.

PR #244, opened 2021-08-31
Added abbreviation and acronyms replacement

ahonore:
This transformation changes abbreviations and acronyms appearing in a text to their expanded form and respectively, 
changes expanded abbreviations and acronyms appearing in a text to their shorter form.
E.g.: `send this file asap to human resources` might be changed to `send this file as soon as possible to HR`. As soon as possible is the expanded form of asap.


transformations/replace_abbreviation_and_acronyms/README.md
```diff
@@ -0,0 +1,50 @@
+# Replace Abbreviations and Acronyms 🦎  + ⌨️ → 🐍
+This transformation changes abbreviations and acronyms appearing in a text to their expanded form and respectively, 
+changes expanded abbreviations and acronyms appearing in a text to their shorter form.
+E.g.: `send this file asap to human resources` might be changed to `send this file as soon as possible to HR`. As soon as possible is the expanded form of asap.
```
kaustubhdhole, line 4: Please add your name/email/affiliation

mmeyer39: Hello, it is done :)

kaustubhdhole: Thanks!

transformations/replace_abbreviation_and_acronyms/abbreviations.txt
```diff
@@ -0,0 +1,130 @@
+ACCT:Account
```
kaustubhdhole, line 1: You might want to add a Data and Code provenance section in the README which mentions the source of this data.

mille-s:
Hello, thanks for the submission! I like the fact that the substitution can be performed in both directions, this will be very helpful. We may want to merge with https://github.com/GEM-benchmark/NL-Augmenter/pull/218?

mmeyer39:
@mille-s thanks for the comment :)
I looked at #218 and there are indeed similar things done here.
However, as you mentioned we can do both direction which complexified our code significantly compared to #218.
How do you propose to merge them?

mille-s:
@mmeyer39 Thanks for the answer! For that it would be nice to assess to what extent the two pieces of code produce different outputs for the same input.  Is this something you can discuss briefly with @Sotwi ? We may want to keep both tranformations separate if they are different enough, and if not you can join forces on this one.

mmeyer39:
@mille-s, I have verified that the transformation works and is coherent with the PR of @Sotwi.
It is the case.

However there remains questions that should be addressed concerning decisions to made about this transformation.
Currently this transformation is `case incensitive`.
This works well in many cases and in our original settings as we consider:
- that people might write things rapidly (e.g.: asap instead of ASAP, new york instead of New York, ...).
- that the domain covered by our abbreviation file is limited and preserves the meaning of the text in most cases.

But, when looking at @Sotwi abbreviation file, this decision of being `case incensitive` might not be the best one.
Some abbreviations correspond to stop words and thus, would break the whole text if replaced...

Furthermore I think that it is important to keep in mind that abbreviation replacement is highly context sensitive in one direction.
For example there is 3 possible replacement for `AT` in @Sotwi list (AT:Ashmore and Cartier Islands; AT:Advanced Technology; AT:Anti-Tank).

@kaustubhdhole , @ahonore, @mille-s, @aadesh11 do you have an opinion on what we should do with the case enforcement?

mille-s:
@mmeyer39 : I suppose that ideally we would have a case-sensitive approach for acronyms that do no correspond to existing words (a.g. "at"), and case-insensitive for acronyms that do not (e.g. "asap"). But for that we'd need another dictionary or something similar. To start with, it could be good to have a parameter to allow for generating both case-sensitive and case-insensitive approaches, and generate a precision-oriented and a recall-oriented conversions respectively.

About the ambiguity of acronyms, I believe that any advanced approach will need to disambiguate indeed, since one single acronym can correspond to a lot of different things. But this can be done later maybe.

transformations/replace_abbreviation_and_acronyms/transformation.py
```diff
@@ -0,0 +1,309 @@
...
+        new_token_text["is_abbreviation"] = True
+        new_token_text["is_expanded_abbreviation"] = True
+
+        new_tokens.append(new_token_text)
+        counter = finish + 1
+
+    return new_tokens
+
+
+class ReplaceAbbreviations(SentenceOperation):
```
aadesh11, line 172: Could you please add docstrings for class also?

transformations/replace_abbreviation_and_acronyms/transformation.py
```diff
@@ -0,0 +1,309 @@
...
+                    counter += finish
+                    break
+            counter += 1
+
+        # merge all found expanded abbreviation and tag them as abbreviation
+        tokens = fuse_token_values_for_expanded_abbreviations(
+            tokens, merge_collection
+        )
+        for t in tokens:
+            print(t)
```
aadesh11, line 254: Do we really need this print statement?

transformations/replace_abbreviation_and_acronyms/transformation.py
mmeyer39: done

transformations/replace_abbreviation_and_acronyms/transformation.py
mmeyer39: removed

mmeyer39:
@mille-s I have added support to select if the transformation is case-sensitive or not.
The rest can be done in another PR later on as mentioned.

mille-s (approved):

aadesh11 (approved):

PR #245, opened 2021-08-31
Adding financial amounts replacement

ahonore:
This transformation replaces consistently financial amounts throughout a text.
The replacement changes the amount, the writting format as well as the currency of the financial amount.
The change is consistent in regard to:
- the modifier used to change all amounts of the same currency throughout the text.
  - e.g., the sentence `I owe Fred € 20 and I need € 10 for the bus.` might be changed to `I owe Fred 2 906.37 Yen and I need 1 453.19 Yen for the bus.`
- the modifier used to change the amounts so that new amounts are relatively close to the original amount.
- the rate used for a change of currency, reflecting the actual bank rate.


marco-digio:
Hi @ahonore 
Thanks for this contribution! 
I really liked this transformation.

My main concern is that the authors state that "the rate used for a change of currency, reflecting the actual bank rate." that is not true (or at least, it will not be always true). As written in `entity_financial_amount.py`, the rates are from August 2021. This could affect the performance, mostly when involving bitcoins, since the rates changes a lot. 

My second concern is about different currencies in the input test. Can you please add a test example with a sentence with 2 currencies? Examples: "I own Fred € 10 and I need $ 10 for the bus" when transformed, the output currency will be the same or they could be transformed into 2 different currencies, as in the input sentence? 

Correctness: All checks have passed
Applicable Tasks & Keywords: I think you have chosen the relevant task and keywords.
Specificity: The change is specific to text classification and generation tasks containing financial amounts.
Adding New Libraries: ftfy
Data and code source: There is a referenced paper
Evaluating Robustness: Robustness evaluation has been performed with no loss in accuracy. 

mmeyer39:
Hi @marco-digio,

Thanks for your comment.
I will see how we can update and improve the transformation to reflect the actual rate, because, as you mentioned, this is indeed currently a limitation.

For the example with two different currencies, I will add one as proposed.
They should indeed be replaced by two different currencies (depending on the randomness).

marco-digio:
Ok, thanks. 
I am not sure that it is worth to implement the transformation so to reflect the actual rate, but I think it is better to clearly highlight that the rates are from August 2021 in README. 

Maybe, if the rates during the last days/weeks/months/years are easily available somewhere, it could be interesting to add as an argument the date of which the rate should be selected. 
However, I am not sure if this is feasible since I am not an expert about this field. I am also not sure that it is worth the effort (this probably depends on how the transformation will be used in the future).

mmeyer39:
@marco-digio, I have updated the code with a new example and I have updated the README.
I found a library that I might be able to use to get the actual bank rate.
I will see if I can use it, but it might not be for this version of the library.

mmeyer39:
@marco-digio, Thanks again for your comment and review.
I have found a solution for obtaining the up-to-date bank rate and I have modified the code accordingly :)

AbinayaM02:
Hi @mmeyer39 : Please don't force-push the changes. You should first "Fetch & Merge" your repository (you can find this on the Github UI). Once done, pull the main branch into your financial_amount_replacement branch and commit your changes and the use the command
```python
git push origin financial_amount_replacement
```
to push your changes and create a PR. Please do this and then I'll start the review process. Thanks!

mmeyer39:
@AbinayaM02, could you elaborate on this? Do you want a single clean commit? 
I have done force-push several times to correct some commit errors on my side.
Why is it necessary to "cancel" those force-push for the review? Would it be okay just to say that I won't do it in the future?

Considering the tests, I believe that it is a harmless addition that helps maintaining this transformation.
We are aware that it will not be run with the transformation tests, but it helps with understanding, maintaining and preventing code breaks.

AbinayaM02:
> @AbinayaM02, could you elaborate on this? Do you want a single clean commit?
> I have done force-push several times to correct some commit errors on my side.
> Why is it necessary to "cancel" those force-push for the review? Would it be okay just to say that I won't do it in the future?
> 
> Considering the tests, I believe that it is a harmless addition that helps maintaining this transformation.
> We are aware that it will not be run with the transformation tests, but it helps with understanding, maintaining and preventing code breaks.

Sure. The force-commits rewrites history of the NL-Augmenter repository with the history of your forked repository. And I see that your forked repository is not the latest. You need to fetch and merge the changes to your forked repository to preserve the NL-Augmenter's code/history. Force-push in general is a bad practice and the history will be lost.

We have automated the checking of tests for all the transformations/filters, which mandates only the presence of test.json. By default, if your transformation's functionality fails, the pytest workflow fails. The pytest workflow is run twice for your PR (see the Checks tab). I suspect it's because you have a test folder also in your transformation folder and workflow picks up the test folder. So, it's ideal to remove the transformation specific test folder. By doing so, we can adhere to the current design of the repository and it will be easy for us to maintain.

Hope this explanation helps!

mmeyer39:
@AbinayaM02, thanks for your answer.
I have rebased the main branch using fetch and merge on our fork and then rebased my branch so that the history is reflecting the main NL-augmenter history. I had to force push this change however.
I hope that it is okay as such.

I have added new changes to remove the tests directory and to improve this transformation according to @marco-digio comment.
Those have been pushed normally.

transformations/replace_financial_amounts/README.md
```diff
@@ -0,0 +1,80 @@
+# Replace Financial amounts 🦎  + ⌨️ → 🐍
+This transformation replaces consistently financial amounts throughout a text.
+The replacement changes the amount, the writing format as well as the currency of the financial amount.
+The change is consistent in regard to:
+- the modifier used to change all amounts of the same currency throughout the text.
```
AbinayaM02, line 5: The change is consistent with respect to:

transformations/replace_financial_amounts/README.md
```diff
@@ -0,0 +1,80 @@
...
+## What are the limitations of this transformation?
+The transformation's outputs are too simple to be used for data augmentation as a standalone module.
+However, combined with other modules, this transformation helps improve the understanding of the context, and generates simple but precise similar sentences (same semantic similarity).
+
+The transformation outputs financial amounts independently of the format expected for the country.
+I.e. we might have `12,38 $` or `12.38 $` output by the generator.
+
+The transformation can only change numeric financial amounts (i.e. it will not handle "two dollars" or "14k euros").
+
+Finally, the transformation is case-sensitive i.e., `13 usd` will not be considered as an amount.
```
AbinayaM02, line 75: Aesthetic change: Can be made into bullet points.

transformations/replace_financial_amounts/transformation.py
```diff
@@ -0,0 +1,437 @@
...
+Token = Dict[str, any]
+
+
+def create_tokens(text: str) -> List[Token]:
+    """
+    Create a list of tokens that represents the input text.
+
+    Parameters:
+        `text`: The text or sentence to split into tokens.
+    """
```
AbinayaM02, line 21: Please add what the function returns as well under "Returns" wherever it is missing.

transformations/replace_financial_amounts/transformation.py
```diff
@@ -0,0 +1,437 @@
...
+    text = ftfy.fix_text(text)
+    nlp = spacy.load("en_core_web_sm")
```
AbinayaM02, line 23: Use this code for using spacy,

```python 
import spacy
from initialize import spacy_nlp

nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")

File: transformations/replace_financial_amounts/entity_financial_amount.py [hunk 5]
```diff
@@ -0,0 +1,249 @@
...
+    ]["symbol"]
+CURRENCY_SYMBOLS = ["$", "£", "₤", "¥", "฿", "€"]
+
+
+def generate_new_amount(
+    amount: float, percentage_financial_amount_variation: int
+) -> float:
+    """
+    | Generate new amount from current amount.
+    |  The value of percentage_financial_amount_variation should be between 1 and 20 to ensure coherence.
```
AbinayaM02, line 49: Please add missing parameter explanation.

AbinayaM02 (commented):
Docstrings are missing param explanation at many places and is not consistent. It needs to be fixed. The transformation needs to be part of the right dictionary in test/mapper.py file, please add it. Added comments in the respective files. Please go through them. 

Also, I would like to understand what error is causing you to force-push changes. Ideally, if you fetch and merge the main branch and pull that into your working branch, you can make the commits and push it without any issue. We should be avoiding force-push at any cost. Let me know the exact error you faced so that we can fix it and avoid force-push.

transformations/replace_financial_amounts/README.md
```diff
@@ -0,0 +1,80 @@
...
+$ python evaluate.py -t ReplaceFinancialAmount
+Undefined task type, switching to default task %s TEXT_CLASSIFICATION
+Loading <imdb> dataset to evaluate <aychang/roberta-base-imdb> model.
+Here is the performance of the model aychang/roberta-base-imdb on the test[:20%] split of the imdb dataset
+The accuracy on this subset which has 1000 examples = 96.0
+Applying transformation:
+100%|███████████████████████████████████████████████████████████████████████████████| 1000/1000 [06:21<00:00,  2.62it/s]
+Finished transformation! 1000 examples generated from 1000 original examples, with 35 successfully transformed and 965 unchanged (0.035 perturb rate)
+Here is the performance of the model on the transformed set
+The accuracy on this subset which has 1000 examples = 96.0
```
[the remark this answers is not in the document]
mmeyer39: done, thanks for the pointer

transformations/replace_financial_amounts/README.md
mmeyer39: changed

transformations/replace_financial_amounts/README.md
mmeyer39: changed

transformations/replace_financial_amounts/transformation.py
mmeyer39: changed where applicable

transformations/replace_financial_amounts/transformation.py
mmeyer39: changed, thanks for the suggestion

transformations/replace_financial_amounts/entity_financial_amount.py
mmeyer39: changed where applicable

AbinayaM02 (approved):
Hi @mmeyer39: Thank your for your contribution. Here are my comments

- The transformation replaces financial amount from one currency to another.
- New libraries are used and README is proper.
- Test cases have been added.

Please pull the latest code and add the robustness evaluation the leaderboard, here https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation. 

Approving the PR anyway :)

kaustubhdhole:
Okay, one more thing - please add a Data And Code Provenance section mentioning the source of your dataset and the source of your code (if you've used any outside code).

mmeyer39:
@AbinayaM02 et @kaustubhdhole  it is done

kaustubhdhole (approved):

PR #247, opened 2021-08-31
French Nouns Transformation

Louanes1:


Louanes1:
Hello @kaustubhdhole , I got this error when trying to build:
image

have you encountered anything similar ? 
I specified the encoding within the setup.py file but I don't think I am supposed to change anything in there.

AbinayaM02:
> Hello @kaustubhdhole , I got this error when trying to build:
> image
> 
> have you encountered anything similar ?
> I specified the encoding within the setup.py file but I don't think I am supposed to change anything in there.

Hi @Louanes1 : Please pull the latest code. It should resolve the workflow failure.

mille-s:
Should there be just one transformation for Verbs https://github.com/GEM-benchmark/NL-Augmenter/pull/250, Nouns and Adjectives https://github.com/GEM-benchmark/NL-Augmenter/pull/249, since the method seems to be the same? The transformation could be called "Lexical susbstitutions for French".

tuetschek:
Hi, I've had a look, I like the idea and the dependencies seem OK, but I agree with @mille-s that merging this with #249 and #250 would make much more sense (you could add a parameter specifying which PoS to replace).

kaustubhdhole:
Agree with the comments above. By default, you can return all (nouns,adjectives and adverbs) and still have it parametrized. Since we are tabulating the PRs, I am closing this. Please address all the suggested changes in all the 3 PRs to https://github.com/GEM-benchmark/NL-Augmenter/pull/250

PR #248, opened 2021-08-31
Multilingual backtranslation - support for 201 languages

Filco306:
This is much like the previous [backtranslation module](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/back_translation/transformation.py) (I even based my codebase on the original backtranslation module in this repository). However, it is significantly more general as it should support backtranslation for 201 different languages. 

It makes use of the [MarianMT](https://huggingface.co/transformers/model_doc/marian.html) models produced by [Helsinki-NLP](https://huggingface.co/Helsinki-NLP). 

As mentioned, it supports multilingual through backtranslation and supports backtranslation (back and forth) for up to 232 language codes and 201 unique languages. By this, it means that 201 languages has at least one other language paired with itself, such that there exists two MarianMT models, one for translating to the intermediate language and one for translating back to the original language.


Filco306:
Oh, I feel stupid now. I see that this is a duplicate of `lost_in_translation`. Feel free to close it.

PR #250, opened 2021-08-31
French Verbs  Transformation

Louanes1:


transformations/french_synonym_verbs_transformation/transformation.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class FrenchVerbsSynonymTransformation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["fr"]
+
```
msobrevillac, line 72: Please add the some [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) too.

transformations/french_synonym_verbs_transformation/transformation.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, seed=0, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
```
msobrevillac, line 74: You don't use the param _max_outputs_. It means that you are generating all possible candidates?

transformations/french_synonym_verbs_transformation/transformation.py
```diff
@@ -0,0 +1,83 @@
...
+@Language.factory('french_lemmatizer')
+def create_french_lemmatizer(nlp, name):
+    return LefffLemmatizer()
+
+@Language.factory('POSTagger')
+def create_POSTagger(nlp, name):
+    return POSTagger()
+
+
+nlp = spacy.load('fr_core_news_md')
```
msobrevillac, line 26: You might want to use spacy like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/d39f0e5f57a79937296351996617ac03ee80fb74/transformations/close_homophones_swap/transformation.py#L55).

transformations/french_synonym_verbs_transformation/transformation.py
```diff
@@ -0,0 +1,83 @@
...
+from textblob_fr import PatternTagger, PatternAnalyzer
+import nltk
+nltk.download('wordnet')
+from textblob.wordnet import NOUN, VERB, ADV, ADJ
+import spacy
+from spacy_lefff import LefffLemmatizer, POSTagger
+from spacy.language import Language
+from nltk.corpus import wordnet
+import nltk
+nltk.download('omw')
```
msobrevillac, line 12: Maybe you might add the nltk in a similar way to spacy in initialize.py

transformations/french_synonym_verbs_transformation/test.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+  
+    
+    {
+      "class": "FrenchVerbsSynonymTransformation",
+      "inputs": {
+        "sentence": "J'ai enfin pu faire remorquer la voiture !"
+      },
+      "outputs": [{
+        "sentence": "J'ai enfin pu faire rouler la voiture !"
```
msobrevillac, line 45: I am not sure if you only change one verb per instance, i.e., you only generate one additional sentence per instance?

msobrevillac (commented):
I think this is an interesting transformation for French. It is at an early stage but it is promising.

mille-s:
Hi, thanks for the submission! Two questions: (i) does the substitution only affect infinitive verbs? And if so, is there a way to overcome this limitation? (ii) do we have an idea of the coverage of the lexicon?
Also, we'd need to be able to control to what extent the meaning is the same as the input sentence, or to justify why it could be interesting to create alternative with a different meaning.

kaustubhdhole:
@Louanes1 ping

Louanes1:
Hi, 
The code is supposed to return many sentences with a different verb for each sentence. (Only one verb is changed)
It is indeed more pertinent to regroup all 3 PRs into a single one. I'll try to group all codes and push them directly here. So that I'll change adjective, verb and noun. 
Is it possible to have a list of unordered sentences as expected output ?

AbinayaM02:
> Hi, The code is supposed to return many sentences with a different verb for each sentence. (Only one verb is changed) It is indeed more pertinent to regroup all 3 PRs into a single one. I'll try to group all codes and push them directly here. So that I'll change adjective, verb and noun. Is it possible to have a list of unordered sentences as expected output ?

Hi @Louanes1 : You've already merged the 3 PRs into one right? If so, we can close this PR. Let me know.

PR #251, opened 2021-08-31
initial commit of alliteration filter

ahoimarie:
A simple alliteration filter that checks if a sentence is an alliteration or not.

### DIFF
--- a/filters/alliteration/README.md
+++ b/filters/alliteration/README.md
@@ -1,0 +1,16 @@
+## keywords filter
+
+**Author: Marie Tolkiehn**\
+Center for Data and Computing in Natural Sciences, Universität Hamburg\
+marie.tolkiehn@desy.de
+
+## What type of a filter is this?
+
+This filter returns True if a sentence is an alliteration and False otherwise. 
+There is an option to remove stopwords from the sentences. 
+
+## Related Work
+
+## What are the limitations of this filter?
+There may be phonetic alliterations that are not captured by a graphematic approach. For example, `Phonetic` and `Fine` are phonetic alliterations but not graphematic ones. 
+This could be ameliorated e.g. by using Carnegie Mellon's pronouncing dictionary to compare each word. 
--- a/filters/alliteration/__init__.py
+++ b/filters/alliteration/__init__.py
@@ -1,0 +1,1 @@
+from .filter import *
--- a/filters/alliteration/filter.py
+++ b/filters/alliteration/filter.py
@@ -1,0 +1,48 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+from initialize import spacy_nlp
+import spacy
+import string
+
+class Alliteration(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self, stopwords: bool=False):
+        super().__init__()
+        self.stopwords = stopwords
+        self.nlp = spacy_nlp if spacy_nlp else spacy.load("en_core_web_sm")
+
+    def filter(self, sentence: str = None) -> bool:
+        def get_phonemes(word: str):
+            # We are adding some digraphs to avoid 'sand' and 'shady' to alliterate.
+            # Then we check for these digraphs first
+            digraphs = ["ch", "ph", "sh", "th"]
+            if word[:2] in digraphs:
+                return word[:2]
+            else:
+                return word[:1]
+
+        # Convert to lower, remove punctuation, tokenize into words
+        sentenceS = sentence.lower().translate(str.maketrans("", "", string.punctuation)).split()
+
+        # if self.stopwords: # This somehow does not work, it always returns interfaces.SentenceOperation.SentenceOperation, even with a getter method
+        if self.stopwords:
+            all_stopwords = self.nlp.Defaults.stop_words
+            # Remove all stopwords from our sentence
+            sentenceS = [word for word in sentenceS if word not in all_stopwords]
+
+        # tokenized = self.nlp(sentence, disable=["parser", "tagger", "ner"])
+        # tokenizedB = [token.text for token in tokenized if token.text.isalpha()]
+        # tokened_text = [w.lower() for w in tokenizedB]  # make it lowercase
+
+        first_phon = get_phonemes(sentenceS[0])
+        start_phon = [get_phonemes(word)==first_phon for word in sentenceS]
+
+        return all(start_phon)
+
+# Alliteration(SentenceOperation).filter("It is I in it.")
+# Alliteration(SentenceOperation).filter("It is not my fault.")
+# print(Alliteration(SentenceOperation).filter("She showed Shawn shady shandy."))
+# print(Alliteration(SentenceOperation).filter("She showed Shawn some shady shandy."))
+# print(Alliteration(SentenceOperation).filter("Peter Piper picked a peck of pickled peppers."))
--- a/filters/alliteration/requirements.txt
+++ b/filters/alliteration/requirements.txt
@@ -1,0 +1,1 @@
+spacytextblob==3.0.1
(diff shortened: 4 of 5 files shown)


filters/alliteration/README.md
```diff
@@ -0,0 +1,41 @@
...
+* filtering based on a polarity score sign
+    * if two polarity scores do not align, the filter returns false
+    (neutral score of 0 is acceptable by positive and negative polarities)
+    * enabled with `strict_polarity = True`
+* filtering based on some allowed difference in polarity scores
+    * polarity scores should not differ by some number from 0 to 1
+    * enabled with `strict_polarity = False`
+    * default value for the allowed difference in polarity scores is 0.5; to change it, set the `diff_allowed` argument
+
+Author: Maria Obedkova
```
kaustubhdhole, line 17: Please add your email & affiliation.

ahoimarie:
Sorry, I did not mean to close the pull request, I wanted to force-reload from the upstream branch to undo my erroneous changes to the wrong folder, which had caused the checks to fail.

james-simon:
Simply superb submission! Alliteration's always an amusing attribute to amplify. I just have a few minor comments.

1) In your README, you don't state what your exact criterion for alliterativity is. It's clear from your code, but it's worth mentioning towards the top of the README.

2) That criterion's pretty strong - few people would say that "Peter Piper picked a peck of pickled peppers" **isn't** alliteration! I think it'd make sense to weaken it (e.g. allow one off-word for every $k$ alliterative words) or allow the user to set the filter strength (e.g. take $k$ as an argument).

3) For multi-sentence input, what's your rationale for ignoring all but the first sentence? I'd intuitively expect that the filter would either check whether the entire passage was alliterative on the same letter or whether each sentence is independently alliterative.

Solid work! Let me know if you have any questions.

rteehas:
I enjoyed this submission as well. I agree with @james-simon, but on the last point I think that it would make the most sense to treat each sentence independently, since that's probably more likely than an entire passage being alliterative.

mille-s:
Thanks for the submission! I think it would be good to add a few words on why sentences with alliterations are a challenge, and for which tasks; right now this is not clear to me (even if I like the idea). As mentioned previously, please also add the method you apply to select the sentences. By the way, this is tagged as "Transformation" but should probably be "Filter".

ahoimarie:
Hi @james-simon, 
gratitude and greetings! Alliterations always attract audiences. 

1. I added a few more lines in the README to explain my criterion for alliterativity. Fixed in [9940840](https://github.com/GEM-benchmark/NL-Augmenter/pull/251/commits/994084009326a3287cbfdca22fbb7f07200f7c95)
2. I have added a new variable `min_alliteration_length` to alleviate it. Now, a sentence is deemed an alliteration if it contains at least `min_alliteration_length` words starting with the (first) phoneme found. Would it make sense not to use the first phoneme but the one that occurs most in each sentence?
3. I wanted to avoid checking if a whole paragraph was alliterative because I thought that would practically never be the case. But I agree that ignoring the remaining sentences is not optimal either. I now fixed it to go through all the input sentences and to return true if any of the sentences are alliterative. 

Does this make sense? 
Thanks again!

ahoimarie:
Hello @mille-s, 
thanks for your feedback! I have added a few more lines in the README to discuss alliterations and their use. Please see [9940840](https://github.com/GEM-benchmark/NL-Augmenter/pull/251/commits/994084009326a3287cbfdca22fbb7f07200f7c95). Would you like me to expand on it? 
Thanks again!

james-simon:
Ahoy! Although arguably ameliorating the aforementioned ails, your changes created certain contrasting concerns:

- if the criterion for alliterativity is for *any* k words to start with the first phoneme, won't a lot of long sentences just satisfy this by accident? For example, if a sentence starts with "The" and includes two other "the"s later, wouldn't that count? This doesn't seem like it actually checks for alliteration; I think the alliterative words do have to be close together. If you agree, you should rethink the rule. As an example, a criterion I might choose would be to take two arguments, `min_alliteration_length` and `allowable_offwords`, and then check if a sentence contains any long run (not necessarily starting with the first word) of `min_alliteration_length` alliterative words, allowing `allowable_offwords` offwords breaking up the run. For example, the sentence `Look, guys: Peter Piper picked a peck of pickled peppers` would satisfy `min_alliteration_length=6` with `allowable_offwords=2` (i.e. "a" and "of"). This criterion seems like it gets much closer to what the typical reader would actually call an alliterative sentence. Does that make sense?
- the header "Why is it a challenge?" doesn't seem right; I'd say something like "Why is this filter important?" instead
- typo in the README: "will they not be removed"

I think checking for a single alliterative sentence makes sense.

mille-s:
@ahoimarie thank for the details, I agree it could be used in a generation setting to see if a model manages to create alliterations, although I suspect that the subpopulations with alliterations in the datasets we are using will be extremely small. But with other (more literary) datasets that could be nice.

ahoimarie:
Hello! Hopefully having handled your handy help here, I imagine it indeed improved its impact. 

- Instead of checking the first phoneme of the sentence, I now use a rolling window of size `min_alliteration_length+allowable_offwords` to check if it contains one phoneme of at least the frequency `min_alliteration_length` in it. If the sentence is shorter than `min_alliteration_length+allowable_offwords`, the window reduces to `min_alliteration_length`. 
- Agreed, and changed.
- Yes, thanks! 

Thanks! Thorough thoughtful things! 
The thing is, the theme of this thrilling thread thrives without thumbing through a thick thesaurus.

james-simon:
Amazing! Apt augmentations, all. Two tiny things:

- I think "If the windowlen is smaller than the length of the data" in your docstring is meaning-reversed
- Is `allowable_offwords` a parameter the user can specify, or is it always 2? It'd be better if it could be chosen

After the aforementioned alterations are achieved, I'll animatedly accept.

ahoimarie:
Awesome as always. Appreciated!

- Impressive inspection! I immediately edited it. 
- `allowed_offwords` is indeed a parameter the user can specify. For example `print(Alliteration(stopwords = False, min_alliteration_length=5, allowed_offwords=3).filter("Illuminating illustration of this inane but interesting instruction."))` will return `True` but calling it with `allowed_offwords=2` won't. 

Any additional amendments?

james-simon:
Ja, I'm happy! Great filter; every doubt's crushed. Bravo! Accept.

mille-s (approved):

Makefile
```diff
@@ -15,5 +15,5 @@ clean: style
 
 # Test
 .PHONY: test
-test:
-	pytest -v --cov-config=pyproject.toml
```
kaustubhdhole, line 19: This line should not be commented.

ahoimarie: It didn't work on my setup with this config, but changed it back now. Thanks!

kaustubhdhole: Please remove this file from your changes too since backtracking errors would be hard. So, basically this file should not appear in your changes.

kaustubhdhole:
All looks good. Please fix the makefile and I am happy to merge :)

Makefile
ahoimarie: Thanks! I removed the makefile from tracking. Hope this is what you meant?

kaustubhdhole:
The makefile still shows in the tracking as deleted.

ahoimarie:
Hi @kaustubhdhole, sorry for the delay, I have been on sick leave because of an injury. 
I've added the Makefile again -- does it work now or shall I create a new PR?

Makefile
kaustubhdhole: Hey, @ahoimarie, this file should not be shown as a part of your changes.

kaustubhdhole:
@ahoimarie I missed this message somehow. Please create a separate PR and link this PR in the comment.

kaustubhdhole:
Ensure that when you click "Files Changed", the makefile remains untouched.

ahoimarie:
@kaustubhdhole I removed the makefile with this little trick mentioned on [stackoverflow](https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request), as creating a separate PR kept failing. But this should solve it too, shouldn't it?

PR #253, opened 2021-08-31
Noun compound paraphraser

juand-r:
This transformation replaces two-word noun compounds with a paraphrase, based on the compound paraphrase dataset from SemEval 2013 Task 4 (Hendrickx et al., 2013).


transformations/noun_compound_paraphraser/README.md
```diff
@@ -0,0 +1,124 @@
...
+union of the trial (train) and test sets, but we made the following modifications:
+
+- in cases where the same compound appeared in both the train and test set, we
+added together their frequency in order to determine the overall paraphrase score.
+- determiners at the beginning of the paraphrase were removed.
+
+## What are the limitations of this transformation?
+
+This transformation currently only checks for noun compounds from
+(Hendrickx et al., 2013) and therefore has low coverage.
```
asnota, line 50: What happens if the phrase entered by the user doesn't have any noun compound combination?

transformations/noun_compound_paraphraser/transformation.py
```diff
@@ -0,0 +1,78 @@
+import random
+
+import os
+import json
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
```
asnota, line 9: Please add docstrings, precising the ourpose of the function, the format description of its aruments and outputs.

transformations/noun_compound_paraphraser/transformation.py
```diff
@@ -0,0 +1,78 @@
...
+            All combinations of paraphrases will be returned whenever more
+            than one noun compound from the dictionary is found in the sentence.
+            Default: 1.
+        seed: initial seed. Defaults: 0.
+    """
+
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
```
asnota, line 41: The `TaskType.TEXT_TAGGING` doesn't seem relevant, since the is no tag tokens in the inputs or outputs.

asnota, line 41: The TaskType.TEXT_TAGGING doesn't seem relevant, since the is no tag tokens in the inputs or outputs.

transformations/noun_compound_paraphraser/transformation.py
```diff
@@ -0,0 +1,78 @@
+import random
+
+import os
+import json
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+def noun_compound_paraphraser(text, compounds):
```
asnota, line 10: Please add docstrings, precising the ourpose of the function, the format description of its aruments and outputs.

asnota (changes requested):

tuetschek:
I agree with @asnota 's comments, otherwise it looks good to me! 

Small remark regarding the `TEXT_TAGGING` problem: You wouldn't necessarily need tags input/output, but if you don't have that, you'd need to have the same number and type of words on the input and output (and that wouldn't work with this transformation).

juand-r:
> I agree with @asnota 's comments, otherwise it looks good to me!
> 
> Small remark regarding the `TEXT_TAGGING` problem: You wouldn't necessarily need tags input/output, but if you don't have that, you'd need to have the same number and type of words on the input and output (and that wouldn't work with this transformation).

Thanks for the review, and for clarifying the`TEXT_TAGGING` issue. I will remove the tag, since it won't work for this transformation.

transformations/noun_compound_paraphraser/transformation.py
juand-r: Thanks for your feedback, @asnota. I will make the requested changes ASAP.

transformations/noun_compound_paraphraser/README.md
juand-r: If there are no noun compound combination, it will return the original sentence, without modifications. I will put this in the README to make it more precise.

asnota: Ok, it works.

mille-s (approved):

kaustubhdhole (approved):

PR #256, opened 2021-08-31
Chinese Simplified Traditional perturbation

timothy22000:


transformations/chinese_simplified_traditional_perturbation/README.md
```diff
@@ -1,28 +1,24 @@
...
 ## Previous Work
 
-1) Database for Chinese characters: https://github.com/pwxcoo/chinese-xinhua
+1) Open Chinese Convert: https://github.com/BYVoid/OpenCC
 
 ## What are the limitations of this transformation?
-There could be Chinese characters that are not within the database of 16142 characters since there are over 50000 Chinese characters.
-However, the commonly utilized characters in modern Chinese are around 7000 - 8000 characters and most modern Chinese dictionaries will list around 16000 - 20000 characters so the database should cover most cases.
+Although it can only perform transformation to convert between Simplified and Traditional Chinese for now, however it can easily be extended to perform more transformation such as
+Japanese Kanji, character-level and phrase-level conversion, character variant conversion and regional idioms among Mainland China, Taiwan and Hong Kong.
```
tongshuangwu, line 21: Thanks for the PR! Is there a particular challenge that makes it hard to include these options in this PR?
(nit pick: Although...~however~ it can easily be)

transformations/chinese_simplified_traditional_perturbation/requirements.txt
```diff
@@ -1,2 +1,2 @@
 # for CloseHomophonesSwap
```
tongshuangwu: seems like this has changed :)

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -1,57 +1,82 @@
...
-        TaskType.TEXT_TO_TEXT_GENERATION,
-        TaskType.TEXT_TAGGING,
+        TaskType.TEXT_TO_TEXT_GENERATION
     ]
-    languages = ["en"]
+    languages = ["zh"]
 
-    def __init__(self, seed=0, max_outputs=1):
-        super().__init__(seed)
+    def __init__(self, seed=0, max_outputs=1, prob=1, config = 's2t'):
```
tongshuangwu, line 52: the code can use a bit of documentation on `prob`, `config`, etc.?

transformations/chinese_simplified_traditional_perturbation/README.md
```diff
@@ -0,0 +1,24 @@
+# Chinese Simplified （简体）and Traditional （繁体）Perturbation 🦎  + ⌨️ → 🐍
+This perturbation adds noise to all types of text sources containing Chinese words and characters (sentence, paragraph, etc.) by changing the words and characters between Simplified and Traditional Chinese.
+
+Author name: Timothy Sum Hon Mun
+Author email: timothy22000@gmail.com
+
+## What type of a transformation is this?
+This transformation perturbes Chinese input text to test robustness. Chinese words or characters that are in Simplified Chinese are picked at random will be replaced with words and characters in
```
mmeyer39, line 8: The last part of the sentence seems strange. Could you reformulate? I think that there is a verb that is miss-placed.

transformations/chinese_simplified_traditional_perturbation/README.md
```diff
@@ -0,0 +1,24 @@
...
+Traditional Chinese. This transformation can also be performed in the opposite direction.
```
mmeyer39, line 9: Is it `can be`? Is it implemented here? Do you need to change a parameter or is it done automatically? 
Could you add more information here?

transformations/chinese_simplified_traditional_perturbation/__init__.py
```diff
@@ -0,0 +1 @@
+from .transformation import *
```
mmeyer39, line 1: Could you add a newline to the file?

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+                          prob,
+                          seed,
+                          max_outputs,
+                          config
+                          ):
+    random.seed(seed)
+
+    perturbed_texts = []
+
+    if(config == 's2t'):
```
mmeyer39, line 19: Do you think that you could detect this automatically? This would be a big improvement to the PR I believe. 
I am not sure that all sentence from a corpus are either in simplified Chinese or in traditional Chinese so the automatic detection would improve the perturbation when working with Chinese content.

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+        converter = opencc.OpenCC('s2t.json')
+    elif (config == 't2s'):
+        converter = opencc.OpenCC('t2s.json')
+    else:
+        print("Configuration not specified")
```
mmeyer39, line 24: Is it intented to only print this here? Maybe you could raise an exception or a warning. Does the perturbation works without configuration?

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+    for _ in itertools.repeat(None, max_outputs):
```
mmeyer39, line 25: You could simplify by removing the import of itertools and using range: `for _ in range(max_outputs):`

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+        butter_text = ""
+        for chinese_character in text:
+            if random.random() <= prob:
+
+                new_chinese_character = converter.convert(chinese_character)
```
mmeyer39, line 30: I am not familiar with Chinese. Does this work with any character? Could this fail? 
What if there is a character that is not Chinese in the sentence? Like a quote or a company name?

transformations/chinese_simplified_traditional_perturbation/transformation.py
mmeyer39: Should the probability be 1 by default?
Maybe the wording of the variable can just be changed to address @tongshuangwu comment. Like changing `prob` to `character_replacement_prob` or similar and changing `config` with `converter_config` (i am not pleased with this either but you can find a better naming if you want to change it :) )

transformations/chinese_simplified_traditional_perturbation/transformation.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self, seed=0, max_outputs=1, prob=1, config = 's2t'):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.prob = prob
+        self.seed = seed
+        self.max_outputs = max_outputs
+        self.config = config
+
+    def generate(self, sentence: str):
+        perturbed_texts = chinese_simplified_traditional_perturbation(
```
mmeyer39, line 60: You can simplify by just doing the return here.

As the function does nothing, maybe put the content of `chinese_simplified_traditional_perturbation` in this function directly.

mmeyer39 (changes requested):

timothy22000:
Thanks @tongshuangwu and @mmeyer39  for your feedback! I will go through each of them :D

transformations/chinese_simplified_traditional_perturbation/README.md
timothy22000: Hi @tongshuangwu , it's quite straightforward to add them since it is just changing an argument :) I'll make the change to reflect it and added comments in the code as well as README to explain which configuration are available

transformations/chinese_simplified_traditional_perturbation/README.md
timothy22000: I've adjusted it a bit. Let me know how is it now. :D

transformations/chinese_simplified_traditional_perturbation/README.md
timothy22000: You'll have to change the `converter_config` argument. I'll update it in the README.md

transformations/chinese_simplified_traditional_perturbation/__init__.py
timothy22000: Addressed

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: I have changed this now so no longer relevant anymore

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: Addressed

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: If it is in English or another language, then it doesn't do anything with it so it won't fail. :)

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: I'll keep it this way for now since if it is extended in the future then it might be useful to keep it separate. What do you think?

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: Hmm, this would be interesting. I think the tricky bit will be how to categorize the words that are the same in both Simplified and Traditional Chinese. Sentences in Traditional Chinese can contain words that are considered Traditional but it will also be the same word in Simplified Chinese.

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: I kept it as `prob` to be consistent with other transformations but I agree that it can be confusing! I'll rename the both of them and add documentation in the code as well as README.md

I'll call them `transformation_prob` and `converter_config`.

I put the probability as 1 by default since most of the time you want to convert the whole text that is in Simplified to Traditional or vice-versa. This is because when you normally type using a Pinyin keyboard - you would be using either Simplified or Traditional. Typically, you won't switch between the two when you type using a Pinyin keyboard. There will be words/characters that are the same in both Simplified and Traditional though. :)

transformations/chinese_simplified_traditional_perturbation/transformation.py
mmeyer39: maybe you could loop over the text and take the option which is in majority. 
E.g. if 70% of words are traditional chinese then we use t2s. Otherwise we use s2t.
What do you think about that?

transformations/chinese_simplified_traditional_perturbation/transformation.py
mmeyer39: ok, thanks :)

transformations/chinese_simplified_traditional_perturbation/transformation.py
mmeyer39: It depends, 
Try to keep it simple for the moment and suggest amelioration in the Readme.
I believe that it is not a good habit to think to much of extension in the main branch. 
Those might never be done or coded and your code is complexified for no reason.
But that is my opinion :)

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: Hmm how would we count characters that are considered both Simplified and Traditional Chinese when looping over the text?  That would affect whether the text is majority Simplified or Traditional. 

Example for "Have you eaten yet?":

Simplified: 你吃饱了吗
Traditional: 你吃飽了嗎

As you can see, 3 out of the 5 characters are the same in both Simplified and Traditional which is "你", "吃", "了".

Therefore, I was thinking that it would still be better for the user to defined which transformation that they would like to do themselves. Now that I have also added the configurations that are available in the README file - they will know the options that are available. :)

What do you think @Nickeilf and @tongshuangwu ?

transformations/chinese_simplified_traditional_perturbation/transformation.py
timothy22000: I can understand your point! I've made the changes since it's a quick change. :)

PR #260, opened 2021-09-01
Special casing Filter

vyraun:
This filter checks if the input sentence has a special casing, i.e. the string is either all lowercased, all uppercased or has title casing.

Traditionally, neural models are trained without any explicit casing information provided by the tokenizer. The most widely used data-driven tokenization algorithms, such as sentencepiece (Unigram LM) or Byte-Pair-Encoding typically do not have any explicit factors to handle case information. Therefore, tackling edge cases pertaining to case is important in text generation problems, e.g. in Machine Translation if a newspaper headline in title casing is provided to the model, the general expectation is that the output should preseve this title case. Same is true for inputs which are fully uppercased (typical in ALERT messages) or fully lowercased (typical in chat messages or asr transcripts).

This filter can be used to create splits of specially cased sentences of 3 types, fully uppercased, fully lowercased or titlecased. Filtering out and testing on examples belonging to such special cased inputs can provide feedback for improving the model or guiding data augmentation accordingly.

The PR includes the context, test cases and related work. Thanks.

### DIFF
--- a/filters/special_casing/README.md
+++ b/filters/special_casing/README.md
@@ -15,6 +15,7 @@ This filter can be used to create splits of specially cased sentences of 3 types
 The proposed filter is quite simple. Related work in handling cases explicitly in neural models includes the work on factored segmenter [1] and Truecasing [2].
 
 [1] https://github.com/microsoft/factored-segmenter
+
 [2] https://arxiv.org/abs/2108.11943v1
 
 ## What are the limitations of this filter?
--- a/filters/special_casing/README.md
+++ b/filters/special_casing/README.md
@@ -1,0 +1,21 @@
+## Special Casing filter
+
+## What type of a filter is this?
+
+This filter checks if the input sentence has a special casing, i.e. the string is either all lowercased, all uppercased or has title casing.
+Author: Vikas Raunak (viraunak@microsoft.com)
+
+## Why is measuring performance on this split important?
+Traditionally, neural models are trained without any explicit casing information provided to the tokenizer. The most widely used data-driven tokenization algorithms, such as sentencepiece (Unigram LM) or Byte-Pair-Encoding typically do not have any explicit factors to handle casing information. Therefore, tackling edge cases pertaining to case is important in text generation problems, e.g. in Machine Translation if a newspaper headline in title casing is provided to the model, the general expectation is that the output should preseve this title case. Same is true for inputs which are fully uppercased (typical in ALERT messages) or fully lowercased (typical in chat messages).
+
+This filter can be used to create splits of specially cased sentences of 3 types, fully uppercased, fully lowercased or titlecased. Filtering out and testing on examples belonging to such special cased inputs can provide feedback for improving the model or guiding data augmentation accordingly.
+
+## Related Work
+
+The proposed filter is quite simple. Related work in handling cases explicitly in neural models includes the work on factored segmenter [1] and Truecasing [2].
+
+[1] https://github.com/microsoft/factored-segmenter
+[2] https://arxiv.org/abs/2108.11943v1
+
+## What are the limitations of this filter?
+Only three types of casing have been designated as special. I am not aware of other special casing types, but the proposed filter can be extended easily to increase coverage of such casing types.
--- a/filters/special_casing/README.md
+++ b/filters/special_casing/README.md
@@ -3,6 +3,7 @@
 ## What type of a filter is this?
 
 This filter checks if the input sentence has a special casing, i.e. the string is either all lowercased, all uppercased or has title casing.
+
 Author: Vikas Raunak (viraunak@microsoft.com)
 
 ## Why is measuring performance on this split important?
--- a/filters/special_casing/filter.py
+++ b/filters/special_casing/filter.py
@@ -1,0 +1,20 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class TextSpecialCasingFilter(SentenceOperation):
+    """Filter if any of the special casings appear in the sentence:
+    All text is capitalized or All text is lowercased or there is Title Casing in the Sentence.
+    """
+
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
+
+    def __init__(self):
+        super().__init__()
+
+    def filter(self, sentence: str = None) -> bool:
+
+        return any(
+            [sentence.isupper(), sentence.islower(), sentence.istitle()]
+        )
--- a/filters/special_casing/filter.py
+++ b/filters/special_casing/filter.py
@@ -8,7 +8,8 @@ class TextSpecialCasingFilter(SentenceOperation):
     """
 
     tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION, TaskType.TEXT_TAGGING, TaskType.QUESTION_ANSWERING, TaskType.QUESTION_GENERATION, TaskType.SENTIMENT_ANALYSIS]
-    languages = ["en"]
+    languages = ["en", "fr", "es", "it", "pt", "fi", "nl", "da", "cs", "hr", "bg", "be", "eu", "ru", "uk", "pl", "sv", "sk", "sl"]
+    keywords = ["casing", "uppercase", "lowercase", "titlecase", "Latin", "Cyrillic"]
 
     def __init__(self):
         super().__init__()
--- a/filters/special_casing/filter.py
+++ b/filters/special_casing/filter.py
@@ -7,7 +7,7 @@ class TextSpecialCasingFilter(SentenceOperation):
     All text is capitalized or All text is lowercased or there is Title Casing in the Sentence.
     """
 
-    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION, TaskType.TEXT_TAGGING, TaskType.QUESTION_ANSWERING, TaskType.QUESTION_GENERATION, TaskType.SENTIMENT_ANALYSIS]
     languages = ["en"]
 
     def __init__(self):
--- a/filters/special_casing/__init__.py
+++ b/filters/special_casing/__init__.py
@@ -1,0 +1,1 @@
+from .filter import *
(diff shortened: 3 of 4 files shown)


filters/special_casing/filter.py
```diff
@@ -0,0 +1,20 @@
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+
+class TextSpecialCasingFilter(SentenceOperation):
+    """Filter if any of the special casings appear in the sentence:
+    All text is capitalized or All text is lowercased or there is Title Casing in the Sentence.
+    """
+
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
```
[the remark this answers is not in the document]
denkle: I second on this question. Would also give a thought to such options as SENTIMENT_ANALYSIS and PARAPHRASE_DETECTION.

filters/special_casing/filter.py
```diff
@@ -0,0 +1,20 @@
...
+    languages = ["en"]
+
+    def __init__(self):
+        super().__init__()
+
+    def filter(self, sentence: str = None) -> bool:
+
+        return any(
+            [sentence.isupper(), sentence.islower(), sentence.istitle()]
```
denkle, line 19: Just as a sanity check, I would suggest to double-check that no such filter has been proposed before.

denkle: Is it important for us to know of which type was the sentence?
My understanding that as of know the filter only tells that the sentence is one of three types but does not tell which one exactly.

filters/special_casing/README.md
```diff
@@ -0,0 +1,22 @@
...
+Author: Vikas Raunak (viraunak@microsoft.com)
+
+## Why is measuring performance on this split important?
+Traditionally, neural models are trained without any explicit casing information provided to the tokenizer. The most widely used data-driven tokenization algorithms, such as sentencepiece (Unigram LM) or Byte-Pair-Encoding typically do not have any explicit factors to handle casing information. Therefore, tackling edge cases pertaining to case is important in text generation problems, e.g. in Machine Translation if a newspaper headline in title casing is provided to the model, the general expectation is that the output should preseve this title case. Same is true for inputs which are fully uppercased (typical in ALERT messages) or fully lowercased (typical in chat messages).
+
+This filter can be used to create splits of specially cased sentences of 3 types, fully uppercased, fully lowercased or titlecased. Filtering out and testing on examples belonging to such special cased inputs can provide feedback for improving the model or guiding data augmentation accordingly.
+
```
richplant, line 12: Can you add a robustness evaluation on some of the standard dataset/model combinations? Check [this](https://github.com/GEM-benchmark/NL-Augmenter/tree/main/evaluation) for details.

filters/special_casing/filter.py
```diff
@@ -0,0 +1,20 @@
...
+from tasks.TaskTypes import TaskType
+
+
+class TextSpecialCasingFilter(SentenceOperation):
+    """Filter if any of the special casings appear in the sentence:
+    All text is capitalized or All text is lowercased or there is Title Casing in the Sentence.
+    """
+
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION]
+    languages = ["en"]
```
richplant, line 11: This should work for most languages, not just English, right?

filters/special_casing/filter.py
richplant, line 19: Why not also check for Title Case and CamelCase as well? Also, how about adding a parameter to select which casings to test for?

filters/special_casing/filter.py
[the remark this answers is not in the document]
vyraun: thanks, yes, it can be applied to any task where the input is a text sequence. Added new tasks.

filters/special_casing/filter.py
vyraun: Yes, any Latin or Cyrillic script based language. Where could I get the language codes that we are using in the repo?

filters/special_casing/filter.py
vyraun: Thanks, I searched the PRs using the 'casing' keyword, it seems there isn't any casing based filter. The filter picks up sentences that are weird in terms of casing (vs normal "text"), so it's a single category in itself. We can return the type too, but does the api allow the return of explicit values for filters? Wouldn't that mean that it is better to have three filters here, one for each casing type so that the user can apply filters compositionally without any post-processing from their side? If we return explicit values, then aren't we basically asking the user to do further filtering themselves? Maybe @kaustubhdhole could weight in. Thanks.

vyraun: Thanks, the filter checks for title casing already. Also, camel casing is for words, right? Is there a sequence analogue to camel casing?

filters/special_casing/README.md
vyraun: Thanks, I am not sure if standard dataset pre-processing already lowercases everything in these tasks. Do you suggest any particular dataset/task from above that reports numbers while preserving input casing?

filters/special_casing/filter.py
richplant: I don't believe it's specified anywhere, but that's the convention I've seen used in all the transformations I've reviewed so far.

filters/special_casing/filter.py
richplant: I think it makes sense from a modularity perspective to have one configurable filter than three separate static filters, but I'm open to arguments. 

Are the types of casing actually mutually exclusive within a sequence though? It seems to me that you could have an input that includes multiple types, e.g. (the following is a STRANGE WAY to Type Words Doesn'tItSeem?)

Oh, and on camel case, it's just any sequence without spaces where the word boundary is instead indicated by capitalisation, so can be arbitrary length.

denkle: > it seems there isn't any casing based filter.

I agree, looks like there no casing based filters.

denkle: > The filter picks up sentences that are weird in terms of casing (vs normal "text"), so it's a single category in itself.

Okay, I buy the argument that all these cases can be considered all together as "weird".

denkle: > Wouldn't that mean that it is better to have three filters here, one for each casing type so that the users could apply filters compositionally without any post-processing from their side?

Fair point. The idea about three simple separate filters might still be valid though.

denkle: I second on the issue of the hybrid "weird" cases. Following the logic of combining several types of casing together such hybrid cases should be filtered as well. 
Also, ideally one would like to take into account the fact that some words such as proper nouns might used capital letter in the middle of the sentence but they should not be considered as "weird" hybrid cases.

mille-s:
Interesting filter, thanks for the submission! I like this idea, since casing can actually be crucial in some contexts, and I don't believe we have such a filter yet.

filters/special_casing/README.md
vyraun: @richplant also the robustness-evaluation seems to be only for transformations, rather than filters, so skipping this as of now.

filters/special_casing/filter.py
vyraun: Thanks @richplant , @denkle for the comments. Regarding weird "hybrid" cases, e.g. "STRANGE WAY to Type Words Doesn'tItSeem?" and "MICHAEL JACKSON charged in Nevada, U.S.A" or "Lion King a massive success", there is no generic logic that will be able to tell apart the "correct" from the "hybrid" weird, right? That's why the proposed filter works only at the sequence level currently. That doesn't cover all "weird" cases, but whatever it covers (all lowercase, all uppercase, title case), it does so with perfect precision.

vyraun: > Are the types of casing actually mutually exclusive within a sequence though? It seems to me that you could have an input that includes multiple types, e.g. (the following is a STRANGE WAY to Type Words Doesn'tItSeem?)

Yes, the cases in the filter are mutually exclusive ( a sentence can be either all lowercased, all uppercased or title-cased). The reason for not getting into the hybrid "weird" cases is that they don't have a consistent logic to exploit. I have an example of why that is in the above comment.

> Oh, and on camel case, it's just any sequence without spaces where the word boundary is instead indicated by capitalisation, so can be arbitrary length.

Yes, but that is for words (IsThisAGoodExample), right not sentences? In languages, barring a few, we typically expect white-space tokenized sentences. The sequence (sentence) analogue of camel case is Title Casing (e.g. "Germany Wins World Cup"), which is already covered.

filters/special_casing/filter.py
vyraun: Thanks, I have added multiple languages (using Latin, Cyrillic scripts) as the supported languages as well.

vyraun:
> Interesting filter, thanks for the submission! I like this idea, since casing can actually be crucial in some contexts, and I don't believe we have such a filter yet.

Thanks @mille-s , @timothy22000 , @denkle for the reviews! I have addressed the issues through the commits (tasks, keywords, languages) and comments above. Please let me know if any more changes would be helpful or if the PR is good to go. Thanks!

denkle (approved):

mille-s (approved):

filters/special_casing/filter.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TextSpecialCasingFilter(SentenceOperation):
+    """Filter if any of the special casings appear in the sentence:
+    All text is capitalized or All text is lowercased or there is Title Casing in the Sentence.
+    """
+
+    tasks = [TaskType.TEXT_CLASSIFICATION, TaskType.TEXT_TO_TEXT_GENERATION, TaskType.TEXT_TAGGING, TaskType.QUESTION_ANSWERING, TaskType.QUESTION_GENERATION, TaskType.SENTIMENT_ANALYSIS]
+    languages = ["en", "fr", "es", "it", "pt", "fi", "nl", "da", "cs", "hr", "bg", "be", "eu", "ru", "uk", "pl", "sv", "sk", "sl"]
+    keywords = ["casing", "uppercase", "lowercase", "titlecase", "Latin", "Cyrillic"]
```
kaustubhdhole, line 12: Seems you've added some new keywords here and was contemplating if all should be a part of the keywords list in docs. Some of them are very specific for your transformation and I also think that can be helpful as general keywords (for search, etc) rather than only for classification. What do you think @mille-s ?

kaustubhdhole:
Thank you @timothy22000 @richplant @denkle and @mille-s for the review and thanks @vyraun for the hard work. I am merging this now.

kaustubhdhole (approved):

PR #261, opened 2021-09-01
Chinese Synonym and Antonym Substitution

timothy22000:
**Changes**

- Added synonym substitution

**To-Do**
- Add Antonym substitution

### DIFF
--- a/transformations/chinese_antonym_synonym_substitution/test.json
+++ b/transformations/chinese_antonym_synonym_substitution/test.json
@@ -1,0 +1,50 @@
+{
+  "type": "chinese_antonym_synonym_substitution",
+  "test_cases": [
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "谢尔曼将访问中国天津，中国外交部主管中美关系的副部长谢峰将和她会谈."
+      },
+      "outputs": [{
+        "sentence": "谢尔曼将访问中国天津，中国外交部朱观中美关系的副部长谢峰蒋和她回滩."
+      }]
+    },
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "本意是指词的起源义，即词的最初意义。引申义是由词的本意引申出来的并经过推演发展而产生的意义"
+      },
+      "outputs": [{
+        "sentence": "本意始指词的起源义，即词的最初一医。引申义驶由词得本谊引申出来的并经过推演发展而缠声得意义"
+      }]
+    },
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "汉字是语素文字，总数非常庞大。汉字总共有多少字？到目前为止，恐怕没人能够答得上来精确的数字。"
+      },
+      "outputs": [{
+        "sentence": "酣紫是语素文字，总数非常庞大。汉字总共有多少字？到目前为止，恐怕美人能够大德上来精确的数字。"
+      }]
+    },
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "恰当的运用反义词，可以形成鲜明的对比，把事物的特点表达得更充分，给人留下深刻难忘的印象。"
+      },
+      "outputs": [{
+        "sentence": "恰当的运用反义词，可以形成鲜明的对比，把石屋的特点表达得更充分，给人溜下深刻难忘的印象。"
+      }]
+    },
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "随着两个遗迹文明的发展，他们终于开始了争斗。遗迹之间的能量冲突是战争的导火索，因为一方出现，另一方的遗迹能量就会相应的颓落。"
+      },
+      "outputs": [{
+        "sentence": "随着两个遗迹文明的发展，他们终于开始了正斗。遗迹之间的能量冲突驶战争德导火索，因为一方出现，另一方的伊籍能量就挥相应德颓落。"
+      }]
+    }
+  ]
+}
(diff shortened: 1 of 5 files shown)


transformations/chinese_antonym_synonym_substitution/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+                    new_chinese_character = random.choice(perturb_antonym)
+                else:
+                    new_chinese_character = perturb_word
+                butter_text += new_chinese_character
+            perturbed_texts.append(butter_text)
+    return perturbed_texts
+
+def load_chinese_antonym_data():
+    dirname = os.path.dirname(__file__)
+    words_list = []
```
Nickeilf, line 57: Why not use a dictionary to store the antonyms? The keys are all the words, and the values are lists of antonyms for the key.
```
antonym_dict = {
    word1: [word1_antonym1, word1_antonym2, ...],
    word2: [...],
    word3: [...],
    ...
}
```
I think using a dict is more straightforward than having a list of word pairs, and in this way, you don't need to loop over the list to get the indices.

transformations/chinese_antonym_synonym_substitution/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+from nlpcda import Similarword
+
+def chinese_antonym_synonym_substitution(text,
+                          prob,
+                          seed,
+                          max_outputs,
+                          config,
+                          chinese_antonym_data
+                          ):
+    random.seed(seed)
```
Nickeilf, line 17: This should be put in the transformation `__init__()` function

transformations/chinese_antonym_synonym_substitution/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+
+class ChineseAntonymAndSynonymSubtitution(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION
+    ]
+    languages = ["zh"]
+    keywords = ["lexical", "rule-based", "api-based", "written", "highly-meaning-preserving", "meaning-alteration", "high-precision"]
+
+    def __init__(self, seed=0, max_outputs=1, prob=1):
```
Nickeilf, line 81: Maybe setting the default probability to a smaller value as it might introduce many grammatical errors?

Nickeilf (commented):
This transformation is very interesting and not implemented in this repo.

I feel that the generated texts might contain many grammatical errors (could you also mention this in the README?) because the dictionary-based synonym and antonym replacement does not guarantee syntactical correctness. Maybe changing the probability to a smaller value might alleviate this issue.

Another point is that this transformation might change the sentence semantics. For example, the "文明" (noun) is replaced with "温文尔雅" (adj). Regarding the antonym substitution, replacing words with antonym does not guarantee that the entire sentence meaning is reversed. Therefore, the label for the generated texts is uncertain. Could you please also add a warning about this issue in the README?

transformations/chinese_antonym_synonym_substitution/test.json
```diff
@@ -0,0 +1,62 @@
+{
+  "type": "chinese_antonym_synonym_substitution",
+  "test_cases": [
+    {
+      "class": "ChineseAntonymAndSynonymSubtitution",
+      "inputs": {
+        "sentence": "汉字是语素文字，总数非常庞大。汉字总共有多少字？到目前为止，恐怕没人能够答得上来精确的数字。"
+      },
+      "outputs": [{
+        "sentence": "汉字是语素言，总额深深的大幅度。单字合计有些微字？到目前为止，恐没人能够答得上来毫厘不爽的数字。"
```
samsontmr, line 10: seems like some characters are dropped: 恐怕没人 => 恐没人

transformations/chinese_antonym_synonym_substitution/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Notice that the first input is the non-perturb version
+        output_list = output_list[1:]
+        perturbed_texts = output_list
+
+    if(config == 'antonym'):
+
+        output = jieba.lcut(text)
+        for _ in itertools.repeat(None, max_outputs):
+            butter_text = ""
```
samsontmr, line 35: You should use a list to accumulate the words before joining into a string instead of appending, which creates a new string every time.

transformations/chinese_antonym_synonym_substitution/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, seed=0, max_outputs=1, prob=1):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.prob = prob
+        self.seed = seed
+        self.max_outputs = max_outputs
+        self.chinese_antonym_data = load_chinese_antonym_data()
+
+    def generate(self, sentence: str, config : str = 'synonym'):
+        perturbed_texts = chinese_antonym_synonym_substitution(
```
samsontmr, line 89: Could you refactor this function to operate on words instead of the entire sentence? The loop over the words in the sentence should be in this `generate` function.

samsontmr (changes requested):

timothy22000:
@kaustubhdhole Thanks for the ping! I'll work on the changes today :)

transformations/chinese_antonym_synonym_substitution/transformation.py
timothy22000: Done!

transformations/chinese_antonym_synonym_substitution/transformation.py
timothy22000: Done!

transformations/chinese_antonym_synonym_substitution/test.json
timothy22000: It looks like this is a weakness of the library that is being used for the synonym substitution where it will drop words if the probability is high, I have added this to the README.

timothy22000:
@kaustubhdhole Hi! I was just checking back and just saw your comments. Thank you for making the last two changes on my behalf already!

PR #262, opened 2021-09-01
Font change

pithysr:
Inspired by social media posts, this transformation add noise to an input sentence by randomly changing the font of words
in a sentence.


marco-digio:
Hi @pithysr 
Thanks for this contribution! 
I really like this idea and I believe it will be really helpful. 

I will leave some comments in the code.

Correctness:All checks have failed, please fix this (maybe fetch upstream is enough)
Applicable Tasks & Keywords: I think you have chosen the relevant task but why not TaskType.TEXT_TO_TEXT_GENERATION, TaskType.TEXT_TAGGING ? Please also insert keywords as explained [here](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md)
Specificity: The change is very general. 
Adding New Libraries: It requires nltk==3.6.2 as explained in requirements.txt
Data and code source: Data and code source missing. Please add references in the README
Test Cases: 6 test cases have been added.
Evaluating Robustness: No robustness evaluation has been performed.
Languages other than English: I believe that the transformation works also for other languages even if the author specify just english.

transformations/font_change/transformation.py
```diff
@@ -0,0 +1,79 @@
...
+        transformed_sentence = list(sentence)
+
+        tokens_to_change = random.sample(
+            tokens_escaped_list,
+            random.randint(1, min(3, len(tokens_escaped_list) - 1)),
+        )
+        for ttc in tokens_to_change:
+            while True:
+                font = random.sample(list(fonts.keys()), 1)[0]
+                if font != "normal":
```
marco-digio, line 41: Why not just removing "normal" from the dictionary in the JSON file?

transformations/font_change/transformation.py
```diff
@@ -0,0 +1,79 @@
...
+                except KeyError:
+                    transformed_sentence[i] = transformed_sentence[i]
+
+        perturbed_texts.append("".join(transformed_sentence))
+    return perturbed_texts
+
+
+class FontChange(SentenceOperation):
+    tasks = [TaskType.TEXT_CLASSIFICATION]
+    languages = ["en"]
```
marco-digio, line 58: Why just English? I believe that this transformation works well for many other languages

transformations/font_change/transformation.py
marco-digio, line 41: Why not just removing "normal" from the dictionary in the JSON file?

transformations/font_change/transformation.py
pithysr: It can work with other languages in its current form but the dictionary I used (reference in the code and in the README file) does not cover the accented letters or characters outside the English alphabet. If such a resource is added, we can use this for other languages without skipping those characters.

transformations/font_change/transformation.py
pithysr: It was by choice. I did not want any discrepancy between my dictionary and the original resource. Of course, it can be removed.

pithysr:
Hello @marco-digio 
Thank you very much for your review. 
I agree that we can add the other two tasks. Should I add them and commit again?

transformations/font_change/transformation.py
```diff
@@ -0,0 +1,1115 @@
...
+
+from nltk import download as nltkdl
+from nltk.corpus import stopwords
+from nltk.tokenize import word_tokenize
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+nltkdl("stopwords")
+nltkdl("punkt")
```
kaustubhdhole, line 15: This could be moved inside the constructor I think..

transformations/hashtagify/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+from nltk.tree import Tree
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+nltkdl("words")
+nltkdl("maxent_ne_chunker")
+nltkdl("punkt")
+nltkdl("averaged_perceptron_tagger")
+nltkdl("stopwords")
```
kaustubhdhole, line 20: Even this could go inside the constructor..

transformations/hashtagify/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=666, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+        self.nlp = spacy.load("en_core_web_sm")
```
kaustubhdhole, line 97: You might want to use spacy like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/d39f0e5f57a79937296351996617ac03ee80fb74/transformations/close_homophones_swap/transformation.py#L55).

aadesh11, line 97: Instead of loading new spacy, please use global spacy like below:

transformations/hashtagify/transformation.py
```diff
@@ -0,0 +1,107 @@
...
+                        "#" + re.sub(" ", "", cand.title()),
+                        transformed_sentence,
+                    )
+
+        perturbed_texts.append(transformed_sentence)
+
+    return perturbed_texts
+
+
+class HashtagifyTransformation(SentenceOperation):
```
aadesh11, line 87: Please add a docstring for all the classes and functions with a proper description of params and return type.

transformations/font_change/transformation.py
```diff
@@ -0,0 +1,1114 @@
...
+                        transformed_sentence[i]
+                    ]
+                except KeyError:
+                    transformed_sentence[i] = transformed_sentence[i]
+
+        perturbed_texts.append("".join(transformed_sentence))
+    return perturbed_texts
+
+
+class FontChange(SentenceOperation):
```
aadesh11, line 53: Same here, add docstrings for classes and functions.

transformations/font_change/transformation.py
```diff
@@ -0,0 +1,1114 @@
...
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = [
```
aadesh11, line 58: As the list is very long, better to create a separate file for this and import it here.

pithysr:
The conflicts are with my other submission that is already approved. Should I open a new PR?

PR #265, opened 2021-09-01
MR value replacement

msobrevillac:
This is a transformation focused on replacing some values in a Key-Value Meaning Representation with related words. This is useful for Data-to-text generation tasks such as the E2E task.


requirements.txt
```diff
@@ -5,6 +5,9 @@ numpy
 # for back_translation
 torch
 
+# for MRValueReplacement
+torchtext==0.9.1
+
```
kaustubhdhole, line 10: Thanks for your changes @msobrevillac! Let this be added in the requirements.txt of your transformation folder.
Also, happy to see a transformation for key-value pairs!

transformations/mr_value_replacement/transformation.py
```diff
@@ -0,0 +1,88 @@
...
+from itertools import product
+import random
+from random import sample
+from tasks.TaskTypes import TaskType
+
+class MRValueReplacement(KeyValuePairsOperation):
+    tasks = [TaskType.E2E_TASK]
+    languages = ["en"]
+  
+    def __init__(
```
kaustubhdhole, line 15: It would be great if you could add [PEP 257](https://www.python.org/dev/peps/pep-0257/) style comments to all the functions.

transformations/mr_value_replacement/transformation.py
```diff
@@ -0,0 +1,88 @@
...
+from interfaces.KeyValuePairsOperation import KeyValuePairsOperation
+import spacy
+from itertools import product
+import random
+from random import sample
+from tasks.TaskTypes import TaskType
+
+class MRValueReplacement(KeyValuePairsOperation):
+    tasks = [TaskType.E2E_TASK]
+    languages = ["en"]
```
kaustubhdhole, line 13: Please add the relevant [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) too.

transformations/mr_value_replacement/transformation.py
```diff
@@ -0,0 +1,88 @@
...
+  
+    def __init__(
+        self, seed=0, n_similar=10, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+
+        random.seed(seed)
+        self.glove = torchtext.vocab.GloVe(name = "6B", dim = "100")
+        self.spacy_model = spacy.load('en_core_web_sm')
```
kaustubhdhole, line 21: You might want to use spacy like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/d39f0e5f57a79937296351996617ac03ee80fb74/transformations/close_homophones_swap/transformation.py#L55).

transformations/mr_value_replacement/transformation.py
```diff
@@ -0,0 +1,88 @@
...
+class MRValueReplacement(KeyValuePairsOperation):
+    tasks = [TaskType.E2E_TASK]
+    languages = ["en"]
+  
+    def __init__(
+        self, seed=0, n_similar=10, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+
+        random.seed(seed)
+        self.glove = torchtext.vocab.GloVe(name = "6B", dim = "100")
```
kaustubhdhole, line 20: Also seems Glove is also used in the "diverse_paraphrase" transformation. It might be better if there is a common Glove model just like a common spacy model?

transformations/mr_value_replacement/transformation.py
Saad-Mahamood: Seconded.

Saad-Mahamood (changes requested):
Once the comments from @kaustubhdhole are addressed I would be happy to merge this.

transformations/mr_value_replacement/transformation.py
msobrevillac: Many thanks for your suggestions. I'll do the changes

transformations/mr_value_replacement/transformation.py
msobrevillac: Sorry, this is the first time contributing and I am not sure where I should put the relevant keywords? should be in the PR? or in the code?

kaustubhdhole: You should add them in a separate variable called "keywords" like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/2eda3b3552597519aa5e88011e81f594afe4d5ee/transformations/add_hashtags/transformation.py#L92)

kaustubhdhole:
Also you'll need to pull main once since there are some conflicts with the test_main file.

transformations/mr_value_replacement/transformation.py
msobrevillac: Sure, I think so. Can I change this? or you could do this?

requirements.txt
msobrevillac: Update done!

transformations/mr_value_replacement/transformation.py
msobrevillac: Done!

transformations/mr_value_replacement/transformation.py
msobrevillac: Done!

transformations/mr_value_replacement/transformation.py
msobrevillac: Done!

transformations/mr_value_replacement/transformation.py
msobrevillac: It is done.

msobrevillac:
Many thanks for your suggestions, @kaustubhdhole and @Saad-Mahamood. Could you check if everything is ok now? I realized that test was not working in github but it worked on my PC.

msobrevillac:
Hi, I updated the code according to your suggestions. Please, could you check if all is ok?. Thank you very much!

msobrevillac:
Hi, I've updated the code according to your suggestions. Please, let me know if everything is ok. Thanks for your time!

kaustubhdhole:
Thanks @msobrevillac for making the changes and adding the test cases too. This seems to be a great addition for data-to-text tasks.

kaustubhdhole (approved):

transformations/mr_value_replacement/transformation.py
Saad-Mahamood: Hi @msobrevillac, the DocStrings for each function is incomplete. You need to add DocStrings for each parameter and return vale for each function.

Saad-Mahamood (changes requested):
Incomplete DocStrings for functions.

kaustubhdhole:
Hi @msobrevillac, please add the comments as @Saad-Mahamood suggested in a separate PR. That will make it clear for users to browse through your code. I am merging this meanwhile.

PR #269, opened 2021-09-01
adding c2s transformation

fabriceyhc:


requirements.txt
```diff
@@ -32,3 +32,6 @@ pre-commit==2.13.0
 
 flake8==3.9.2
 isort==5.9.1
+
+# for concept2sentence
+sibyl_tool==0.1.2
```
kaustubhdhole, line 37: Hi @fabriceyhc thanks for your changes. Please add this under the transformations folder.

fabriceyhc: Will do. Just to confirm, I'll be removing it from `NL-Augmenter\requirements.txt` and creating my own `requirements.txt` in the `concept2sentence` folder. 

Also, I see that the build is failing for not having a test.json file. I thought it was optional? My transformation has some randomness in it that may not yield the same outputs every time so the tests are likely to fail since they're looking for exact match. How can we work around this? If necessary, I can make the outputs deterministic, but it undermines a diversifying goal of the transformation.

kaustubhdhole: Hello @fabriceyhc my apologies for the delayed reply! Yes - you are right about the hierarchy. You can temporarily add the test cases like [this](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/transformations/diverse_paraphrase/test.json).

transformations/concept2sentence/transformation.py
```diff
@@ -0,0 +1,70 @@
...
+
+from sibyl import Concept2Sentence
+
+
+class C2S(SentenceAndTargetOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION
+    ]
+    languages = ["en"]
```
kaustubhdhole, line 15: Please add the relevant [keywords](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md) too.

transformations/concept2sentence/transformation.py
```diff
@@ -0,0 +1,70 @@
...
+            self.task_config = {
+                'input_idx': [1],
+                'tran_type': 'INV',
+                'label_type': 'hard',
+                'task_name': 'topic'
+            }
+
+    def generate(self, sentence: str, target: int):
+        new_sentence, new_target = self.c2s.transform_Xy(sentence, target, self.task_config)
+        return new_sentence[0], new_target
```
kaustubhdhole, line 48: This should generate a list of pairs or a list of tuples if I'm not wrong.

transformations/concept2sentence/test.json
```diff
@@ -0,0 +1,4 @@
+{
+    "type": "concept2setence",
+    "test_cases": []
+}
```
kaustubhdhole, line 4: If you are skipping these test cases - which I wouldn't recommend, it is better to add them in the README for everyone's visibility.

transformations/concept2sentence/transformation.py
```diff
@@ -0,0 +1,92 @@
...
+                'input_idx': [1],
+                'tran_type': 'INV',
+                'label_type': 'hard',
+                'task_name': 'topic'
+            }
+
+    def generate(self, sentence: str, target=None):
+        new_sentences, new_targets = [], []
+        for _ in range(self.max_outputs):
+            new_sentence, new_target = self.c2s.transform_Xy(sentence, target, self.task_config)
```
kaustubhdhole, line 67: Also you might want to mention the specific types of targets that "transform_Xy" expects.

transformations/concept2sentence/test.json
fabriceyhc: Added to the README

transformations/concept2sentence/transformation.py
fabriceyhc: Added to the README and to the transformation method's docstring

transformations/concept2sentence/transformation.py
fabriceyhc: Corrected

transformations/concept2sentence/transformation.py
fabriceyhc: Done

fabriceyhc:
Hi Everyone, with only a few days left in the review process I'm hoping that I can get your review feedback as soon as possible, make any additional changes, and get this transformation accepted. Please let me know if there's anything more I can do on my end. 

Regards, 
Fabrice

transformations/concept2sentence/transformation.py
```diff
@@ -0,0 +1,112 @@
...
+        "possible-meaning-alteration",
+        "low-precision",
+        "high-coverage",
+        "high-generations"
+    ]
+
+    def __init__(self, 
+                 seed=0, 
+                 max_outputs=1, 
+                 dataset=None,
```
kaustubhdhole, line 37: Okay, this looks good to me overall - can you just add a default dataset here? And seems you were unable to get the tests working because of that. You can take a look [here](https://github.com/GEM-benchmark/NL-Augmenter/blob/main/filters/length/test.json) how args can be specified in your test.json.

fabriceyhc: I don't think it would be wise to have a default dataset. The dataset argument is used to download a fine-tuned transformer from huggingface_hub, which will in turn be used to identify the most salient tokens in the input for concept selection. However, that's only relevant for that particular dataset. When dataset is None, the transformation uses a lightweight graph-based approach to pick interesting concepts from the input - its more general and less likely to cause issues for users down the road. 

I removed the tests because the transformation picks a random number of concepts to use each time to boost diversity, so the output is potentially different each run.

kaustubhdhole (approved):

PR #275, opened 2021-09-01
Random Walk Perturbation using BERT

sajantanand:
We using masked token prediction to perform a random walk on a sentence; i.e., we choose a word and block it out, relying on a language model like BERT to determine what word should be placed in that blank. By repeating this several times, we can reach a new sentence that is perturbed from the original one.

### DIFF
--- a/transformations/random_walk/transformation.py
+++ b/transformations/random_walk/transformation.py
@@ -1,0 +1,117 @@
+import itertools
+import random
+
+from interfaces.SentenceOperation import SentenceOperation
+from tasks.TaskTypes import TaskType
+
+"""
+Base Class for implementing the different input transformations a generation should be robust against.
+"""
+
+
+def butter_finger(text, prob=0.1, keyboard="querty", seed=0, max_outputs=1):
+    random.seed(seed)
+    key_approx = {}
+
+    if keyboard == "querty":
+        key_approx["q"] = "qwasedzx"
+        key_approx["w"] = "wqesadrfcx"
+        key_approx["e"] = "ewrsfdqazxcvgt"
+        key_approx["r"] = "retdgfwsxcvgt"
+        key_approx["t"] = "tryfhgedcvbnju"
+        key_approx["y"] = "ytugjhrfvbnji"
+        key_approx["u"] = "uyihkjtgbnmlo"
+        key_approx["i"] = "iuojlkyhnmlp"
+        key_approx["o"] = "oipklujm"
+        key_approx["p"] = "plo['ik"
+
+        key_approx["a"] = "aqszwxwdce"
+        key_approx["s"] = "swxadrfv"
+        key_approx["d"] = "decsfaqgbv"
+        key_approx["f"] = "fdgrvwsxyhn"
+        key_approx["g"] = "gtbfhedcyjn"
+        key_approx["h"] = "hyngjfrvkim"
+        key_approx["j"] = "jhknugtblom"
+        key_approx["k"] = "kjlinyhn"
+        key_approx["l"] = "lokmpujn"
+
+        key_approx["z"] = "zaxsvde"
+        key_approx["x"] = "xzcsdbvfrewq"
+        key_approx["c"] = "cxvdfzswergb"
+        key_approx["v"] = "vcfbgxdertyn"
+        key_approx["b"] = "bvnghcftyun"
+        key_approx["n"] = "nbmhjvgtuik"
+        key_approx["m"] = "mnkjloik"
+        key_approx[" "] = " "
+    else:
+        print("Keyboard not supported.")
+
+    prob_of_typo = int(prob * 100)
+    perturbed_texts = []
+    for _ in itertools.repeat(None, max_outputs):
+        butter_text = ""
+        for letter in text:
+            lcletter = letter.lower()
+            if lcletter not in key_approx.keys():
+                new_letter = lcletter
+            else:
+                if random.choice(range(0, 100)) <= prob_of_typo:
+                    new_letter = random.choice(key_approx[lcletter])
+                else:
+                    new_letter = lcletter
+            # go back to original case
+            if not lcletter == letter:
+                new_letter = new_letter.upper()
+            butter_text += new_letter
+        perturbed_texts.append(butter_text)
+    return perturbed_texts
+
+
+"""
+Butter Finger implementation borrowed from https://github.com/alexyorke/butter-fingers.
+"""
+
+
+class ButterFingersPerturbation(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_CLASSIFICATION,
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_TAGGING,
+    ]
+    languages = ["en"]
+
+    def __init__(self, seed=0, max_outputs=1):
+        super().__init__(seed, max_outputs=max_outputs)
+
+    def generate(self, sentence: str):
+        perturbed_texts = butter_finger(
+            text=sentence,
+            prob=0.05,
+            seed=self.seed,
+            max_outputs=self.max_outputs,
+        )
+        return perturbed_texts
+
+
+"""
+# Sample code to demonstrate usage. Can also assist in adding test cases.
+# You don't need to keep this code in your transformation.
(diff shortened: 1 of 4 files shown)


sajantanand:
Pytest currently fails, but this is because the transformation produces a different output each time it is ran on a sentence, as the masked word is chosen at random.

transformations/random_walk/transformation.py
```diff
@@ -0,0 +1,103 @@
...
+
+            for token in top_k:
+                word = tokenizer.decode([token])
+                new_sentence = new_text.replace(tokenizer.mask_token, word)
+                sentences.append(new_sentence)
+
+        old_sentences = copy.deepcopy(sentences)
+    assert len(sentences) == k**steps
+    return sentences
+
```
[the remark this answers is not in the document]
sajantanand: I refactored the code as suggested.

sajantanand:
@kaustubhdhole can you explain how the pytest ran by github works? I can pass the pytest locally using the command in the ReadME, but the one run by github that includes the "light" filter does not succeed.

transformations/random_walk/transformation.py
```diff
@@ -0,0 +1,215 @@
...
+        model ([type]): model
+        tokenizer ([type]): tokenizer
+        k (int, optional): how many similar words to find for a given tokenized-word. Defaults to 5.
+
+    Returns:
+        [list]: list of top k words
+    """
+    inputs = tokenizer.encode_plus(tokenized_text, return_tensors='pt', truncation=True, max_length = 512)
+    index_to_mask = torch.where(inputs.input_ids[0] == tokenizer.mask_token_id)
+    if index_to_mask[0].numel() == 0: # Since we are truncating the input to be 512 tokens (BERT's max), we need to make sure the mask is in these first 512.
```
RoyRin, line 48: nitpick: don't let text go over 80 characters in width (break comments into several lines, and run `yapf` or `black`)

transformations/random_walk/transformation.py
```diff
@@ -0,0 +1,215 @@
...
+    Args:
+        sentence ([type]): sentence to perform random walk on
+        tokenizer ([type]): tokenizer
+        model ([type]): model
+        k (int, optional): how many replacement words to try. Defaults to 5.
+
+    Returns:
+        [list]: k-sentences with masked word replaced with top-k most similar words
+    """
+    #text_split = re.split('[ ?.,!;"]', sentence)
```
RoyRin, line 78: nit: remove comment.

transformations/random_walk/transformation.py
```diff
@@ -0,0 +1,215 @@
...
+    score = util.pytorch_cos_sim(emb_A, emb_B)
+    return score
+
+class RandomWalk(SentenceOperation):
+    tasks = [
+        TaskType.TEXT_TO_TEXT_GENERATION,
+        TaskType.TEXT_CLASSIFICATION
+    ]
+    languages = ["en"]
+
```
RoyRin, line 155: don't forget to include keywords: https://github.com/GEM-benchmark/NL-Augmenter/blob/main/docs/keywords.md 
like:
`keywords = [
        "model-based", "lexical", "possible-meaning-alteration",
        "high-coverage", "high-generations"
    ]`

AbinayaM02:
> @kaustubhdhole can you explain how the pytest ran by github (`pytest -s --t=light --f=light`) works? I can pass the pytest locally using the command in the ReadME (`pytest -s --t=random_walk`), but the one run by github that includes the "light" filter does not succeed.

Hi @sajantanand: We're running only the test cases for light transformations in the Github actions. Since yours is a heavy transformation, you need to run the test locally and check if passes.

juand-r (approved):
Hi @sajantanand,
I was assigned to review your transformation. I think this should be accepted, but have a few comments and suggestions below.

Correctness: The code passed all the tests.
Interface: Looks good.
Applicable Tasks and Keywords: Languages: good. **Tasks: I'm not sure TEXT_TO_TEXT_GENERATION would really work correctly; see my comments below.** **Keywords: please add!**
Specificity: Not specific to a particular task. This could be used for data augmentation; however, it should be noted it is low-precision, and generations are sometimes unnatural.
Novelty: This is not already implement in NL-Augmenter.
Adding new libraries: sentence-transformers and transformers specifed in requirements.txt (with versions).
Description: The README looks good. (But if there are any publications using this kind of transformation, please add them)
Data and code source: You could rename section **Extras** from the README to "Data and code provenance", and add license information for the models.
Paraphrasers and data augmenters: while this is low-precision, I still think it is an interesting transformation for data augmentation.
Test cases: 5 cases, good.
Evaluating robustness: not present.
Languages: English only.  This could be expanded to other languages in the future.
Documentation: most functions have docstrings, and the ones that don't are short and easy to read.

One suggestion: depending on how this is to be used, it might be helpful to have the option to not change named entities. Actually, as this currently stands, wouldn't there be possible issues when using this for some text to text task such as summarization? (e.g., couldn't it modify some entity in the source but not in the reference, or modify them in different ways?)

sajantanand:
I have addressed Roy's comments, added keywords, added the $heavy$ flag, and am currently running the evaluation scripts. These scripts do take quite a while using the default parameters, as 32 sentences are generated for each input. If the scripts finish running on collab, I'll update the ReadMe with the results. Unfortunately I don't have the compute resources to run this locally.

Now I'll address @juand-r's helpful comments.

- Tasks: I am not sure any tasks type other than Task.TEXT_TO_TEXT_GENERATION makes sense as the transformation generates a new sentence of possibly-altered meaning. I agree that the model can generate new text that isn't consistent with the inputted text for the reason you mentioned. A single word is selected and replaced, so if there were a Named entity, only one iteration would be replaced. One approach we have to deal with this is to set a high $sim_req$ when initializing the class, requiring generated sentences to be similar in meaning to the original sentence.
- Description: I don't know of any publications that do similar transformations, so I've instead linked some other instances of random walks in NLP.
- Data and code source: I've made this change.

As to your suggestion, I don't know of any way to exclude named entities. That being said, I am not very experienced in NLP work, so I'm open to any implementation suggestions.

juand-r:
Thanks for adding the references on random walks for sentence similarity.

I thought the list of tasks specify the kind of tasks this transformation could be used for in practice? So for my comment regarding  Task.TEXT_TO_TEXT_GENERATION -- say you have a summarization task, with a dataset with text pairs (source text, reference summary). Then if the transformation changes named entities which appear in both the source and the reference summary, but in an inconsistent way, this would introduce noise if used as data augmentation (and it would not be a reliable evaluation dataset either). This would probably not be as much of an issue with TEXT_CLASSIFICATION (although I suppose it could sometimes; e.g., say the task is sentiment classification and you accidentally swap the sentiment by modifying words like "good" or "great" to "bad").

You can find a list of named entities (and their character or token offsets) with an off-the-shelf package like spacy. Then you would only need to keep track of which words to exclude when looking for replacements.

sajantanand:
@juand-r Sorry for the long delay in response. I added an option to exclude named entities using spacy.

sajantanand:
@kaustubhdhole @vgtomahawk any chance this PR can get a second review? I added rudimentary evaluation results from Colab.

sajantanand:
@kaustubhdhole @mille-s One last bump to try and get a second review. Thanks for doing so much work organizing this!

james-simon:
@kaustubhdhole @mille-s This transformation has one accepting review, and as far as we can tell, the only reason it wasn't merged was that the second reviewer never reviewed it. We think this is an interesting transformation. We understand it's late in the project, but for the sake of closure, could you either give this a second review or definitively reject it?

Thanks!

mille-s:
@james-simon @sajantanand Sorry, very busy times! It looks like this transformation is part of a small group that escaped our radars, really sorry about that. At this point we carried out the analysis and plan to finalise the paper soon, so it is a bit difficult to add new things I'm afraid. I think that at this point the best option is to merge it later (we're actually running a GEM hackathon these days to add a lot of transfos/filters within the next weeks); since you have other accepted perturbations and are thus co-authoring the paper I hope this is not too much of an inconvenience. @kaustubhdhole does this sound good?

sajantanand:
@mille-s Sounds good to me! Let me know if there are any changes that need to be made before this is eventually merged.

PR #278, opened 2021-09-04
All transformations should now add keywords

kaustubhdhole:
Note that we would be sending a note shortly to everyone to include a list of keywords alongwith your transformations/filters.

### DIFF
--- a/docs/keywords.md
+++ b/docs/keywords.md
@@ -33,6 +33,9 @@ Keyword | Description
 **The naturalness of the generation** |
   `unnatural-sounding` | if the generations might not sound natural eg. if synthesized to speech, would they not sound natural as their source sentences. Eg, adding noun definitions might make the sentence factually correct but people might not use the same language while speaking.  
   `unnaturally-written` | generations in written format which might not look natural to a reader. homophonic translations, filler words
+  `visual` | The outputs are visually similar to the inputs (eg. Leet letters)
+  `aural` | The outputs sound similar to the inputs (eg. homophones)
+  `written` | The transformation makes font changes, or changes which would look different in different written formats
 **The potential accuracy & precision of the generation** | 
   `highly-meaning-preserving` | The output preserves the meaning of the source text
   `possible-meaning-alteration` | The output might change the meaning of the source text
@@ -43,13 +46,8 @@ Keyword | Description
   `low-coverage` | The transformation might not return an output for all types of inputs but only a handful of inputs
   `high-generations` | The transformation generates a lot of (possibly diverse) outputs for a single input (eg beam search or substitutions with large number of alternate choices)
   `low-generations` | The transformation generates few outputs for a single input (substitution with lesser number of alternate choices)
-**The text-specific modality being addressed (if applicable)** |
-  `visual` | The outputs are visually similar to the inputs (eg. Leet letters)
-  `aural` | The outputs sound similar to the inputs (eg. homophones)
-  `written` | The transformation makes font changes, or changes which would look different in different written formats
 **The skills required to differentiate between the input and the output of the transformation (if applicable)** |
   `causal-reasoning` | ability to reason about cause and effect
-  `visual-reasoning` | ability to reason by the appearance of the text
   `social-reasoning` | ability to understand and reason about human social interactions
   `figurative-language` | ability to understand figures of speech
   `world-knowledge` | ability to understand world knowledge
--- a/docs/keywords.md
+++ b/docs/keywords.md
@@ -27,9 +27,9 @@ Keyword | Description
   `external-knowledge-based` | if the implementation uses a corpora, a separate knowledge base, etc.
   `transformer-based` | if the implementation uses a transformer based model (eg. BERT, T5)
   `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
-  `tokenizer` | if the implementation uses a tokenizer
-  `chunker` | if the implementation uses a chunker
-  `discourse-segmenter` | if the implementation uses a discourse segmenter
+  `tokenizer-required` | if the implementation uses a tokenizer
+  `chunker-required` | if the implementation uses a chunker
+  `discourse-segmenter-required` | if the implementation uses a discourse segmenter
 **The naturalness of the generation** |
   `unnatural-sounding` | if the generations might not sound natural eg. if synthesized to speech, would they not sound natural as their source sentences. Eg, adding noun definitions might make the sentence factually correct but people might not use the same language while speaking.  
   `unnaturally-written` | generations in written format which might not look natural to a reader. homophonic translations, filler words
--- a/docs/keywords.md
+++ b/docs/keywords.md
@@ -32,7 +32,7 @@ Keyword | Description
   `discourse-segmenter` | if the implementation uses a discourse segmenter
 **The naturalness of the generation** |
   `unnatural-sounding` | if the generations might not sound natural eg. if synthesized to speech, would they not sound natural as their source sentences. Eg, adding noun definitions might make the sentence factually correct but people might not use the same language while speaking.  
-  `unnatural-looking` | generations in written format which might not look natural to a reader. homophonic translations, filler words
+  `unnaturally-written` | generations in written format which might not look natural to a reader. homophonic translations, filler words
 **The potential accuracy & precision of the generation** | 
   `high-precision` | The transformation generates what it intends to generate
   `low-precision` | The transformation sometimes does not generate what it intends to generate (eg. often changes meaning)
--- a/docs/keywords.md
+++ b/docs/keywords.md
@@ -15,40 +15,40 @@ Keyword | Description
 ------- | -----------
 **Type of Linguistic Change** |
   `morphological` | character level or inflectional/morphological changes
-  `lexical` | the transformation substitutes different words or phrases
-  `syntactic` | the transformation changes the syntax of the input (eg. active to passive)
-  `word-order` | if the order of the words or phrases is changed (eg. topicalisation, changing the order of semantic roles)
-  `discourse` | operating on or generating units of language longer than a sentence (eg. paragraph, multiple utterances in a dialog setting, etc)
-  `noise` | The generation adds random noise which would perturb the examples in a non-traditional manner (eg. removing punctuations, changing case, repeating words/characters)
+  `lexical` | the transformation substitutes different words or phrases
+  `syntactic` | the transformation changes the syntax of the input (eg. active to passive)
+  `word-order` | if the order of the words or phrases is changed (eg. topicalisation, changing the order of semantic roles)
+  `discourse` | operating on or generating units of language longer than a sentence (eg. paragraph, multiple utterances in a dialog setting, etc)
+  `noise` | The generation adds random noise which would perturb the examples in a non-traditional manner (eg. removing punctuations, changing case, repeating words/characters)
 **Type of algorithm** |
-  `rule-based` | if the implementation uses a heuristic or rules based approach
-  `model-based` | if the implementation uses a machine learning model
-  `api-based` | if the implementation uses an external api or tool
-  `external-knowledge-based` | if the implementation uses a corpora, a separate knowledge base, etc.
-  `transformer-based` | if the implementation uses a transformer based model (eg. BERT, T5)
-  `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
-  `tokenizer` | if the implementation uses a tokenizer
-  `chunker` | if the implementation uses a chunker
-  `discourse-segmenter` | if the implementation uses a discourse segmenter
+  `rule-based` | if the implementation uses a heuristic or rules based approach
+  `model-based` | if the implementation uses a machine learning model
+  `api-based` | if the implementation uses an external api or tool
+  `external-knowledge-based` | if the implementation uses a corpora, a separate knowledge base, etc.
+  `transformer-based` | if the implementation uses a transformer based model (eg. BERT, T5)
+  `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
+  `tokenizer` | if the implementation uses a tokenizer
+  `chunker` | if the implementation uses a chunker
+  `discourse-segmenter` | if the implementation uses a discourse segmenter
 **The naturalness of the generation** |
   `natural-sounding` | if the generations might sound natural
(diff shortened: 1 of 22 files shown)


docs/keywords.md
```diff
@@ -0,0 +1,51 @@
...
+  `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
+**The naturalness of the generation** |
+  `natural-sounding` | if the generations might sound natural
+  `natural-looking` | if the generations might look natural.
+  `unnatural` | if the generations might be valid pieces of text but not necessarily natural sounding
+**The potential accuracy & precision of the generation** | 
+   `accurate` | The output of the transformation generates highly accurate sentences
+   `near-accurate` | The generation would be mostly accurate
+   `high-precision` | The generations are highly precise
+   `low-precision` | The generations are not very precise
```
Nickeilf, line 38: I think the precision here is sort of repeating the accuracy. I feel that It's a bit confusing how to differentiate the two. The coverage makes sense.

I guess we might need another metric like "high/low number of generation"?

Nickeilf (approved):
Looks good!

docs/keywords.md
```diff
@@ -0,0 +1,51 @@
+  # NL-Augmenter Keywords
+
+Every transformation should have keywords that identify the type of transformation that is being written. The following list can be used as reference.
+- the type of linguistic change that the transformation is attempting eg. morphological, lexical, syntactic, word-order, discourse, noise etc.
+- the type of algorithm that is written eg. rule-based, model-based, api-based, external-knowledge-based, transformer-based, parser-based, etc.
+- the naturalness of the generation eg. natural-sounding, natural-looking, unnatural, etc.
+- the potential accuracy & precision of the generation eg. accurate, near-accurate, high-precision, high-coverage, etc.
+- the text-specific modality being covered (if applicable) eg. visual, aural, written, etc.
+- the skills required to differentiate between the input and the output of the transformation (if applicable) eg. causal-reasoning, visual-reasoning, social-reasoning, humor, figurative-language, grammar, world-knowledge, etc. 
+- other transformation specific keywords
```
vgtomahawk, line 10: Would it make sense to have another category of keywords to specify the preprocessing prerequisites required? (or maybe we can roll those into ``type of algorithm" category. e.g POS tagger, dependency parser, chunker, EDU segmenter, RST parser, tokenizer etc etc - after all preprocessing is also part of the model - but nonetheless we should maybe have these keywords available on there from the start)

docs/keywords.md
kaustubhdhole: I'm in two minds since there can be transformations which do not generate on certain utterances and there could be differences in P and A, isn't it?

And I will add two more keywords - "high-generations", "low-generations". Sounds like a good idea!

vgtomahawk (commented):
Is "high-generations/low-generations" the same as high-recall/low-recall = high-coverage/low-coverage ?

docs/keywords.md
kaustubhdhole: I think Zhenhao meant high-generations for a large number of generations (i.e. outputs) while high coverage was added for how much input space is covered. I will try to make that distinction prominent.

kaustubhdhole: Okay, I had added `parser-based` as a general category but now I've added some more fine-grained ones - `tokenizer`, `chunker`, `discourse-segmenter`, etc.

vgtomahawk: Thanks!!

vgtomahawk (approved):

docs/keywords.md
```diff
@@ -0,0 +1,56 @@
+  # NL-Augmenter Keywords
+
+Every transformation should have keywords that identify the type of transformation that is being written. The following list can be used as reference.
+- the type of linguistic change that the transformation is attempting eg. morphological, lexical, syntactic, word-order, discourse, noise etc.
+- the type of algorithm that is written eg. rule-based, model-based, api-based, external-knowledge-based, transformer-based, parser-based, etc.
+- the naturalness of the generation eg. natural-sounding, natural-looking, unnatural, etc.
+- the potential accuracy & precision of the generation eg. accurate, near-accurate, high-precision, high-coverage, etc.
```
Sohl-Dickstein, line 7: I don't know what accuracy and precision mean in this context. I would elaborate and/or rephrase.

Does accurate mean that the transformation preserves the original meaning of the string? Does precise mean that if the transformation changes the meaning, it does so in a consistent way across multiple strings (eg, it changes names or numbers consistently)?

The same comment applies to the "**The potential accuracy & precision of the generation** " section below.

docs/keywords.md
```diff
@@ -0,0 +1,56 @@
...
+- the text-specific modality being covered (if applicable) eg. visual, aural, written, etc.
+- the skills required to differentiate between the input and the output of the transformation (if applicable) eg. causal-reasoning, visual-reasoning, social-reasoning, humor, figurative-language, grammar, world-knowledge, etc. 
+- other transformation specific keywords
+
+Some of the categories can be subjective and they are meant to be used as best estimated. It is okay to include multiple of each type. Example, many transformer based generators might fit into most of the linguistic categories.  If your transformation is best described by keywords not yet included in the list below, please add the additional keywords to this file as part of your pull request.
+
+Keyword | Description
+------- | -----------
+**Type of Linguistic Change** |
+  `morphological` | character level or inflectional/morphological changes
```
Sohl-Dickstein, line 17:
```suggestion
  `morphological` | character level or inflectional/morphological changes
```

In BIG-bench, I indented with unicode space characters, which are preserved during markdown display (see proposed edit to this line). You might want to consider doing something similar here, if you want the keywords to appear indented from the section headers. You can also use \&nbsp; if you want it to be more explicit.

docs/keywords.md
```diff
@@ -0,0 +1,56 @@
...
+  `lexical` | the transformation substitutes different words or phrases
+  `syntactic` | the transformation changes the syntax of the input (eg. active to passive)
+  `word-order` | if the order of the words or phrases is changed (eg. topicalisation, changing the order of semantic roles)
+  `discourse` | operating on or generating units of language longer than a sentence (eg. paragraph, multiple utterances in a dialog setting, etc)
+  `noise` | The generation adds random noise which would perturb the examples in a non-traditional manner (eg. removing punctuations, changing case, repeating words/characters)
```
Sohl-Dickstein, line 22: what does traditional mean here?

docs/keywords.md
```diff
@@ -0,0 +1,56 @@
...
+  `api-based` | if the implementation uses an external api or tool
+  `external-knowledge-based` | if the implementation uses a corpora, a separate knowledge base, etc.
+  `transformer-based` | if the implementation uses a transformer based model (eg. BERT, T5)
+  `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
+  `tokenizer` | if the implementation uses a tokenizer
+  `chunker` | if the implementation uses a chunker
+  `discourse-segmenter` | if the implementation uses a discourse segmenter
+**The naturalness of the generation** |
+  `natural-sounding` | if the generations might sound natural
+  `natural-looking` | if the generations might look natural.
```
Sohl-Dickstein, line 35: Could you expand on the distinctions between looking and sounding natural?

Sohl-Dickstein (commented):
I would also add a test that verifies that
- transformations define keywords
- those keywords are all defined in this keywords.md file

docs/keywords.md
kaustubhdhole: Thanks for the clarity. Accuracy was intended to give a measure of how well the transformation generates what it intends to generate - and yes in the case of sentence operation- it would imply preserving meaning. And since I've merged precision and accuracy both now, I will leave out precision (I initially saw precision aligned with your definition to cater to rule-based approaches)  

I have updated that. Happy to revise it if it looks confusing.

docs/keywords.md
kaustubhdhole: I've removed that. I think that can be confusing.

docs/keywords.md
kaustubhdhole: Okay, thanks for pointing out. I've reverted the definitions to `unnatural-sounding` and `unnaturally-written`. Maybe this is clearer and helpful?

docs/keywords.md
```diff
@@ -0,0 +1,56 @@
...
+  `discourse` | operating on or generating units of language longer than a sentence (eg. paragraph, multiple utterances in a dialog setting, etc)
+  `noise` | The generation adds random noise which would perturb the examples in a non-traditional manner (eg. removing punctuations, changing case, repeating words/characters)
+ 
+**Type of algorithm** |
+  `rule-based` | if the implementation uses a heuristic or rules based approach
+  `model-based` | if the implementation uses a machine learning model
+  `api-based` | if the implementation uses an external api or tool
+  `external-knowledge-based` | if the implementation uses a corpora, a separate knowledge base, etc.
+  `transformer-based` | if the implementation uses a transformer based model (eg. BERT, T5)
+  `parser-based` | if the implementation uses any syntactic (dependency, constituency, etc) or shallow semantic parser or other semantic parsers (eg. semantic role labelling)
```
tongshuangwu, line 29: Are we expecting people to pick multiple types here?
e.g. If I'm making a call to an existing model service online, would it count as both `model-based` and `api-based`? And would `external-knowledge` count as a specific kind of `rule-base`, and `parser/transformer` part of `model`?

If so, maybe we can explicitly say "feel free to pick all that applies"?

kaustubhdhole:
@Sohl-Dickstein I've also added a test check for keywords (partially).

docs/keywords.md
```diff
@@ -32,7 +32,7 @@ Keyword | Description
   `discourse-segmenter` | if the implementation uses a discourse segmenter
 **The naturalness of the generation** |
   `unnatural-sounding` | if the generations might not sound natural eg. if synthesized to speech, would they not sound natural as their source sentences. Eg, adding noun definitions might make the sentence factually correct but people might not use the same language while speaking.  
-  `unnatural-looking` | generations in written format which might not look natural to a reader. homophonic translations, filler words
+  `unnaturally-written` | generations in written format which might not look natural to a reader. homophonic translations, filler words
 **The potential accuracy & precision of the generation** | 
   `high-precision` | The transformation generates what it intends to generate
   `low-precision` | The transformation sometimes does not generate what it intends to generate (eg. often changes meaning)
```
tongshuangwu, line 38: I think people would be inclined to choose `low-precision` for what they create, ha XD

`highly-meaning preserving` v.s. `possible-meaning alternation`? or do we expect other forms of low precision?

docs/keywords.md
```diff
@@ -32,7 +32,7 @@ Keyword | Description
   `discourse-segmenter` | if the implementation uses a discourse segmenter
```
tongshuangwu, line 32: why is `tokenizer`, `chunker`, `discourse` not `-based`?

docs/keywords.md
```diff
@@ -0,0 +1,53 @@
...
+  `unnaturally-written` | generations in written format which might not look natural to a reader. homophonic translations, filler words
+**The potential accuracy & precision of the generation** | 
+  `high-precision` | The transformation generates what it intends to generate
+  `low-precision` | The transformation sometimes does not generate what it intends to generate (eg. often changes meaning)
+  `low-coverage` | The transformation might not return an output for all types of inputs but only a handful of inputs
+  `high-coverage` | The transformation can return an output for almost all types of inputs
+  `high-generations` | The transformation generates a lot of (possibly diverse) outputs for a single input (eg beam search or substitutions with large number of alternate choices)
+  `low-generations` | The transformation generates few outputs for a single input (substitution with lesser number of alternate choices)
```
tongshuangwu, line 42: Switch the order of `high-generations` and `low-generations`?

docs/keywords.md
```diff
@@ -0,0 +1,53 @@
...
+**The text-specific modality being addressed (if applicable)** |
+  `visual` | The outputs are visually similar to the inputs (eg. Leet letters)
+  `aural` | The outputs sound similar to the inputs (eg. homophones)
+  `written` | The transformation makes font changes, or changes which would look different in different written formats
```
tongshuangwu, line 46: This seems to be very related to `naturalness`. Maybe move it to be before `naturalness`?

docs/keywords.md
```diff
@@ -0,0 +1,53 @@
...
+**The skills required to differentiate between the input and the output of the transformation (if applicable)** |
+  `causal-reasoning` | ability to reason about cause and effect
+  `visual-reasoning` | ability to reason by the appearance of the text
+  `social-reasoning` | ability to understand and reason about human social interactions
+  `figurative-language` | ability to understand figures of speech
+  `world-knowledge` | ability to understand world knowledge
```
tongshuangwu, line 52: Would it make sense to somehow merge modality into this category? I feel we are having too many categories, and e.g. `visual reasoning` would closely overlap with `written`, `visual`, or `unnaturally written` in previous sections.

docs/keywords.md
kaustubhdhole: Yes, multiple types should be possible. In your example, we should consider both. 

I had mentioned about multiple types in line 12, but I'll make it more explicit. I have changed it.

docs/keywords.md
kaustubhdhole: I guess other forms would be possible in other tasks which explicitly change the label too, hence I added 'precision'. But a majority of the transformations are indeed paraphrasers so I'll include these keywords too.

docs/keywords.md
kaustubhdhole: made it consistent for all.

docs/keywords.md
kaustubhdhole: You are right. I've removed visual-reasoning now for now.

docs/keywords.md
kaustubhdhole: Makes sense. Combined.

docs/keywords.md
kaustubhdhole: I didn't use based because it would probably mean that the whole transformation is based on the tokenizer or chunker. But anyways, I have appended based now.

evaluation/README.md
```diff
@@ -42,7 +42,7 @@ Note that it's highly possible that some of the evaluate_* functionality won't w
 
 ## Leaderboard
 
-Here, we provide a leaderboards for each default task, by executing transformations on typical models in each task. If you would like to join the leaderboard party encourage you to submit pull requests!
+Here, we provide a leaderboards for each default task, by executing transformations on typical models in each task (p=20). If you would like to join the leaderboard party encourage you to submit pull requests!
```
kaustubhdhole, line 46: @tongshuangwu just wanted to confirm this number?

tongshuangwu: yes~

tongshuangwu (approved):

PR #207, opened 2021-08-14
Discrimination filter

asnota:
@tongshuangwu, I'm creating a pull request on the main branch, as you requested. No conflicts detected.

PR #280, opened 2021-09-07
Discrimination filter

asnota:
Renamed the filter to group inequity.


kaustubhdhole:
Merged this assuming the code in this is the same as https://github.com/GEM-benchmark/NL-Augmenter/pull/207. Please correct me @asnota if I'm wrong.

asnota:
@kaustubhdhole you are totally right - the last changes related to the renaming of the filter were pulled here. Thank you!

PR #283, opened 2021-09-09
Enhancement: Modified setup.py to install light transformation requirements

AbinayaM02:
This PR includes changes to setup.py to install only light transformations' requirements by default. The issues faced during GitHub actions installation of heavy packages will be resolved by the fix. 

**Note:** 
There are two dictionaries each keeping tracking of the light and heavy transformations and filters in test/mapper.py. All the future PRs (yet to be merged) must have an entry in the test/mapper.py in the corresponding dictionary.


setup.py
```diff
@@ -8,8 +8,10 @@
 def all_folders():
     folder_names = [
         "transformations/" + f
-        for f in list(OperationRuns.get_all_folder_names())
-    ]
+        for f in list(
+            OperationRuns.get_all_folder_names("transformations", "light")
```
aadesh11, line 12: Though will use 'light' transformations only, I think it will be good if we take this as param in all_folders function and pass it from recursive_requirements i.e line no. 49. Because in case if the user wants to change the behaviour from 'light' to 'all' or something else, then will just need to change the param in line 49 instead of traversing the code and then changing it in line 12.

AbinayaM02: Yes @aadesh11, I had made that changes when I made this PR a draft. Looks like some other issue is happening when we install light transformations only. Will fix the all the issues and push the changes again :)

TestRunner.py
```diff
@@ -83,20 +88,54 @@ def _load_all_transformation_test_case(
...
-
-                is_heavy = cls.is_heavy()
+
+        # Import light transformations only
+        if heavy is False and search == "transformations":
+            for m in map_transformation["light"]:
+                print(f"Directory = {m}")
+                if m in disable_tests_for:
+                    continue
+                self._load_single_transformation_test_case(m, search)
```
aadesh11, line 98: We need to handle one more case where a transformation.py has multiple classes, which can have a mix of heavy and light transformation. As we are only checking the package name to distinguish b/w heavy and light, we need to pass a param 'is_heavy' in function "_load_single_transformation_test_case" and after import_module we can check for which class we need to load the test cases based on 'is_heavy' param.

TestRunner.py
```diff
@@ -83,20 +88,54 @@ def _load_all_transformation_test_case(
...
+        # Import heavy transformations only
+        elif heavy is True and search == "transformations":
+            for m in map_transformation["heavy"]:
```
aadesh11, line 101: Duplicate code 101-105: Can we create a function and pass is_heavy param?

TestRunner.py
AbinayaM02: Done!

TestRunner.py
AbinayaM02: Done!

AbinayaM02:
Though the checks are passing, the expected behavior for test case execution is not happening! Will check and fix it in a while and let you know @aadesh11.

AbinayaM02:
> Though the checks are passing, the expected behavior for test case execution is not happening! Will check and fix it in a while and let you know @aadesh11.

It's done! Now the behavior is working as expected. By default, only light transformations will be installed and pytest will execute only light transformations and filters.

aadesh11 (approved):

aadesh11 (approved):

aadesh11 (approved):

kaustubhdhole (approved):

PR #292, opened 2021-09-17
Update requirements.txt

Quetzalcohuatl:
I tried running the evaluate.py and ran into a TON of libraries that needed to be installed.
I wrote down the ones that I had to install. They were all for transformations. I added them in this commit.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,50 @@ torch
 num2words==0.5.10
 word2number==1.1
 
+# for change_date_format
+dateparser
+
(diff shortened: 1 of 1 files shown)


AbinayaM02:
Hi @Quetzalcohuatl: You can follow the steps for installation specified in the README.md [https://github.com/GEM-benchmark/NL-Augmenter#installation]. The main requirements.txt will contain some common libraries only and the transformation/filter specific libraries are present in their respective folders. Thanks!

Closing the PR.

PR #296, opened 2021-09-20
fix format of English inflection task.

sebastianGehrmann:
This adds the appropriate flags to the task which were previously set as object attributes


kaustubhdhole:
Thank you very much @sebastianGehrmann for fixing this.

kaustubhdhole (approved):

PR #301, opened 2021-09-25
Refined polarity filter

maobedkova:


kaustubhdhole:
Thank you very much @maobedkova!

Issue #310, opened 2021-10-06
`GermanGenderSwap` Violates Naming Convention


boyleconnor:
Our transformation packages so far have all been in `snake_case`. @raft001, I believe this is your transformation. Could you change this, please?

raft001:
@cascadianblue , Yes, that was an oversight on my part. I have made a new pull-request to fix this, the original one can be removed.

PR #312, opened 2021-10-07
Added german_gender_swap

raft001:
Added in response to this issue: https://github.com/GEM-benchmark/NL-Augmenter/issues/310#issue-1017654140


kaustubhdhole:
@raft001 you should also delete the earlier one too...

PR #325, opened 2021-10-09
NL-Augmenter Transformations Keyword Mapping Analysis

Saad-Mahamood:
Initial first version of the script to extract and group keywords from each of the transformations into a CSV file.


analysis/transformations_mapping_analysis.py
```diff
@@ -0,0 +1,154 @@
...
+import inspect
+import os
+import pandas as pd
+from importlib import import_module
+import json
+
+from interfaces.Operation import Operation
+
+class TransformationMappingAnalysis():
+    ignore_list = ["GermanGenderSwap", "disability_transformation", "gender_neutral_rewrite",
```
kaustubhdhole, line 17: The GermanGenderSwap has been appropriately renamed now to the underscore format in https://github.com/GEM-benchmark/NL-Augmenter/pull/312.

PR #272, opened 2021-09-01
Adding Diacritic Filter

vyraun:
This filter checks whether any character in the sentence has a diacritic.

Diacritics are marks placed above or below (or sometimes next to) a letter in a word to indicate a particular pronunciation -- in regard to accent, tone, or stress — as well as meaning, especially when a homograph exists without the marked letter or letters. For example, pâte refers to clay whereas pate refers to the head, and résumé or resumé is used for a work history versus resume, which means "to begin again."

Neural Sequence models have discrete vocabularies and commonly use subword segmentation techniques to achieve an ‘open vocabulary.’  This filter can be used to create splits of the dataset where the sentences have diacritics. Such accented characters are typically among the rarer characters (& will likely result in different subword representations) and checking the model performance on such a split would be useful with respect to model robustness.

This PR includes the test cases & references as well. Thanks.

PR #349, opened 2021-10-18
adding diacritic filter to mapper

vyraun:
For this PR: https://github.com/GEM-benchmark/NL-Augmenter/pull/272

### DIFF
--- a/test/mapper.py
+++ b/test/mapper.py
@@ -99,6 +99,7 @@ map_filter = {
         "speech_tag",
         "token_amount",
         "yesno_question",
+        "diacritic_filter",
     ],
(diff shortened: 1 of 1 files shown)


AbinayaM02:
> Thanks @AbinayaM02 , yes, I had pointed this out to @aadesh11 . @aadesh11 please take a look at #272 and #349 . Thanks!

When #272 gets merged, the tests for diacritic filter will not be run as part of our workflow since the filter is not added to the test/mapper.py. So, instead of having the addition of filter name to test/mapper.py as a separate PR, you can add it to #272 and all the checks will happen at once.

PR #359, opened 2021-10-30
bugfix indexing error on multilingual_lexicon_perturbation

SamuelCahyawijaya:


SamuelCahyawijaya:
I think the error is caused by `ChangeDateFormat`  transformation which is not related to our changes whatsoever.  How could we go through this problem? Please advice 👍🏻

PR #360, opened 2021-10-30
Bugfix/numeric to word

SamuelCahyawijaya:
- Change isnumeric() to isdigit()
- Handle unhandled cases in some number transformations


SamuelCahyawijaya:
I think the error is caused by `ChangeDateFormat`  transformation which is not related to our changes whatsoever.  How could we go through this problem? Please advice 👍🏻

AbinayaM02:
> I think the error is caused by `ChangeDateFormat` transformation which is not related to our changes whatsoever. How could we go through this problem? Please advice 👍🏻

Hi @SamuelCahyawijaya: The error with `ChangeDateFormat` is fixed a while ago. The current error is because of a mismatch in `numeric_to_word` transformation only. Please pull the latest code and fix the error.

PR #362, opened 2021-11-02
fix the bug for typo dicts rather than dict

JZCS2018:


JZCS2018:
Sorry for the confusion. I don't know why it pulled requests for twice. I found errors when I run the colab and fixed this issue.

kaustubhdhole (approved):

PR #363, opened 2021-11-02
Added default argument for instantiating ToxicityFilter class

saqibns:
Added a default argument to enable the instantiation of the `ToxicityFilter` class without providing any arguments. This was required because the evaluation interface for robustness lacks the option of providing custom options for instantiating classes.

### DIFF
--- a/filters/toxicity/filter.py
+++ b/filters/toxicity/filter.py
@@ -24,7 +24,7 @@ class ToxicityFilter(SentenceOperation):
 
     def __init__(
         self,
-        toxicity_type: ToxicityTypes,
+        toxicity_type: ToxicityTypes = ToxicityTypes.TOXICITY,
         op: str = ">",
         threshold: float = 0.5,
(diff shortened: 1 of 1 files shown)


AbinayaM02 (approved):
Hi @saqibns: Thank you for adding this enhancement. Please pull the main once and we can merge your changes.

filters/toxicity/filter.py
```diff
@@ -24,7 +24,7 @@ class ToxicityFilter(SentenceOperation):
```
aadesh11, line 24: It would be great if you can add docstring for your classes and functions.

filters/toxicity/filter.py
```diff
@@ -24,7 +24,7 @@ class ToxicityFilter(SentenceOperation):
 
     def __init__(
```
aadesh11, line 25: Please add keywords for your filter.

PR #368, opened 2021-11-16
Updated installation options for NL-Augmenter

AbinayaM02:
As part of this PR, the following are done
- Added default installation option (all light transformations and filters and common requirements will be installed by default)
- Added options to install individual heavy transformations/filters using pip install once the packaging is done. 

Usage:
```python
pip install NL-Augmenter[toxicity]
```
- Next step is to build and upload the package to pypi [will be added as a separate PR]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,22 @@ def get_extra_requirements() -> dict:
     from the generated dictionary will be picked up and installed along with the default
     requiremens.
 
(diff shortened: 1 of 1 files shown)


aadesh11 (approved):

AbinayaM02:
Merging the changes since @aadesh11 needs to work on packaging.

